PR #249, opened 2015-02-10
Fix MVV NumberFormatException plus Cleanup

pfent:
Fixed MVV NumberFormatException that caused several bugs on 1.2.1
Cleanup contains several suggested fixes from codeinspection


kordianbruck:
Nice job!

Issue #269, opened 2015-05-13
MVV Widget


enricogigantiello:
when the mvv research will be working( #245) it would be nice to introduce a widget to display needed timetables.

kordianbruck:
@enricogigantiello please use the v14 branch when creating the PR

Carlinho89:
Done. Pull request #293

kordianbruck:
Done.

mammuth:
:tada:

PR #270, opened 2015-05-15
Add a homescreen Widget displaying the information from MainActivity cards

pfent:
This PR depends on API Level 11, I don't really know how many users are still on Android 2.3 and below, but tumcabe bug reports suggest there are few.

The CardsWidget displays the information of the Cards used on the MainActivity in condensed form and lets the user access the Cards information directly from the homescreen

screenshot_2015-05-15-17-09-50


nilstk:
About 450 (of 7200) active installations of the TCA run on an Android system of version 2.3.3 and lower.
All other are using version 4.x and higher.

kordianbruck:
Current numbers from the database from clients which upgraded to 1.2 version of the App:
1 	3.2.1
37 	2.3.7
77 	2.3.6
29 	2.3.5
23 	2.3.4
14 	2.3.3

Total is about 7000 devices... 
(Google shows also devices which use the old 1.1.x app version)

app/build.gradle
```diff
@@ -24,10 +24,10 @@ android {
 
 dependencies {
     compile fileTree(dir: 'libs', include: ['*.jar'])
-    compile 'com.android.support:appcompat-v7:21.0.3'
-    compile 'com.android.support:support-v4:21.0.3'
+    compile 'com.android.support:appcompat-v7:22.1.1'
+    compile 'com.android.support:support-v4:22.1.1'
```
kordianbruck: Were there any problems with the updated libs?

(Any deprecations in GMS?)

pfent: ActionBarActivities are deprecated now and we probably should use AppCompatActivities now.

haven't noticed anything else yet, so you can probably pull this seperately from the pull request

pfent:
So in general AppWidgets with collections (i.e. ListView) need API level 11, on API < 11 this CardWidget does not work at all.
We will probably run into this with #269 as well, so we maybe should consider creating a separate feature branch and publish 2 separate apks?
https://developer.android.com/google/play/publishing/multiple-apks.html

PR #271, opened 2015-05-17
Update dependencies

pfent:
Bumped the version numbers on several dependencies. As a separate PR from #270

Tested this and could not see anything broken.

Emojicon now only builds with the deleted v4 intermediate files.


kordianbruck:
Thanks! This looks good!

PR #273, opened 2015-05-21
Create a v14 feature branch for the Cards Widget

pfent:
As discussed in #270 I created a v14 feature branch
According to https://developer.android.com/google/play/publishing/multiple-apks.html#SupportedFilters this can easily be published in the Google play store (needs to have a bigger, unique version code than )

I'll probably rebase this branch on master regularely, so just keep pushing to master.


kordianbruck:
Good idea! 

I cannot guarantee, that we'll publish a v14 APK right away, but we can definably look into it if we have more features in the v14 branch.

PR #281, opened 2015-05-28
fix build warnings

pfent:
I noticed some warnings in the travis build logs:
+ application@android:icon was tagged at AndroidManifest.xml:60 to replace other declarations but no other declaration present
+ libpng warning: iCCP: Not recognizing known sRGB profile that has been edited
+ libraries/emojicon/src/github/ankushsachdeva/emojicon/EmojiconSpan.java uses or overrides a deprecated API

There shouldn't be any noticable differences in the images though


kordianbruck:
Ok, but what about upgrade compatibility of the emoji lib if we start changing small passages?

PR #282, opened 2015-05-29
Update gradle files

pfent:
remove unused okhttp dependency,
updated licensesdialog to 1.7.0 and
enabled javac Xlint, so tthese are included in the travis output.


kordianbruck:
okhttp is used by retrofit.... I guess it defines it own dependencies...

pfent:
hm. I'm not sure about that okhttp.
https://square.github.io/retrofit/ recommends to add it to your gradle dependencies, so we probably should add it again...

PR #289, opened 2015-07-14
TU Kino

thomasmauerer:
#264


app/src/main/res/values-de/strings.xml
```diff
@@ -123,6 +133,7 @@
     <string name="organisations">Organisationsbaum</string>
     <string name="mvv">MVV</string>
     <string name="news">News</string>
+    <string name="kino">TU Kino</string>
```
ibhh: Why do you add this line?
Android uses the string from ../values/string.xml if not defined here,  isn't it?

thomasmauerer: I think you are right.
However, I just stick to the status quo. It's the same with e.g. mvv

PR #293, opened 2015-07-15
Feature/mvv widget

Carlinho89:


app/src/main/java/de/tum/in/tumcampus/models/managers/MockMoodleManager.java
```diff
@@ -0,0 +1,154 @@
...
+import de.tum.in.tumcampus.models.MoodleEventsList;
+import de.tum.in.tumcampus.models.MoodleToken;
+import de.tum.in.tumcampus.models.MoodleUser;
+import de.tum.in.tumcampus.models.MoodleUserCourseList;
+
+/**
+ * Created by a2k on 6/7/2015.
+ * This class is a stub for RealMoodleManager
+ */
+public class MockMoodleManager extends MoodleManager{
```
ibhh: Is this class still used?

Carlinho89: Totally forgot about it, we used it when we started developing. It can be removed.

kordianbruck: @Carlinho89 please create a new PR with a proper fix, thanks!

PR #296, opened 2015-07-16
Removed unused MockManager class

Carlinho89:


.idea/vcs.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="VcsDirectoryMappings">
+    <mapping directory="$PROJECT_DIR$" vcs="Git" />
+    <mapping directory="$PROJECT_DIR$" vcs="Git" />
+  </component>
+</project>
+
```
kordianbruck: Why are you commting these files?!

Carlinho89: I think it was Android Studio, maybe they aren't in the gitignore?

kordianbruck: Well yeah, it is an android studio file but don't you guys check the files you commit before commiting them?

Carlinho89: how can I solve this issue?

kordianbruck: Forget it, I'll do some cleanup later..

PR #304, opened 2015-08-30
Replace the current SwipeDismissList with a RecyclerView for MainActivity

pfent:
The current implementation of the Cards in the MainActivity has several flaws. Cards could get stuck while swiping and the animations keep stuttering, even on newer devices.

This PR replaces the ListView with the custom SwipeDismissList with a Recyclerview.


kordianbruck:
No immediate errors were found

PR #305, opened 2015-09-01
replace deprecated Apache HTTP with HttpURLConnections

pfent:
This is a rewrite of the NetUtils to account for the fact that Google deprecated and removed the Apache HTTP libraries from the Android SDK.

This probably needs some testing, although it seemed to work fine in the simulator.
We should especially test, if we still receive stacktraces.


kordianbruck:
PR looks good. Lot's of unrelated changes though... Maybe you could make the commits smaller, so it is easier to check them out.

Issue #284, opened 2015-06-10
Background service: WiFi only


kordianbruck:
Only download data in the background via Wifi. Make this configurable.

kordianbruck:
@brdvlps will take a look at it

kordianbruck:
Implementd.

PR #306, opened 2015-09-11
Background Service: Add Wifi-only option

brdvlps:
Users can choose via settings menu if they want the background service download to start always or only when connected to a wifi network. (see #284)

It seems to work, but I haven't tested it properly atm.


pfent:
Looks good to me.

We already have a method of testing the WiFi connection with `de.tum.in.tumcampus.trace.Util.isWifiOn()`, but that whole Util class is in desperate need of a rewrite.

@kordianbruck feel free to merge

kordianbruck:
Aweome! Yeah, I will refactor the trace class at some point once again...

PR #310, opened 2015-09-21
Update Ressources

pfent:
The biggest change should be the new Icon and an updated material color scheme

I changed the Icon to satisfy the Material Design guidelines:
* The new Icon is slightly bigger and thous equal in size to other round App-Icons (e.g. Chrome, Google Plus)

> new:  
> new TCA icon

> old:  
> old TCA icon

* The main color of the color scheme is now darker, i.e. the contrast between white text and blue background is now better

> new:
> new color sheme

> old:
> old color sheme

I also removed all unused ressources. If you need to access them, you can still find them in the git history


kordianbruck:
Looks great! 

You deleted the beta logo, but oh well... weren't really using that in the first place. Maybe you could send me over the source image file (PSD or equivalent) in case I feel the urge to re-add that?

pfent:
There is no source image file, because I just generated it with this tool: <https://romannurik.github.io/AndroidAssetStudio/icons-launcher.html>

kordianbruck:
Ah, works as well... Thanks!

PR #312, opened 2015-09-22
Enable travis runs of unit tests

pfent:
With this pull request, travis should be able to automatically run all our unit tests in `/app/src/androidTest/`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ branches:
 
 language: android
 jdk: oraclejdk7
+
 android:
   components:
     - tools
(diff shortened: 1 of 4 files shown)


kordianbruck:
Not gonna work out?

pfent:
Nope. I ran in too many problems with the travis configuration.
E.g. travis does not support running x86 emulators and as fallback I needed to use armeabi-v7a images which are horribly slow, so the app installation times out...

I don't think we can reliably do any testing with emulators on the free travis instances...

kordianbruck:
Correct me if I am wrong, but do we need to really do the testing on the emulator? Why can't we use something more simple? ([just found this](http://blog.stablekernel.com/unit-testing-continuous-delivery-for-android-part-3/))

It would be enough to directly test the code without the emulator...

pfent:
Sure, but we obviously can't test our Activities then.

PR #314, opened 2015-09-22
Silence Service: remember old state

DerGenaue:
a feature urgently requested by a friend of mine.

I'm not sure whether this corresponds to any internal style guidelines


pfent:
Looks good to me.

kordianbruck:
Apart from having asdf as a merge commit message, it looks good. 

A little explanation would be nice, what you actually want to achieve with this...

DerGenaue:
That asdf part was only the "commit" to get my fork updated to the base fork.
I didn't think that would actually not be committed...

DerGenaue:
Well... I want to get the "Silence Server [to] remember [the] old state"
and switch back to it.

Like, when a user has his Phone on Silent, goes to a lecture and comes back it won't get set to normal but remembers it was already set to silence before the Silence Service became active.

Currently the Silence Service would always set it back to normal

kordianbruck:
Gotch ya, looks good then!

PR #315, opened 2015-09-25
Replace AOSP Calendar with android-week-view

pfent:
The current implementation of the calendar is no good solution. The AOSP calendar code was not maintained since ~2 years.

In this implementation I am using [Android Week View](https://github.com/alamkanak/Android-Week-View) to display the calendar, which significantly reduces our code complexity (net change: -6000 LOC).

Furthermore, this resolves _all_ warnings being reported by the compiler.


kordianbruck:
Looks good! Good idea, we were using some other library before and searched for a proper alternative to the old buggy one but couldn't find any. The Android week view way first created just after we released the new version with the ripped out google cal.

Might have to check with the formatting of the time axis, sometimes (with language set to English e.g.: time in the format of 2:00 PM) text is cut off.

pfent:
Could you screenshot this?
I tested on English and German and everything looked fine :-O

kordianbruck:
screenshot from 2015-10-01 22 45 37
screenshot from 2015-10-01 22 46 09

PR #316, opened 2015-09-30
Use an AlarmManager with SilenceService

pfent:
Keeping the SilenceService running to check for lectures wasn't a good idea to begin with.

The current problems are:
* If the TCA is killed, the SilenceService won't work. This is especially a problem on devices with low RAM
* The SilenceService is causing periodic device wakeups, which won't work with [Android Marshmallow's Doze](https://developer.android.com/training/monitoring-device-state/doze-standby.html)
* The TCA needs to keep quite some RAM to be constantely running in background

Switching to AlarmManager fixes those problems and thous should reduce battery drain

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -43,6 +43,9 @@
     <!-- For QR code scanner -->
     <uses-permission android:name="android.permission.CAMERA" />
 
+    <!-- To start the SilenceService when the device boots -->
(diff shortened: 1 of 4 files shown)


pfent:
It would probably also be nice to have the SilenceService automatically starting on device boot and not needing to start the TCA once to enable the SilenceService.

kordianbruck:
@Check: it already resets the silenceservice in the **StartSyncReceiver** class - never checked if this works.

Looks good code wise to me, should deploy a first beta in the next week, so we can stick to schedule and push the next update before Oktober 15th

PR #324, opened 2015-10-04
Improve translation

JonWel:
I quickly improved the translation.
I'll do more later when I'll have some time (I need to compare to the English and German version to translate correctly certain fields).


pfent:
Thanks for your contribution. Improoving translations is especially appreciated.
If you have AndroidStudio installed, you can use the Translations Editor. In it you can directly compare English/German/Spanish/French…

JonWel:
Thanks for the AndroidStudio advice, this will definitely be useful.

Issue #338, opened 2015-10-22
If chat is not properly setup, disable it


kordianbruck:
```
java.lang.NullPointerException: Attempt to invoke virtual method 'int de.tum.in.tumcampus.models.ChatMember.getId()' on a null object reference
	at de.tum.in.tumcampus.models.ChatVerification.<init>(ChatVerification.java:22)
	at de.tum.in.tumcampus.activities.ChatRoomsActivity.createOrJoinChatRoom(ChatRoomsActivity.java:205)
	at de.tum.in.tumcampus.activities.ChatRoomsActivity.onItemClick(ChatRoomsActivity.java:291)
```

We are still getting a lot of the above exceptions. We need to handle those events and disable the chat when the setup is not correct. (I.e.: PK creation failed, did not activate the token and so on)

kordianbruck:
This is still an issue and we need to fix this asap. Leads to FC of the App.

kordianbruck:
@pfent any other ideas on how to fix this issue? I don't have any idea on how it happens that the chat member is not set.

pfent:
Oh wow, I wanted to reproduce this and ran into multiple issues.
Startup Wizzard: Ignore Email -> click on Next -> App crashes
Next startup, the Chat thinks, it is already enabled and doesn't show any new prompt to enable the chat.
Chat is still accessible from side menu

When clicking on the Chat from the side menu, the app crashes (probably the cause of #339)

pfent:
Also we currently only check, if the user has ticked the "enable Chat" checkbox to show the Drawer item for the Chats. We should probably rework that, especially for older not probably setup users.

Probably introduce a new setting, if the chat is properly confirmed

kordianbruck:
Should also be fixed, now that #339 is fixed

PR #350, opened 2015-12-28
Feature new gcm

### LINKED ISSUE #349: Rework GCM registration
Our current approach to registering the App to GCM is handled in the ChatRoomsActivity, so users that did not start this Activity once are not registered. We probably should do this in the StartupActivity.

Furthermore, the GoogleCloudMessaging.register() method is deprecated since May 2015 and we should switch to the new Instance ID API (see: https://developers.google.com/cloud-messaging/registration and
https://developers.google.com/cloud-messaging/android/legacy-regid)

kordianbruck:
This PR fixes #349 and probably will help with #338.

From now on every device will generate a key pair on startup even if the chat is not enabled. Also it will directly try to register with gcm if it is installed so the alarm system works for everyone out of the box. 

Beside that I improved many details as there is no a central class doing all the work with signing. This makes it very easy to implement a verification of the device on every request in the future, which will reduce abuse of our apis. 

@pfent @brdvlps please test if you got any spare time


app/src/main/AndroidManifest.xml
```diff
@@ -73,23 +73,17 @@
...
-            android:permission="com.google.android.c2dm.permission.SEND">
+            android:name="com.google.android.gms.gcm.GcmReceiver"
+            android:exported="true"
+            android:permission="com.google.android.c2dm.permission.SEND" >
             <intent-filter>
                 <action android:name="com.google.android.c2dm.intent.RECEIVE" />
-                <action android:name="com.google.android.c2dm.intent.REGISTRATION" />
-
-                <category android:name="de.tum.in.tumcampus" />
+                <category android:name="gcm.play.android.samples.com.gcmquickstart" />
```
pfent: wat?

PR #355, opened 2016-01-04
Convert project into a lib project 

kordianbruck:
Follow up to #352 after the recent PR

Everything should work just fine now


kordianbruck:
Found a great source for correct proguard files for libs: https://github.com/krschultz/android-proguard-snippets

Issue #231, opened 2014-11-25
Study rooms


kordianbruck:
implement this somehow: https://campus.tum.de/tumonline/wbrisweb.status_mobile

more infos here: http://www.asta.tum.de/themen-projekte/lernraeume/

brdvlps:
So ... any news about that api?

kordianbruck:
@brdvlps I just asked about that again, but we might need to fall back to the good old fashion html parsing...

brdvlps:
Sorry for being so quiet lately. Just a quick status update: I'm still working on this feature. It just needs some UI fixing (maybe also in other sections, we'll see) and should then be ready.
Will continue and hopefully finish my work the next weekend(s).

PR #363, opened 2016-04-08
Feature: Study Rooms

brdvlps:
Finally, the long-awaited study rooms feature. See #231.


app/src/main/java/de/tum/in/tumcampusapp/activities/generic/ActivityForSearching.java
```diff
@@ -92,7 +92,6 @@ public boolean onCreateOptionsMenu(Menu menu) {
         // If activity gets called via Intent with a search query set SearchView accordingly
         if (mQuery != null) {
             mSearchView.setQuery(mQuery, false);
-            mOpenSearch = true;
```
brdvlps: Notice I changed the behavior here, i.e. the search textbox won't be opened if there is a predefined query.
Makes more sense imho.

kordianbruck:
Awesome! Please fix the lint errors so we can proceed to merge 👍

brdvlps:
Done. I ignored the remaining errors concerning the new 24.0.0 alpha of the android support libraries.

pfent:
@brdvlps thanks again for the PR, I finally came around to merge it...

Please note, that I changed the generic base Activity, since I got some loading problems, when my phone initially was only connected via mobile data. See: dcfa28ba6b5b2c120503ef86b924117037ecddc7

PR #367, opened 2016-06-05
MMV-Plans feature

aserhany:
- Created file provider to share downloaded pdfs with external apps
- Defined the file provider in the manifest
- Created FileDownloader help class to download the mvv plans from the website
- modified plansActivity respectively.

Please consider the following:
1. There is a new version of the document "Nachliniennetz" compared to the one currently used in the app
2. In the "Overallplan" we did not crop the rings only, since we are taking the whole pdf
3. Only the first 4 documents will be downloaded from the mvv website, the rest is not from mvv (we assume)

Downloaded pdfs:
1. Schnellbahnnetz mit Tram und Expressbus: http://www.mvv-muenchen.de/fileadmin/media/Dateien/plaene/pdf/Netz_2016_Version_MVG.PDF
2. Nachtliniennetz: http://www.mvv-muenchen.de/fileadmin/media/Dateien/plaene/pdf/Nachtnetz_2016.pdf
3. Tramnetz: http://www.mvv-muenchen.de/fileadmin/media/Dateien/plaene/pdf/Tramnetz_2016.pdf
4. MVV Overall-Network: http://www.mvv-muenchen.de/fileadmin/media/Dateien/3_Tickets_Preise/dokumente/TARIFPLAN_2016-Innenraum.pdf

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -108,6 +108,15 @@
             android:name="de.tum.in.tumcampusapp.auxiliary.LectureSearchSuggestionProvider"
             android:authorities="de.tum.in.tumcampusapp.auxiliary.LectureSearchSuggestionProvider"
             android:exported="false" />
+        <provider
+            android:name="android.support.v4.content.FileProvider"
+            android:authorities="de.tum.in.tumcampusapp.fileprovider"
+            android:exported="false"
+            android:grantUriPermissions="true">
+            <meta-data
+                android:name="android.support.FILE_PROVIDER_PATHS"
+                android:resource="@xml/fileprovider" />
+        </provider>
 
         <activity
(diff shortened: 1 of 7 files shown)


TCA-Team:

kordianbruck:
add to whitelist

pfent:
@aserhany Please don't introduce a new FileDownloader class, when we already have a NetUtils class, that already implements the same functionality.

Also, an inline display for the PDFs would be nice to have. There is a [PDFRenderer](https://developer.android.com/reference/android/graphics/pdf/PdfRenderer.html) available for API 21+, even with an exhaustive [example on GitHub](https://github.com/googlesamples/android-PdfRendererBasic). Then you can use the fileprovider as a fallback for lower APIs.

kordianbruck:
I've clean the feature up a bit, but currently its not really working as it should. Somehow the files do not get downloaded/saved to the application directory.

PR #395, opened 2016-07-09
Feature/favourite dish

aserhany:


aserhany:
@kordianbruck @mhmadayad 
Here is the current status: 
- The code will be commented tomorrow
- We still have to handle setting new alarms for favorite dishes upon updating the Mensa data (e.g. new dates for fav. dishes, that were not in db upon marking the dish as favorite). We already have an idea how to implement it.

kordianbruck:
Why did you include files from the build directory?

aserhany:
@kordianbruck Those files were included accidentally after merge as I didn't know they are from the build directory. I reverted to the commit before and pushed it (you will find only the 5 changed files)

PR #401, opened 2016-10-08
Update the wizzard images to match TUMonline

pfent:
We haven't updated our wizard images since TUMonline got the major redesign...


kordianbruck:
Awesome :+1: 

I'm currently refactoring the signup process, as we can now use emails to activate! This is already live on the production system of TUMO. Check out my fork for the changes.

Issue #419, opened 2016-10-28
while writing long text messages in chat edit box, the edit box should expand to show the full message like whatsapp 


deepeshp:
while writing long text messages in chat edit box, the edit box should expand to show the full message like whatsapp

pfent:
Fixed in  #420

Issue #417, opened 2016-10-25
Chat message look really bad 


kordianbruck:
On a Nexus 7 - 2013
![screenshot_2016-10-25-14-39-17](https://cloud.githubusercontent.com/assets/298860/19686558/f05da626-9ac1-11e6-8aca-d2367f19b833.png)

pfent:
Fixed in  #420

kordianbruck:
This is a issue with the message layout, not the editbox.

PR #420, opened 2016-10-31
edit text should expand for long messages

deepeshp:


pfent:
Awesome!
This should fix #419 and #417.

@kordianbruck can you verify, that this fixes the issue on the Nexus 7 as well?

PR #429, opened 2016-11-23
Merge the Hackatum chat bot

konstin:
During the Hackatum Hackaton a chat bot with speech recognition was implemented. There [an official presentation](https://docs.google.com/presentation/d/15UJRzb9DbSt-Gs11kjTdpVwBgNMBvHlDTT1iBQNVoEM/edit#slide=id.p) and a [devpost page](https://devpost.com/software/tum-assistent) showing the features.

It would be awesome to have this integrated in the app. The code does of course need testing and probably some refactoring, so this PR is meant more as a tracking issue than a real ready-to-merge-PR. 

Disclaimer: I'm not affiliated with the project. I just want to have this feature ;)

CC @camodude009 @dimart @rpadovani


kordianbruck:
The rather interesting question would be, if the core technology would be available to us free of charge (https://www.luis.ai/) cause this was given by Microsoft as part of the Hackathon only as far as I remember.

pfent:
@kordianbruck Pricing for LUIS is [here](https://www.microsoft.com/cognitive-services/en-us/language-understanding-intelligent-service-luis).
Free for 10K calls per month, $0.75 per 1000 calls in the "Standard" plan.

May I ask, who registered the current API? Seems like, if we publish this in the current form, whoever registered with microsoft pays our bill…
BTW: You also might consider to redact or revoke your subscription key in the `AssistantService.java`, that's probably not intended to be public.

If we can just use the free service, that's fine for me.

konstin:
Just taking the "start for free" plan and see how it goes would imho be a good start. If we hit the API limits than we could kindly ask microsoft for a hackatum/non-profit/university discount and see if that works out.

rpadovani:
Thanks for bringing this out @konstin, it always a pleasure when your own work is appreciated :)

The one inserted at the moment is a development key (this is why we didn't care to strip it out) and it will expire on 31/12/16

I suggest to use it during the development, and then see to generate a key and store it securely somehow (maybe adding it during the build of the APK)

kordianbruck:
So basically the summary is the following: we need someone willing to bring this whole thing up to speed? I know I don't have the time, nor does @pfent. If someone is willing to put the finishing touches and organize something better than a DEV key, let me know. 

The printing part is nice, but we are not allowed by the TU IT guidelines to save a users password. So your implementation would unfortunately not work in compliance with those guidelines.

camodude009:
Hi, one of devs here :)
The only real way of solving this problem would be to make the user enter the password, every time he/she prints something. This wouldn't fit with the voice assistant though.
Sadly I do not have the time (atm) to further develop this project.
I would be happy to help or explain our code where ever needed!

pfent:
So, after letting this settle a while, I'll conclude:
* Currently, we have merge conflicts
* The current PR doesn't even compile
* We don't have a LUIS licence, but someone would need to register a free licence
* Noone currently feels responsible for making the code releasable, not even speaking of maintaining it.

However, I don't see a problem with saving Informatics logins. The RBG is more relaxed with the password security… I don't find an english version of the T&Cs, but the german version basically says: "You're responsible for whatever happens under your account, be careful".

@konstin Thanks for bringing this to our attention, but I don't see any direction where this is heading.

When there's anyone willing to work on this feature, feel free to re-open the PR, but I'm closing it for now.

Issue #443, opened 2017-01-25
Get rid of SQLite - Use Objectbox


kordianbruck:
We should refactor our code and get rid of SQLite. It does introduce many problems with exceptions and migrations. We need to do better, for the simple data model that we have, the SQLite system is too volatile.

- http://greenrobot.org/announcement/introducing-objectbox-beta/
- http://greenrobot.org/objectbox/documentation/introduction/
- https://github.com/greenrobot/ObjectBox

bkunters:
https://realm.io/products/realm-mobile-database/

kordianbruck:
@bkunters http://kpgalligan.tumblr.com/post/133281929963/my-talk-at-droidcon-uk

realm is just like greenDAO (which is made by the same guys) - objectbox is a complete rewrite in native code.

kordianbruck:
So this thing is still really unstable and segfaults at various occasions. I think this is too much of a change for this app for now. Might revisit soonish.

PR #444, opened 2017-01-27
Setup objectbox

kordianbruck:
WIP #443


app/src/main/java/de/tum/in/tumcampusapp/entities/MyBoxStore.java
```diff
@@ -0,0 +1,31 @@
+package de.tum.in.tumcampusapp.entities;
+
+
+import android.content.Context;
+
+import io.objectbox.BoxStore;
+
+public class MyBoxStore {
```
pfent: MyBoxStore is probably not a good name…

pfent:
Ehhh…
> Once a an updated License OR a final release is available, you need to **stop using any of the material** made available to you with this beta license and switch to the then current release version with its terms and conditions  

http://greenrobot.org/objectbox/license/
Does not seem like this library will be open-source or under a compatible license. Probably can't merge it with such a license…

kordianbruck:
I wouldn't worry about the license for now. I mean we can ask in the public repository, what happens when the stable comes out. Good that you spotted that problem.

app/src/main/java/de/tum/in/tumcampusapp/entities/MyBoxStore.java
kordianbruck: Kay, renamed.

kordianbruck:
@pfent their answer to the question about licensing https://github.com/greenrobot/ObjectBox/issues/29#issuecomment-276294643

kordianbruck:
Ping

kordianbruck:
On hold for some time.

Issue #441, opened 2017-01-18
eduroam assistant should check secure configuration


xsrf:
## Steps to Reproduce

1. setup "eduroam" by hand or using TCA <=1.4.7 in a not entirely secure way
2. open TCA

## Expected Result

TCA should offer "update eduroam profile to latest security standards" (it may also say "Your eduroam setup is NOT secure, please update" but we don't want users get crazy, right?)

## Actual Result

_imagine some tumbleweed passing by_

## Problem

Lots of users don't configure eduroam properly and put their credentials on risk without knowing. They should be made aware of that not every wifi is always a happy place.

## Suggested Enhancement :sparkles:

Check users eduroam profile (if exists) against latest security standards (root certificate match, server certificate subject match) and notify them if the profile is not secure.

### Environment
* Phone: Any
* OS: Android 4.x+
* TCA Version: 1.4.7

pfent:
Absolutely! Seems like a good addition to our "setup eduroam" feature and would be an easy security upgrade for all our users.

xsrf:
Oh btw. you might have users originating from another university whose profile you don't want to overwrite in this case...

If the username is within the radius zones mwn.de, *.mwn.de, tum.de, *.tum.de, mytum.de, ... or matches against "[a-z]{2}[0-9]{2}[a-z]{3}" you can be pretty sure that you know what a valid configuration should look like and check it.

If it doesn't, you might apply additional checks or just tell the user that a unknown configuration was found and ask if he is a TUM member and let him dismiss the warning.

I guess any secure eduroam configurations should do either:
* Any certificate validation + server subject match
* Validation against custom CA with or without server subject match

TUM configurations should do:
* PEAP (or TTLS)
* MSCHAPv2 (or PAP)
* Validate certificate against Telekom Root CA 2
* Server subject match against radius.lrz.de
* use anonymous identity

pfent:
should be fixed with #449

PR #449, opened 2017-03-31
Check users eduroam configuration and offer to fix it

kordianbruck:
## Issue

This fixes the following issue(s): #441

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/cards/EduroamFixCard.java
+++ b/app/src/main/java/de/tum/in/tumcampusapp/cards/EduroamFixCard.java
@@ -1,0 +1,160 @@
+package de.tum.in.tumcampusapp.cards;
+
+import android.annotation.SuppressLint;
+import android.annotation.TargetApi;
+import android.app.Notification;
+import android.content.Context;
+import android.content.Intent;
+import android.content.SharedPreferences;
+import android.net.wifi.WifiConfiguration;
+import android.net.wifi.WifiManager;
+import android.os.Build;
+import android.preference.PreferenceManager;
+import android.support.v4.app.NotificationCompat;
+import android.support.v7.widget.RecyclerView;
+import android.view.LayoutInflater;
+import android.view.View;
+import android.view.ViewGroup;
+import android.widget.LinearLayout;
+import android.widget.TextView;
+
+import java.io.InputStream;
+import java.security.cert.CertificateException;
+import java.security.cert.CertificateFactory;
+import java.security.cert.X509Certificate;
+import java.util.List;
+import java.util.Vector;
+
+import de.tum.in.tumcampusapp.R;
+import de.tum.in.tumcampusapp.activities.SetupEduroamActivity;
+import de.tum.in.tumcampusapp.auxiliary.Utils;
+import de.tum.in.tumcampusapp.cards.generic.NotificationAwareCard;
+import de.tum.in.tumcampusapp.managers.CardManager;
+
+import static de.tum.in.tumcampusapp.managers.EduroamManager.RADIUS_DNS;
+
+
+public class EduroamFixCard extends NotificationAwareCard {
+
+    private Vector<String> errors;
+    private TextView errorsTv;
+    private WifiConfiguration eduroam;
+
+    public EduroamFixCard(Context context) {
+        super(CardManager.CARD_EDUROAM_FIX, context, "card_eduroam_fix_start", false, true);
+    }
+
+    public static CardViewHolder inflateViewHolder(ViewGroup parent) {
+        View view = LayoutInflater.from(parent.getContext()).inflate(R.layout.card_eduroam_fix, parent, false);
+        return new CardViewHolder(view);
+    }
+
+    @Override
+    public void updateViewHolder(RecyclerView.ViewHolder viewHolder) {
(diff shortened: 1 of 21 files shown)


app/src/main/java/de/tum/in/tumcampusapp/cards/EduroamFixCard.java
```diff
@@ -0,0 +1,148 @@
...
+
+    @Override
+    public void updateViewHolder(RecyclerView.ViewHolder viewHolder) {
+        super.updateViewHolder(viewHolder);
+        mCard = viewHolder.itemView;
+        mLinearLayout = (LinearLayout) mCard.findViewById(R.id.card_view);
+        errorsTv = (TextView) mCard.findViewById(R.id.eduroam_errors);
+
+        String txt = "";
+        for (String e : errors) {
```
pfent: Probably use guavas [Joiner](https://github.com/google/guava/wiki/StringsExplained), so you don't get a trailing newline

app/src/main/java/de/tum/in/tumcampusapp/cards/EduroamFixCard.java
```diff
@@ -0,0 +1,148 @@
...
+
+    @Override
+    protected void discard(SharedPreferences.Editor editor) {
+        SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(mContext);
+        prefs.edit().putBoolean("card_eduroam_fix_start", false).apply();
+    }
+
+    @Override
+    protected Notification fillNotification(NotificationCompat.Builder notificationBuilder) {
+        return null;
```
pfent: Maybe show a Notification once, so more users are aware of the problem?

app/src/main/java/de/tum/in/tumcampusapp/cards/EduroamFixCard.java
```diff
@@ -0,0 +1,148 @@
...
+import de.tum.in.tumcampusapp.cards.generic.NotificationAwareCard;
+import de.tum.in.tumcampusapp.managers.CardManager;
+import de.tum.in.tumcampusapp.managers.EduroamManager;
+
+import static de.tum.in.tumcampusapp.managers.EduroamManager.RADIUS_DNS;
+
+
+public class EduroamFixCard extends NotificationAwareCard {
+
+    private Vector<String> errors;
```
pfent: Please just use a List interface, then initialize it with an ArrayList! Vector is almost always the wrong List implementation in Java.

app/src/main/java/de/tum/in/tumcampusapp/cards/EduroamFixCard.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    @Override
+    public int getId() {
+        return 0;
+    }
+
+
+    private boolean isConfigValid() {
+        errors = new Vector<>();
```
pfent: Also, probably just use a `.clear()` instead of generating a new object

app/src/main/java/de/tum/in/tumcampusapp/managers/EduroamManager.java
```diff
@@ -47,18 +47,21 @@ public EduroamManager(Context context) {
...
-    public boolean isConfigured() {
-        WifiManager wifiManager = (WifiManager) mContext.getSystemService(Context.WIFI_SERVICE);
+    static public WifiConfiguration getEduroamConfig(Context c) {
+        WifiManager wifiManager = (WifiManager) c.getApplicationContext().getSystemService(Context.WIFI_SERVICE);
         List<WifiConfiguration> list = wifiManager.getConfiguredNetworks();
+
+        //We didn't get a list, so maybe theres no wifi?
         if (list == null) {
-            return true;
+            return null;
```
pfent: Any reason not returning an Optional<WifiConfiguration> ?

app/src/main/java/de/tum/in/tumcampusapp/cards/EduroamFixCard.java
kordianbruck: I really don't want to bother the user with this. We can always do this at a later stage.

Problem is, that this detection might be a bit error prone. So for now, lets just skip that.

app/src/main/java/de/tum/in/tumcampusapp/managers/EduroamManager.java
kordianbruck: The five places that this is used, all check for null values specifically. So no need for optional overhead.

app/src/main/java/de/tum/in/tumcampusapp/cards/EduroamFixCard.java
kordianbruck: Ahh, good idea! Implemented.

kordianbruck:
@pfent can I merge?

app/src/main/java/de/tum/in/tumcampusapp/managers/EduroamManager.java
pfent: A little bit overhead is probably worth avoiding future NullPointerExceptions…
I'll hold you responsible for any future NPEs :P

pfent:
Go ahead and merge and release :ok_hand:

PR #450, opened 2017-04-05
Fix androidlint warnings

pfent:
* Update dependencies, finally extracting some version strings
* fix WifiManager leaks
* declare camera hardware as `required="false"`


app/build.gradle
```diff
@@ -45,40 +45,45 @@ android {
     publishNonDefault true
 }
 
+def androidSupportVersion = '25.3.1'
+def retrofitVersion = '2.2.0'
+def espressoVersion = '2.2.2'
```
kordianbruck: Good Idea!

Issue #453, opened 2017-05-02
App crashes when not able to download questions for the survey card


lnsp:
## Steps to Reproduce

1. Have no network connection
2. Fetch survey card questions
3. Crash and burn

## Expected Result
App should log the error and not display any survey cards.

## Actual Result
App crashes.

## Problem
Trying to get the size of a null-pointer list.

## Suggested Enhancement :sparkles:
Log and return when encountering an IOException in `getOpenQuestions()`.

### Environment
* Phone: Nexus 5
* OS: Android
* TUMOnline App: v1.4.8-dev
* Language: German

kordianbruck:
Fixed in merged PR.

PR #454, opened 2017-05-02
Fix crash when survey questions can't be fetched

lnsp:
## Issue

This fixes the following issue(s): #453 

One crash less to worry about :+1:


kordianbruck:

kordianbruck:
ok to test

Issue #430, opened 2016-11-25
Filter MVV next departures


mammuth:
It might be a good idea to filter the public transit lines that are shown in the mvv card (and the corresponding detail view) since person A always uses the U6, whilst person B mostly uses bus line 230 or U6.

What you guys think?

kordianbruck:
Are you volunteering? :+1:

kordianbruck:
Implemented. Kind of.

PR #455, opened 2017-05-03
Add public transport favorites and highlighting

lnsp:
## Issue
This change adds user customizable highlighting of public transport lines. In the detailed view, you can tap the transport lines you want to track. It is a lightweight visual aid that is in my opinion more intuitive than selective filtering.
This fixes the following issue(s): #430 

## Screenshot
screenshot_1493809703
screenshot_1493809708
screenshot_1493809714

The user can highlight his most used public transportation lines. This helps to track the latest departures he cares about.

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/activities/TransportationDetailsActivity.java
+++ b/app/src/main/java/de/tum/in/tumcampusapp/activities/TransportationDetailsActivity.java
@@ -25,6 +25,7 @@ public class TransportationDetailsActivity extends ActivityForLoadingInBackgroun
 
     private LinearLayout mViewResults;
     private RecentsManager recentsManager;
+    private TransportManager transportManager;
 
     public TransportationDetailsActivity() {
         super(R.layout.activity_transportation_detail);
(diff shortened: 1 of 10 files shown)


kordianbruck:

kordianbruck:
ok to test

kordianbruck:
Can we maybe have a info button in the toolbar to explain this? It would be hard to spot otherwise...

lnsp:
Good idea, I will look into it.

lnsp:
I added a help button in the toolbar explaining the general functionality. I am still looking into auto-refreshing the MVVCard in the MainActivity when the transport favorites are changed by the user. The same problem occurs when favoriting a dish in the cafeteria details view though.

## Screenshots
screenshot_1493815027
screenshot_1493815030

lnsp:
A workaround would be to refresh the cards in the MainActivity when resuming to it. This would also fix the issue with the non-refreshing cafeteria cards.

kordianbruck:
Probably a good idea. Lets do that in a separate PR.

PR #459, opened 2017-05-25
Sebmarste/269 mvv widget

sebmarste:
## Screenshots
screenshot_20170525-195410
the widget

screenshot_20170525-195420
configure activity open

screenshot_20170525-195428
configure activity with executed search open

Finally you can have a widget with the departures of you favourite station on your homescreen.

## Open TODOs in the code
The idea was to select the station, which is shown in the widget by the current location. So ive added a setting to the database and the parameter to the function calls. As this is not in the scope of this task i only commented where to do what. Should this be deleted or opened in a new issue?


build.gradle
```diff
@@ -9,7 +9,7 @@ buildscript {
         }
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.3.0'
+        classpath 'com.android.tools.build:gradle:2.3.2'
```
sebmarste: should i revert this?

app/src/main/res/values/strings.xml
```diff
@@ -499,6 +499,9 @@ Signature: %5$s</string>
 
     <string name="transport_action_usage">Usage</string>
     <string name="transport_help_text">Tap your favorite transportation lines to highlight them. Tap again to revert them back to normal.</string>
+    <string name="mvv_widget" translatable="false">MVV Widget TUM</string>
+    <string name="mvv_widget_select_station">Select station</string>
+    <string name="mvv_widget_no_station">No station selected</string>
```
sebmarste: how do you get good translations for other languages?

app/src/main/java/de/tum/in/tumcampusapp/widgets/MVVWidget.java
```diff
@@ -0,0 +1,146 @@
...
+        TransportManager transportManager = new TransportManager(context);
+        Cursor c = transportManager.getWidget(appWidgetId);
+        if (c.getCount() >= 1) {
+            c.moveToFirst();
+            station = c.getString(c.getColumnIndex("station"));
+            station_id = c.getString(c.getColumnIndex("station_id"));
+            Boolean use_location = c.getInt(c.getColumnIndex("location")) != 0;
+            c.close();
+            if (use_location) {
+                // TODO implement nearest station (replace the station_id string with the calculated station)
```
pfent: What's the plan for this feature?

As I see this, we would need to add another endpoint in the TransportManager, i.e. a new `XML_STOPFINDER_REQUEST` with the current coordinates:

Maybe similar to this?:
http://efa.vrr.de/vrr/XML_STOPFINDER_REQUEST?language=de&type_sf=coord&name_sf=3435366.4594533667:463680.7198482931:NAV3:5

app/src/main/AndroidManifest.xml
```diff
@@ -511,15 +511,45 @@
...
+            </intent-filter>
+
+            <meta-data
+                android:name="android.appwidget.provider"
+                android:resource="@xml/mvv_widget_info" />
+        </receiver>
+
         <service
-            android:name=".widgets.CardsWidgetService"
+            android:name=".widgets.MVVWidgetService"
```
sebmarste: this is just wrong behaviour in git... ive added the service for the mvv widget, the card wiged stays untouched and is behind the mvv widget in the manifest file

app/src/main/AndroidManifest.xml
```diff
@@ -530,6 +562,10 @@
                 android:resource="@xml/cards_widget_info" />
         </receiver>
 
+        <service
+            android:name=".widgets.CardsWidgetService"
+            android:permission="android.permission.BIND_REMOTEVIEWS" />
```
sebmarste: here we do have the cards service

app/src/main/res/values/strings.xml
sebmarste: so the is german and english fine for now or should i use the translators magic?

build.gradle
sebmarste: ok, noted ;)

app/src/main/java/de/tum/in/tumcampusapp/widgets/MVVWidget.java
sebmarste: exactly, something like that was my idea, but i did not want to put time in this without talking to anyone

app/src/main/java/de/tum/in/tumcampusapp/widgets/MVVWidget.java
```diff
@@ -0,0 +1,146 @@
...
+     * If no alarm is running yet a new repeating alarm is started
+     */
+    private static void setAlarm(Context context) {
+        if (alarmIsSet) return;
+        AlarmManager am = (AlarmManager) context.getSystemService(Context.ALARM_SERVICE);
+        Intent intent = new Intent(context, MVVWidget.class);
+        intent.setAction(BROADCAST_ALARM_NAME);
+        PendingIntent pi = PendingIntent.getBroadcast(context, 0, intent, 0);
+        //After after 30 or 60 (android may delay) seconds
+        am.setRepeating(AlarmManager.RTC, System.currentTimeMillis(), 30000, pi);
```
sebmarste: So this is a litte bit tricky
1) the usual onUpdate method of a widget is called every hour, at most, so this does not work for our usecase
2) using an alarm (as i do) has a min interval with about one minute (but can be delayed) and is not called when the device is not unlocked (which we want)
3) in every alarm step i use a time (which is exact) to update the widget after 30s and 60s to prevent late alarm steps -> at least every minute is one update (only with unlocked device)
4) i do download data only every second minute (at most) to reduce data and battery usage, between that the information is cached and the departure times are calculated
So overall we have downloadInterval, alarmInterval and 2x timerInterval to be configured... what would you suppose?

app/src/main/java/de/tum/in/tumcampusapp/widgets/MVVWidget.java
```diff
@@ -0,0 +1,146 @@
...
+        String station_id = "";
+
+        // Get the settings for this widget from the database
+        TransportManager transportManager = new TransportManager(context);
+        Cursor c = transportManager.getWidget(appWidgetId);
+        if (c.getCount() >= 1) {
+            c.moveToFirst();
+            station = c.getString(c.getColumnIndex("station"));
+            station_id = c.getString(c.getColumnIndex("station_id"));
+            Boolean use_location = c.getInt(c.getColumnIndex("location")) != 0;
```
sebmarste: boxing? i just wanted to have it in there and i think this is the way of getting booleans from the sqlite db, right?

app/src/main/res/values/strings.xml
kordianbruck: Fine for now.

app/src/main/java/de/tum/in/tumcampusapp/widgets/MVVWidget.java
pfent: boxing is the difference between a primitive `boolean` (lowercase) and a `Boolean` object. I think you can safely switch to the primitive type here

app/src/main/java/de/tum/in/tumcampusapp/widgets/MVVWidget.java
pfent: Thanks for that detailed explanation!
I just think that a named constant would be better readable than an inline constant
```java
private static final int UPDATE_DELAY = 30 * 1000;
```

sebmarste: Great, done

app/src/main/java/de/tum/in/tumcampusapp/widgets/MVVWidget.java
sebmarste: okay, thank you :) done as well

kordianbruck:
@sebmarste ready to merge? If it ain't, please add the work in progress label.

sebmarste:
@kordianbruck could you then, as we said, merge this branch please
ill do the functionality to to disable the automatic reload the in a new branch

Issue #368, opened 2016-06-08
Add PDF Viewer to App


kordianbruck:
As suggested by pfent in https://github.com/TCA-Team/TumCampusApp/pull/367 we should add a PDF viewer for the PlansActivity to directly display all the plans in the App. 

"Also, an inline display for the PDFs would be nice to have. There is a [PDFRenderer](https://developer.android.com/reference/android/graphics/pdf/PdfRenderer.html) available for API 21+, even with an exhaustive [example on GitHub](https://github.com/googlesamples/android-PdfRendererBasic). Then you can use the fileprovider as a fallback for lower APIs."

t2m4k1:
So after reading a little bit, the task seems to be to implement a fragment containing a pdf renderer and then depending on the client's api version either continue the way we currently do it or redirect the newly downloaded file to the pdf-fragment.

kordianbruck:
@t2m4k1 exactly! Maybe you can even find a lightweight solution for older android versions.

t2m4k1:
Ok I'll see what I can do. Maybe a simple integrated webview is supported by older android versions. Would that be sufficient? Btw. which android version are we referring to as the minimum supported? API16?

kordianbruck:
min is 24 currently: https://github.com/TCA-Team/TumCampusApp/blob/master/app/build.gradle

t2m4k1:
These issues seem to be fixed in:
https://stackoverflow.com/questions/2456344/render-a-pdf-file-using-java-on-android/21383356#21383356

t2m4k1:
For documents stored online like the plans' pdfs I've had success in opening them by using a url which redirects to google-drive pdf viewer. For this purpose its not even necessary to use any external libraries. Is that sufficient for now?
```        webView.loadUrl("https://docs.google.com/gview?embedded=true&url=" + path);
```
Also for the fragments to work, I'm currently turning the plans activity into two separate fragments (list and pdf).

kordianbruck:
Can you zoom and pan around on that webview then? I mean I'm all in for nice and easy solutions!

kordianbruck:
Done

PR #463, opened 2017-05-27
Added PdfViewer Fragment and refactored PlansActivity

t2m4k1:
## Issue

This fixes the following issue(s): #368 

## Screenshot
widescreen
portrait
zoomedwide

 

No need to leave the app anymore to view Pdfs,
backwards compatibility way beyond minimum api level,
reusability of pdf-viewfragment for other purposes


t2m4k1:
The change in Base activity basically only makes sure that the back button on the blue menu bar doesn't completely close the plans activity and instead just switches back to the list view fragment. I tried it with different activities and it didn't seem to have any negative side effects. In other words I think it's more intuitive to use a back arrow to go back to the previously seen fragment/activity instead of ignoring fragments and just going backwards in terms of activities.

app/src/main/java/de/tum/in/tumcampusapp/activities/PlansActivity.java
```diff
@@ -1,140 +1,27 @@
...
-public class PlansActivity extends BaseActivity implements OnItemClickListener {
-
-    private static final String[][] FILES_TO_DOWNLOAD = {
-            {"http://www.mvv-muenchen.de/fileadmin/media/Dateien/plaene/pdf/Netz_2016_Version_MVG.PDF", "Schnellbahnnetz.pdf"},
-            {"http://www.mvv-muenchen.de/fileadmin/media/Dateien/plaene/pdf/Nachtnetz_2016.pdf", "Nachtliniennetz.pdf"},
-            {"http://www.mvv-muenchen.de/fileadmin/media/Dateien/plaene/pdf/Tramnetz_2016.pdf", "Tramnetz.pdf"},
-            {"http://www.mvv-muenchen.de/fileadmin/media/Dateien/3_Tickets_Preise/dokumente/TARIFPLAN_2016-Innenraum.pdf", "Tarifplan.pdf"},
-    };
-    ProgressBar progressBar;
-    private final Thread downloadFiles = new Thread() {
```
kordianbruck: Why does it make more sense to do the download in the fragment?

app/src/main/java/de/tum/in/tumcampusapp/fragments/PdfViewFragment.java
```diff
@@ -0,0 +1,43 @@
...
+    private File pdf;
+    @Override
+    public View onCreateView(LayoutInflater inflater, ViewGroup container,
+                             Bundle savedInstanceState) {
+        this.inflater = inflater;
+        View view = this.inflater.inflate(R.layout.fragment_pdf_view, container, false);
+        pdfView = (PDFView) view.findViewById(R.id.pdf_view);
+        pdfView.addOnLayoutChangeListener(new View.OnLayoutChangeListener() {
+            @Override
+            public void onLayoutChange(View view, int i, int i1, int i2, int i3, int i4, int i5, int i6, int i7) {
```
kordianbruck: Why is the naming of the variables so bad? What are all those ints passed to the function?

app/src/main/java/de/tum/in/tumcampusapp/fragments/PdfViewFragment.java
```diff
@@ -0,0 +1,43 @@
...
+
+import de.tum.in.tumcampusapp.R;
+
+public class PdfViewFragment extends Fragment {
+    private LayoutInflater inflater;
+    private PDFView pdfView;
+    private File pdf;
+    @Override
+    public View onCreateView(LayoutInflater inflater, ViewGroup container,
+                             Bundle savedInstanceState) {
```
kordianbruck: Why is there a new line here?

app/src/main/java/de/tum/in/tumcampusapp/activities/PlansActivity.java
t2m4k1: Since i currently disable the listview after the download finished and also set the progressbar as invisible, i thought it makes sense to put these things together. If you prefer the download in the activity itself, I can try to make it work!

kordianbruck: Nah, thats all good - just wanted to know the thinking behind it :+1:

app/src/main/java/de/tum/in/tumcampusapp/fragments/PdfViewFragment.java
```diff
@@ -37,7 +40,32 @@ public void onResume() {
...
-        this.pdf = pdf;
+    public boolean setPdf(File pdf){
+        if (isFilePdf(pdf)){
+            this.pdf = pdf;
+            return true;
+        }
+        return false;
+    }
+
+    private boolean isFilePdf(File file){
```
kordianbruck: Please document what this function is doing. Why is it in this class? Why not in an auxiliary helper?

t2m4k1: E.g. if we receive a 404 from the mvv website, it is still downloaded as html and therefore it's necessary to check if we're trying to open a valid pdf (first bytes starting with %PDF). From my experiments the PdfView's implementation would just be blank, if it tries to open a non-pdf file.
Therefore I added this method as an additional check before opening a file. If you prefer that I redesign the code to use a custom "FileIsNotPdfException", I can do so. In my opinion our only purpose of checking whether a file is a valid pdf or not, is to open it. And because this responsibility only lies in the PdfViewerFragment, I think it's in a valid spot there. Let me know if that fits your idea.

pfent: @t2m4k1 Solid research 👍 

But please document that in the function javadoc, so that knowledge doesn't get lost.

Also: please use the `Utils.log()` methods instead of `e.printStackTrace()` or debugging becomes hell

app/src/main/java/de/tum/in/tumcampusapp/fragments/PlansViewFragment.java
```diff
@@ -0,0 +1,197 @@
...
+import de.tum.in.tumcampusapp.activities.PlansDetailsActivity;
+import de.tum.in.tumcampusapp.adapters.PlanListAdapter;
+import de.tum.in.tumcampusapp.adapters.PlanListAdapter.PlanListEntry;
+import de.tum.in.tumcampusapp.auxiliary.NetUtils;
+import de.tum.in.tumcampusapp.auxiliary.Utils;
+
+
+/**
+ * Created by Thomas-Notebook on 14.05.2017.
+ */
```
pfent: Actual comments would be appreciated

kordianbruck:
@t2m4k1 please merge when you are ready. You got my go.

t2m4k1:
It says I'm not authorized to merge this pull request:
"This pull request can be automatically merged by project collaborators
Only those with write access to this repository can merge pull requests."
Did I forget to do something?

kordianbruck:
Sorry, my bad. I think I fixed it now... can try again with the next PR.

Good job!

PR #466, opened 2017-05-30
Use more kotlin

kordianbruck:
Yea, just testing around with this.


pfent:
I'd generally wait for Android Studio 3.0 to be released before using Kotlin for "production" use, we can never know what can change when this is officially supported.

kordianbruck:
Absolutely right. Just was playing around with it yesterday at the Jetbrains Meetup and wanted to start somewhere.

kordianbruck:
meh, will start again, once AS 3.0 comes out.

PR #469, opened 2017-06-02
Mvv widget auto reload

sebmarste:
## Issue
added mode where you have to press the reload button to get new data (default)
added auto mode where the data gets updated automatically (can be selected in configure activity)

### switch to select reload mode
screenshot_20170608-143425
### reload button in widget
screenshot_20170608-143438


app/src/main/java/de/tum/in/tumcampusapp/widgets/MVVWidgetService.java
```diff
@@ -27,10 +27,12 @@ public RemoteViewsFactory onGetViewFactory(Intent intent) {
         private final Context applicationContext;
         private List<TransportManager.Departure> departures = new ArrayList<>();
         private int appWidgetID;
+        private boolean forceLoadDepartures = false;
 
         MVVRemoteViewFactory(Context applicationContext, Intent intent) {
             this.applicationContext = applicationContext.getApplicationContext();
             this.appWidgetID = intent.getIntExtra(AppWidgetManager.EXTRA_APPWIDGET_ID, -1);
+            this.forceLoadDepartures = intent.getBooleanExtra(MVVWidget.MVV_WIDGET_FORCE_RELOAD, false);
```
kordianbruck: As I do not know this: when is this contructor called? Does it rerun once the setting has been updated?

kordianbruck: Its done by the broadcast above right?

sebmarste: the Factory gets instantiated by the intend in MVVWidgets updateWidget method. So every time the widged is updated a new factory is used

app/src/main/java/de/tum/in/tumcampusapp/managers/TransportManager.java
```diff
@@ -126,10 +126,10 @@ public TransportManager(Context context) {
                 "id INTEGER PRIMARY KEY AUTOINCREMENT, symbol VARCHAR)");
 
         db.execSQL("CREATE TABLE IF NOT EXISTS transport_widgets (" +
```
kordianbruck: A better name would be `widgets_transport`. (Optional)

app/src/main/java/de/tum/in/tumcampusapp/managers/TransportManager.java
```diff
@@ -196,26 +193,21 @@ public void deleteWidget(int widget_id) {
...
-            if (use_location) {
-                // TODO implement nearest station (replace the station_id string with the calculated station)
-                station = "use location";
-            }
         }
-        WidgetDepartures wD = new WidgetDepartures(station, station_id);
-        TransportManager.widgetDepartures.put(widget_id, wD);
-        return wD;
+        TransportManager.widgetDeparturesList.put(widget_id, widgetDepartures);
+        return widgetDepartures;
```
kordianbruck: Thanks for properly naming these variables.

app/src/main/java/de/tum/in/tumcampusapp/managers/TransportManager.java
```diff
@@ -407,20 +399,42 @@ public StationResult(String station, String id, int quality) {
...
-        String station;
-        String station_id;
-        List<Departure> departures;
-        long last_load;
-        boolean is_offline = false;
+
+        private String station;
+        private String station_id;
+        private boolean use_location;
+        private boolean auto_reload;
```
kordianbruck: Please use camelCase for all variables ;-)

app/src/main/java/de/tum/in/tumcampusapp/managers/TransportManager.java
```diff
@@ -407,20 +399,42 @@ public StationResult(String station, String id, int quality) {
     }
 
     public static class WidgetDepartures {
```
kordianbruck: Can you please put this into a separate file? Nested classes are nasty.

app/src/main/java/de/tum/in/tumcampusapp/widgets/MVVWidget.java
```diff
@@ -121,33 +125,36 @@ static void updateAppWidgets(Context context, AppWidgetManager appWidgetManager,
...
+        rv.setTextViewText(R.id.mvv_widget_station, widgetDepartures.getStation());
 
         // Set up offline symbol (may be shown one update delayed)
         rv.setViewVisibility(R.id.mvv_widget_offline, widgetDepartures.isOffline() ? View.VISIBLE : View.INVISIBLE);
 
         // Set up the configuration activity listeners
         Intent configIntent = new Intent(context, MVVWidgetConfigureActivity.class);
         configIntent.putExtra(AppWidgetManager.EXTRA_APPWIDGET_ID, appWidgetId);
-        PendingIntent pendingIntent = PendingIntent.getActivity(context, appWidgetId, configIntent, PendingIntent.FLAG_UPDATE_CURRENT);
+        PendingIntent pendingIntent = PendingIntent.getActivity(context, appWidgetId * 10, configIntent, PendingIntent.FLAG_UPDATE_CURRENT);
```
kordianbruck: Random constant somewhere in the code? BAAAAD :dagger:

app/src/main/java/de/tum/in/tumcampusapp/widgets/MVVWidget.java
```diff
@@ -121,33 +125,36 @@ static void updateAppWidgets(Context context, AppWidgetManager appWidgetManager,
...
         rv.setOnClickPendingIntent(R.id.mvv_widget_header, pendingIntent);
 
+        // Set up the reload functionality
+        Intent reloadIntent = new Intent(context, MVVWidget.class);
+        reloadIntent.setAction(MVVWidget.MVV_WIDGET_FORCE_RELOAD);
+        reloadIntent.putExtra(AppWidgetManager.EXTRA_APPWIDGET_ID, appWidgetId);
+        PendingIntent pendingReloadIntent = PendingIntent.getBroadcast(context, appWidgetId + 10 + 1, reloadIntent, PendingIntent.FLAG_UPDATE_CURRENT);
```
kordianbruck: Again. Why is it 10+1 now?

build.gradle
```diff
@@ -15,6 +15,8 @@ buildscript {
 
 allprojects {
     tasks.withType(JavaCompile) {
+        options.fork = true  // Fork your compilation into a child process
+        options.forkOptions.setMemoryMaximumSize("4g") // Set maximum memory to 4g
```
kordianbruck: Thanks, probably a good idea to have this on by default.

app/src/main/java/de/tum/in/tumcampusapp/widgets/MVVWidget.java
sebmarste: well actually this is an id for the intent, so it does not work, like expected, if there are some with the same id
the most simpel solution was to append an intend index to the widget id:
appWidgetID * 10       -> xxx0
appWidgetID * 10 + 1 -> xxx1

sebmarste: bin indeed it is not nice, any idea?

build.gradle
sebmarste: oh, i did not actively push this, but yea it does make sense...

app/src/main/java/de/tum/in/tumcampusapp/widgets/MVVWidget.java
kordianbruck: What do you mean by "if there are some with the same id"?

app/src/main/java/de/tum/in/tumcampusapp/managers/TransportManager.java
sebmarste: renamed 👍

app/src/main/java/de/tum/in/tumcampusapp/managers/TransportManager.java
sebmarste: ok i extracted all the nested classes from the transport manager to a model package called efa

app/src/main/java/de/tum/in/tumcampusapp/managers/TransportManager.java
sebmarste: im using camel case now

app/src/main/java/de/tum/in/tumcampusapp/widgets/MVVWidget.java
sebmarste: the pending intent to open the config activity and the one to reload, usually u take the appWidgetId there, but we have two intents

kordianbruck: You should use the same id and add a extra/bundle to pass the sub-action.

kordianbruck:
@sebmarste need to fix the ID thing here and resolve the string conflicts, then we can merge this as well

sebmarste:
shure, ill do this right after ive eaten 👍

app/src/main/java/de/tum/in/tumcampusapp/widgets/MVVWidget.java
sebmarste: seems like the request code does not even make any difference

sebmarste:
i think its ready to merge

PR #490, opened 2017-06-28
fix a bug where user may select items of the supper screen in sub screen

Yakuky:
## Issue

Fix a bug where user may select items of the supper screen in sub screen.


kordianbruck:
@Yakuky what exactly is this PR fixing? I'm not getting the problem that exists.

Yakuky:
@kordianbruck This PR fixes the problem I showed you in a previous email by the video. 

In preference page, when you use google talkback, you will select a supper-screen item in a sub-screen. This bug also happens when you use keyboard to select items in a sub preference screen.

kordianbruck:
Thanks!

PR #497, opened 2017-07-22
Implement Roomfinder with Retrofit

Yakuky:
## Issue


Yakuky:
@kordianbruck I am done with this.

Should I fix https://github.com/TCA-Team/TumCampusApp/issues/431 on this branch or create another one? 
Creating a new branch from master now will definitely cause merge conflict.

kordianbruck:
@Yakuky excellent! Thank you very much! <3

yea, please create a separate branch for that issue. Not sure if we can merge this right away.

app/src/main/java/de/tum/in/tumcampusapp/activities/RoomFinderActivity.java
```diff
@@ -3,39 +3,42 @@
...
-    // HTTP client for sending requests to MyTUM roomFinder
-    private TUMRoomFinderRequest roomFinderRequest;
+public class RoomFinderActivity extends ActivityForSearchingInBackground<List<RoomFinderRoom>>
+        implements OnItemClickListener {
 
     private RecentsManager recentsManager;
     private StickyListHeadersListView list;
     private RoomFinderListAdapter adapter;
 
+    private AsyncTask<String, Void, List<RoomFinderRoom>> searchRoomAsyncTask;
```
kordianbruck: This is not used anywhere or is it?

app/src/main/java/de/tum/in/tumcampusapp/activities/RoomFinderActivity.java
```diff
@@ -64,92 +66,79 @@ public void onCreate(Bundle savedInstanceState) {
...
-    public void onFetch(List<Map<String, String>> result) {
-        if (result.isEmpty()) {
+    protected void onSearchFinished(Optional<List<RoomFinderRoom>> result) {
+        if(!result.isPresent()){
+            if (NetUtils.isConnected(this)) {
+                showErrorLayout();
+            } else {
+                showNoInternetLayout();
+            }
+            return;
```
kordianbruck: This is not needed right?

app/src/main/java/de/tum/in/tumcampusapp/activities/RoomFinderActivity.java
```diff
@@ -64,92 +66,79 @@ public void onCreate(Bundle savedInstanceState) {
...
+        }
+
+        List<RoomFinderRoom> searchResult = result.get();
+        if (searchResult.isEmpty()) {
```
kordianbruck: Sure this list cannot be null? Why are you using an optional then? Maybe its just late and I'm not thinking clearly...

app/src/main/java/de/tum/in/tumcampusapp/activities/RoomFinderDetailsActivity.java
```diff
@@ -218,24 +191,125 @@ protected void onLoadFinished(Optional<File> result) {
...
+
+        return Optional.absent();
+    }
+
+    final void startLoadingMapList(String... params){
+        if (mapListAsyncTask != null) {
+            mapListAsyncTask.cancel(true);
+        }
+
+        mapListAsyncTask = new AsyncTask<String, Void, Optional<List<RoomFinderMap>>>() {
```
kordianbruck: You could just use Retrofit Callbacks / Asynchronous Requests for this - no?

https://futurestud.io/tutorials/retrofit-synchronous-and-asynchronous-requests

app/src/main/java/de/tum/in/tumcampusapp/api/Helper.java
```diff
@@ -102,6 +103,17 @@ public Response intercept(Chain chain) throws IOException {
...
     }
 
+    /**
+     * encodes an url
+     *
+     * @param pUrl input url
+     * @return encoded url
+     */
+    public static String encodeUrl(String pUrl) {
+        String url = pUrl.replace("/", ""); //remove slashes in queries as this breaks the url
```
kordianbruck: Wait. You just remove all slashes from a URL? Shouldn't escape be fine?

Yakuky: It party is only applied to the arch_id of a room.
And it is not necessary unless there is a slash in the arch_id of a room, which is not the case.

So this can be removed.

app/src/main/java/de/tum/in/tumcampusapp/activities/RoomFinderActivity.java
Yakuky: In line 84 I checks if the list `isPresent()` or not and if not the method directly returns so I think the list cannot be null.

app/src/main/java/de/tum/in/tumcampusapp/activities/RoomFinderActivity.java
Yakuky: It is not necessary but just showing different error message according to the internet access situation. We can use `showErrorLayout()` in both cases thought.

app/src/main/java/de/tum/in/tumcampusapp/activities/RoomFinderActivity.java
Yakuky: No it is not used. I declare this variable before changing the subclass from ActivityForSearching to ActivityForSearchingInBackground. 
It is should be deleted.

app/src/main/java/de/tum/in/tumcampusapp/activities/RoomFinderDetailsActivity.java
Yakuky: Got it. I will work on this.

app/src/main/java/de/tum/in/tumcampusapp/api/Helper.java
kordianbruck: Yea please do. Someone else might come along and think to use this function without understanding that this is some corner case in some class.

app/src/main/java/de/tum/in/tumcampusapp/activities/RoomFinderActivity.java
kordianbruck: Then its a case of just being too late.

app/src/main/java/de/tum/in/tumcampusapp/activities/RoomFinderActivity.java
kordianbruck: I mean just the `return;` statement ;)

app/src/main/java/de/tum/in/tumcampusapp/activities/RoomFinderDetailsActivity.java
kordianbruck: Thanks - this will make the code more readable. Plus: asynctasks aren't too great.

app/src/main/java/de/tum/in/tumcampusapp/activities/RoomFinderActivity.java
Yakuky: `return; ` is necessary because there is no result to display. Otherwise it will cause `NullPointerException` .

Yakuky:
@kordianbruck I have finished switching to Asynchronous Requests.

app/src/main/java/de/tum/in/tumcampusapp/activities/RoomFinderActivity.java
kordianbruck: How is that possible? Its a void function? It does not return anything?!

kordianbruck: Ahh, wait, I see. The function does not end there. Sorry my mistake.

app/src/main/java/de/tum/in/tumcampusapp/activities/RoomFinderDetailsActivity.java
```diff
@@ -16,38 +17,51 @@
...
 
-    private Bundle roomInfo;
+    private RoomFinderRoom room;
     private String mapId = "";
-    private List<Map<String, String>> mapsList;
+    private List<RoomFinderMap> mapsList;
     private boolean infoLoaded;
     private Fragment fragment;
 
+    private AsyncTask<String, Void, Optional<Geo>> geoASyncTask;
```
kordianbruck: This is not needed ;)

app/src/main/java/de/tum/in/tumcampusapp/activities/RoomFinderDetailsActivity.java
```diff
@@ -7,6 +7,7 @@
 import android.content.pm.PackageManager;
 import android.content.pm.ResolveInfo;
 import android.net.Uri;
+import android.os.AsyncTask;
```
kordianbruck: Not needed.

app/src/main/java/de/tum/in/tumcampusapp/activities/RoomFinderActivity.java
```diff
@@ -3,34 +3,35 @@
 import android.app.SearchManager;
 import android.content.Intent;
 import android.database.Cursor;
+import android.os.AsyncTask;
```
kordianbruck: Not needed.

Yakuky:
Now asyncs are all removed.

kordianbruck:
@pfent can you please double check this, just that I didn't overlook anything?

pfent:
Looks good to me. Usage of Optional seems good, so we don't get any NullPointerExceptions.

IMHO, the only thing that's debatable is if we really want the complexity of a `Optional<List>>`, instead of just returning an empty list and defaulting to a generic error layout. Since the handling is already implemented, I don't mind it too much.
@kordianbruck  go ahead and merge!

kordianbruck:
Thanks @Yakuky and @pfent - good work! :heart:

PR #503, opened 2017-07-25
Upgrade dependencies

pfent:
Upgrade to API 26 and fix the usual subsequent deprecation warnings

Testing with Android O devices / emulators for regressions would be much appreciated, but on my emulator everything seems to work fine.

@kordianbruck I changed the hash algorithm for the cache in 57d91fb, so we might wanna issue a cache cleaning on app upgrade

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/managers/EduroamManager.java
+++ b/app/src/main/java/de/tum/in/tumcampusapp/managers/EduroamManager.java
@@ -118,8 +118,7 @@ public class EduroamManager {
             return false;
         }
 
-        //Save, enable and exit
-        wifiManager.saveConfiguration();
+        //Enable and exit
         wifiManager.enableNetwork(networkId, true);
         return true;
     }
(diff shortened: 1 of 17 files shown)


app/src/main/java/de/tum/in/tumcampusapp/managers/EduroamManager.java
```diff
@@ -118,8 +118,7 @@ public boolean configureEduroam(String lrzId, String networkPass) {
             return false;
         }
 
-        //Save, enable and exit
-        wifiManager.saveConfiguration();
```
kordianbruck: Is this not needed?

pfent: Nope. The method's deprecated now.
https://developer.android.com/reference/android/net/wifi/WifiManager.html#saveConfiguration()
> This method was deprecated in API level 26.
> There is no need to call this method - addNetwork(WifiConfiguration), updateNetwork(WifiConfiguration) and removeNetwork(int) already persist the configurations automatically.

PR #507, opened 2017-09-17
Migrate GCM to firebase (FCM)

mammuth:
Guide I followed:
https://developers.google.com/cloud-messaging/android/android-migrate-fcm

I added `com.google.android.gms:play-services-base` in order to use `GoogleApiAvailability` which is used for checking whether PlayServices are installed on a couple of places. 

I got it building with this changes, but I didn't test any functionality. 👷 Help with testing whether I broke something is welcome.

ToDo:
- [ ] Test whether FCM is functional
- [ ] Understand why Jenkins is failing

### DIFF
--- a/app/.gitignore
+++ b/app/.gitignore
@@ -1,1 +1,2 @@
 /build
+google-services.json
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -84,3 +84,5 @@ dependencies {
     androidTestCompile "com.android.support.test.espresso:espresso-core:$espressoVersion"
     androidTestCompile "com.android.support.test.espresso:espresso-contrib:$espressoVersion"
 }
+
+apply plugin: 'com.google.gms.google-services'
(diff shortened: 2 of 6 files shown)


mammuth:
Strange finding: 
In the app level gradle.build the `applicationId` is set to `de.tum.in.tumcampus`. However, he java package and the packagename in the AndroidManifest are `de.tum.in.tumcampusapp`. 

I registered the app with `de.tum.in.tumcampus` on firebase.

Also it might be needed that you download the google-service.json file in order to build the app; this can be done at the [firebase console](https://console.firebase.google.com/project/tca-backend-0001/settings/general/android:de.tum.in.tumcampusapp); otherwise I can send it to you. Adding it to git is probably a bad idea 🙃 

I clicked a bit through the app and didn't come across any crashes.  We should still test, whether FCM is functional.

kordianbruck:
@mammuth yea, the package missmatch is due to the playstore not allowing to rename packages as they use it as the core ID.

As said: I think we need to adapt the function in the backend that connects to GCM and sends out notifications.

kordianbruck:
It was a POST request or something. We'll have to test this tomorrow.

PR #536, opened 2017-10-18
Reorganize the side drawer and remove OrganizationActivity

pfent:
* Removed the OrganisationActivity. Nobody I talked to actively used it.  
   (if we get complaints, we can still add it again :speak_no_evil:)
* Rearranged the order of the items, that appear in the side drawer;  
  though I'm open for feedback


pfent:
> One forgets that they ever existed

IMHO good riddance. I can't imagine a proper use-case for them.
We talked about linking from lectures to chairs in the organization tree, but I can't find a way to link them and the information provided in the OrganizationDetailsAcitvity is kinda sparse and thus useless anyways.

Also: I think you know my stance about having dead code in the repo... That's not what HEAD is for.
If you really want a reminder of that activity, please open an issue, where you can reference 3ee6f5309d519dac161e4296c82d22656314c027. This commit will be in our repo forever and the code doesn't get "lost" :heart:

kordianbruck:
@pfent mimimi - kay, convinced :heart:

Issue #526, opened 2017-10-16
'My Grades' showing error when there are no grades to show


jacqueline8711:
## Actual Result
"Error accessing TUMonline. Please make sure your TUMonline-Access-Token has been activated and the corresponding has been activated ..."

## Problem
Confusing for many users

PR #537, opened 2017-10-18
Don't show TUMOnline error when valid empty response is returned

jacqueline8711:
## Issue

This fixes the following issue(s): #526 

## Screenshot
no_grades_error

Students won't be confused by wrong error message

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/managers/TumManager.java
+++ b/app/src/main/java/de/tum/in/tumcampusapp/managers/TumManager.java
@@ -11,6 +11,7 @@ import java.util.Date;
 import de.tum.in.tumcampusapp.auxiliary.DateUtils;
 import de.tum.in.tumcampusapp.auxiliary.Utils;
 import de.tum.in.tumcampusapp.models.tumo.Error;
+import de.tum.in.tumcampusapp.tumonline.TUMOnlineRequest;
 
 /**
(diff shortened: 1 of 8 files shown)


app/src/main/java/de/tum/in/tumcampusapp/managers/TumManager.java
```diff
@@ -110,14 +110,16 @@ public String addLock(String url, String data) {
         //Try to parse the error
         String msg = "";
         try {
-
+            if(data.contains("<rowset>\n</rowset>")){
```
kordianbruck: You could use the constant from TUMOnlineRequest here or move it to the Constants file

jacqueline8711:
I changed it to use the constant from TUMOnlineRequest. But it's also good to know there is a constants file, I didn't see that before

app/src/main/java/de/tum/in/tumcampusapp/tumonline/TUMOnlineRequestFetchListener.java
```diff
@@ -34,4 +34,9 @@
 	 */
 	void onFetchError(String errorReason);
 
+	/**
+	 * Called if there was no error accessing tum online but there are no results to show (no entries)
+	 */
+	void onNoDataToShow();
```
pfent: Great idea, this is probably also useful for a couple of other Activities, where we currently just show generic error messages...

PR #538, opened 2017-10-19
Changed the photo_not_available and star image + deleted 3 unused images

jacqueline8711:
## Issue
The photo_not_available image wasn't really nice and definitely didn't fit into the round imageView in the menu.
The stars seemed kinda old-fashioned, so I changed those too. I changed the buttons from ToggleButton to a simple ImageView because the toggle button wasn't really made to deal with images and it works just as well like this.

## Screenshot
screenshot_20171019-214457
screenshot_20171019-220230
screenshot_20171019-220319

It just looks a bit nicer


pfent:
Those photo_not_available images look like they are generated from vector graphics...

Any chance you still have those vector graphics and instead of using PNGs, we can use SVGs and [import them to Vector Drawables](https://developer.android.com/studio/write/vector-asset-studio.html#svg) for better multi-DPI support?

jacqueline8711:
unfortunately I made those two images with a drawing program that can't create vector images

PR #541, opened 2017-10-20
Contacts permission info message + crash fix, Fixes #540

### LINKED ISSUE #540: "Lorem Ipsum" message + crash when app doesn't have contact permission
## Steps to Reproduce
1. go to Person Search
2. select/search for a person
3. click the 'add person to contacts' icon in the app bar
4. "Lorem Ipsum" dialog with option "ok" is shown if the app doesn't have contacts permission
5. app crashed when you click on "ok"

## Expected Result
A message showing the user that the contact permission is needed and allowing the user to grant the permission

jacqueline8711:
## Issue

This fixes the following issue(s): #540

## Screenshot
contacts_permission

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/activities/PersonsDetailsActivity.java
+++ b/app/src/main/java/de/tum/in/tumcampusapp/activities/PersonsDetailsActivity.java
@@ -407,7 +407,7 @@ public class PersonsDetailsActivity extends ActivityForAccessingTumOnline<Employ
      * @param id the request id
      * @return If the contacts permission was granted
      */
-    private boolean isPermissionGranted(int id) {
+    private boolean isPermissionGranted(final int id) {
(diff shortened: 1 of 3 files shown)


app/src/main/java/de/tum/in/tumcampusapp/activities/PersonsDetailsActivity.java
```diff
@@ -418,15 +418,16 @@ private boolean isPermissionGranted(int id) {
 
                 // Display an AlertDialog with an explanation and a button to trigger the request.
                 new AlertDialog.Builder(this)
-                        //.setMessage(getString(R.string.permission_calendar_explanation))
-                        .setMessage("Lorem Ipsum")
```
pfent: We still had a Lorem Ipsum in there? And git blame even tells me it was my commit... 🤦‍
Good catch

kordianbruck: tztztztzzt - better add a todo marker next time - just sayin'

Issue #533, opened 2017-10-18
MVV Results still wonky


pfent:
After an attempt to fix special characters (ß, like in "Theresienstraße") in #528, the our implementation now has problems with umlauts (e.g. ü, like in "München").

## Steps to Reproduce

1. Open MVVActivity
2. Search for München
3. There is no result

## Problem
Umlauts don't work right. Somehow searching for `M&uuml;nchen` works, though I'm not sure why, plus tapping on the recent stations view still gives an empty result :disappointed:

lnsp:
Most likely because the recent stations have been looked up using the old search.

![screenshot_1508506828](https://user-images.githubusercontent.com/3391295/31823704-3a4a2c4e-b5ad-11e7-8b06-8d7c0817929e.png)

When you search for "münchen", you get good results.

![screenshot_1508506957](https://user-images.githubusercontent.com/3391295/31823726-52cbd5ba-b5ad-11e7-89cf-29991e5334e8.png)

There is an issue using the standard stations though, for example there is only "München, Universität" now and no "Universität" station.

lnsp:
I think isolated the issue. When searching for a result and tap on a result, you query the MVV EFA using the station ID (a number like 1001928 for München, Manzostraße). If you use the "recent" list instead, it queries the server using the station name instead of the station ID.

pfent:
@lnsp thanks for the investigation, you were spot on! 👏 

Due to the RecentsManager only having a single text column, I now serialize the objects to JSON 🙈

pfent:
fixed in #542

PR #542, opened 2017-10-21
Fix the MVV results

pfent:
## Issue

This fixes the following issue(s): **#533**


kordianbruck:
@lnsp are these not deleted deleted upon upgrade? Probably there should be a try/catch around the gson part. @pfent

lnsp:
It seems like they get wiped on update (in the StartupActivity), but these errors should be catched regardless in case of corrupt data.

kordianbruck:
=> Please create an issue with the milestone if you can :heart:

PR #546, opened 2017-10-27
Migrate the project to Java8

pfent:
This was made possible with the introduction of AndroidStudio 3.0 and the new `desugar` process

I mostly auto-converted a lot of the code, e.g: 
* using try-with-resources for all Coursors / StreamBuffers etc.
* converting all anonymous inner classes to lambdas / method references

Also, since this already conflicted with all other branches, i took the freedom to include reformatting all files in this PR. @kordianbruck don't already get nervous about that reformatting, I put that all in a nice separate commit, so you don't even need to look at it when reviewing :)


kordianbruck:
Thanks a lot! Way overdue to finally have Java8 Language Features.

PR #547, opened 2017-10-28
Use moar Kotlin

pfent:
Followup from #466

Converted almost all the models to kotlin data classes. Maybe take a look at the  `CafeteriaMenu` to see why Kotlin is much nicer there.
The Kotlin code is also much more concise for those classes, around 2k less LOC 🎉 

However, I noticed, that simplexml seemed to have some problems with those classes, so this probably needs to be tested to check, if some requests fail


app/src/main/java/de/tum/in/tumcampusapp/models/cafeteria/Cafeteria.kt
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * new Cafeteria
+ *
+ * @param id        Cafeteria ID, e.g. 412
+ * @param name      Name, e.g. MensaX
+ * @param address   Address, e.g. Boltzmannstr. 3
+ * @param latitude  Coordinates of the cafeteria
+ * @param longitude Coordinates of the cafeteria
+ */
+data class Cafeteria(val id: Int, val name: String, val address: String, val latitude: Double, val longitude: Double) : Comparable<Cafeteria> {
```
kordianbruck: Maybe it would be nicer to have one attribute per line?

kordianbruck:
Can you also check why Jenkins is not liking your kotlin code? Do we need to update the SDK again on the server probably right?

pfent:
hm. upgrading the jenkins plugins... maybe that solves it

SDK should now come from googles maven server, so that shouldn't be a problem anymore. A mismatched SDK would also cause build failures on the main branch...

pfent:
Meh. Can't get Jenkins to like this PR, It seems the gradle daemon dies each time this build gets triggered.

Building the project in a separate directory on the server works as expected.
I tried:
* rebooting the server
* fiddling with the gradle config to disable the daemon

@kordianbruck maybe you wanna have a try at it?

kordianbruck:
Let's just merge. I'll take care of jenkins some other time.

Issue #412, opened 2016-10-19
Let user choose their default campus in a card


pfent:
Probably avoids a lot of confusion about the default campus and is probably a pretty good guess

kordianbruck:
Faculty has been removed, so this is no longer possible. I would vouch for a selection in the mensa card, the first time it is shown.

foreverigor:
I'd say this makes sense. But isn't the Campus also used in public transport navigation?

kordianbruck:
@foreverigor yes, it is used in the mvv card/widget/activity - it should really stop us tho?

kordianbruck:
One can now choose their mensa that should be displayed. So that should fix this as well, as most people just wanted a specific mensa.

PR #548, opened 2017-10-29
Set the default Campus based on user's faculty

foreverigor:
Implemented a method, which selects the default Campus based on the user's selected faculty and writes it to the settings.

## Issue

#412 "Set the default campus accoding to the users entered faculty" 
https://github.com/TCA-Team/TumCampusApp/issues/412

Automatic selection means the users won't have to go to the settings to choose their default campus.

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/activities/wizard/WizNavStartActivity.java
+++ b/app/src/main/java/de/tum/in/tumcampusapp/activities/wizard/WizNavStartActivity.java
@@ -104,6 +104,7 @@ public class WizNavStartActivity extends ActivityForLoadingInBackground<String, 
                         try (Cursor c = sm.getFacultyID((String) adapterView.getItemAtPosition(i))) {
                             if (c.moveToFirst()) {
                                 Utils.setInternalSetting(getApplicationContext(), "user_major", c.getString(c.getColumnIndex("faculty"))); // save faculty number in shared preferences
+                                setDefaultCampus( c.getString(c.getColumnIndex("faculty")) );
                                 Utils.setInternalSetting(getApplicationContext(), "user_faculty_number", String.valueOf(userMajorSpinner.getSelectedItemPosition())); // save choosen spinner poistion so that in case the user returns from the  WizNavCheckTokenActivity to WizNavStart activity, then we the faculty gets autm. choosen.
                             }
                         }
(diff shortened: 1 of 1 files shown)


kordianbruck:

pfent:
ok to test

app/src/main/java/de/tum/in/tumcampusapp/activities/wizard/WizNavStartActivity.java
```diff
@@ -196,6 +197,72 @@ public void onClick(DialogInterface dialog, int which) {
         }
     }
 
+    private void setDefaultCampus ( String faculty_number ) {
+
+        String Campus = "0";
+        switch (faculty_number)
+        {
```
pfent: Could you please run AndroidStudio's auto-formatter ("Code -> Reformat Code") for this?

pfent:
Seems good. Thanks 👍

app/src/main/java/de/tum/in/tumcampusapp/activities/wizard/WizNavStartActivity.java
foreverigor: Ok done, but almost nothing changed – https://github.com/foreverigor/TumCampusApp/commit/0c805282d31b5a02ab588b046149bbc452418da8

The project is built with -Werror so that's why the switch statement looks so clumsy (case fall-through is not allowed)

pfent: Fallthrough is certainly allowed. I don't think -Werror / javas linting even detects this.
I think we already have a couple of fallthroughs in our code, e.g. see the [CardsAdapter](https://github.com/TCA-Team/TumCampusApp/blob/8219d041c7e5f525f74ec32db5d644b0e3c26b19/app/src/main/java/de/tum/in/tumcampusapp/adapters/CardsAdapter.java#L45)

foreverigor: Dann schau ich mir das nochmals an, warum das bei mir nicht funktionierte.

foreverigor:
Okay, that's how it was originally with the fall-through - should I open another pull request ?

https://github.com/foreverigor/TumCampusApp/commit/8377b08536e56e0b5d5537ae01bb14901b9a80c8

pfent:
sure, just open another PR, that's the most convenient way to merge your code.

while you're at it, please use a proper String comparison with `.equals` instead of:
```java
Campus != "0"
```
and please remove that `return` as the last statement in a void function

Issue #552, opened 2017-11-02
Use TUM id endpoint to receive person details


kordianbruck:
The endpoint webservicesbasic.id now also returns the id for the current user:
```
<?xml version="1.0" encoding="utf-8"?>
<rowset>
<row>
    <kennung>ga97lod</kennung>
    <vorname>Kordian</vorname>
    <familienname>Bruck</familienname>
    <obfuscated_id>6*D0780F2xxx</obfuscated_id>
    <obfuscated_ids>
        <student>5*F788EDBExxx</student>
        <bedienstete>3*F84C76D1xxx</bedienstete>
        <extern>6*D0780F2xxx</extern>
    </obfuscated_ids>
</row>
</rowset>
```

we can then use the webservicesbasic.personDetails to receive the users image

PR #553, opened 2017-11-02
Use proper endpoint to determine users tumonline ids

kordianbruck:
## Issue

This fixes the following issue(s): #552 

Works for everyone - yay.

## TODO:

- [ ] Let user select which role he currently has
- [ ] Use the according ID as the main id

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/activities/wizard/WizNavCheckTokenActivity.java
+++ b/app/src/main/java/de/tum/in/tumcampusapp/activities/wizard/WizNavCheckTokenActivity.java
@@ -2,14 +2,11 @@ package de.tum.in.tumcampusapp.activities.wizard;
 
 import android.content.Intent;
 import android.os.Bundle;
-import android.support.annotation.Nullable;
 import android.text.method.LinkMovementMethod;
 import android.view.View;
 import android.widget.TextView;
 
 import com.google.common.base.Optional;
-
-import java.util.List;
 
 import de.tum.in.tumcampusapp.R;
 import de.tum.in.tumcampusapp.activities.generic.ActivityForLoadingInBackground;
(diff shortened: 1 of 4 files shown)


app/src/main/java/de/tum/in/tumcampusapp/activities/wizard/WizNavCheckTokenActivity.java
```diff
@@ -87,22 +88,26 @@ protected Integer onLoadInBackground(Void... arg) {
...
+                    .toString());
 
             // Save the TUMOnline id to preferences
-            String pID = getUserPIdentNr(id.get()
-                                           .toString());
-            if (pID != null) {
-                Utils.setSetting(this, Const.TUMO_PIDENT_NR, pID);
-            }
+            Utils.setSetting(this, Const.TUMO_PIDENT_NR, identity.getObfuscatd_id());
+            Utils.setSetting(this, Const.TUMO_ST_NR, identity.getObfuscatd_ids().getStudierende());
```
pfent: Could we please not use those cryptic TUMonline identifiers, but readable names?

app/src/main/java/de/tum/in/tumcampusapp/activities/wizard/WizNavCheckTokenActivity.java
```diff
@@ -132,35 +137,4 @@ protected void onStart() {
...
-
-    /**
-     * Get the user's pident nr to identify him a little bit more. This allows information equal to
-     * the PersonDetailsActivity
-     *
-     * @param name The users full name
-     * @return the users pID, or null
-     */
-    @Nullable
-    private String getUserPIdentNr(String name) {
```
pfent: good riddance!

kordianbruck:
We still need the user to be able to select his profile. Student, Employee or Alumni and based on that select the correct ID. The currently proposed ID is randomly chosen from the three. In the future this should be the currently main / active ID, but this is not yet implemented as it was kind of complicated. 

We could couple this with the Cafeteria Pricing that also allows to chose which role one has.

pfent:
Meh. Since we advertise our app as "for and from students", I'd just default to the student ID. It's just for the picture anyhow....

Also: When you're not finished with a pull request, please use the "Work in progress" label 🐤

PR #555, opened 2017-11-08
Pfent/fix554

### LINKED ISSUE #554: Access Token configured properly, "Identity could not be accessed"
This is not a bug I can reproduce on different phones but an issue that keeps me from using the app at all, so hopefully this is the right place to submit it ;)
Everything worked fine on the same phone before, then I reset it. I also applied the latest Android security patch (Oxygen 3.6.0 -> 3.6.1, Krack fixed) before installing it, but this looks more like a problem on TUM Online's side (at least I tried it ~10 times by now with the same result every time):

## Steps to ~~Re~~produce

1. Installed the App via Google Play
2. Started App, Startup Wizard is shown
3. Entered my LRZ ID, faculty informatics, tapped "Next"
4. In the browser, checked all five token rights, activated token
5. Tried to leave setup step 2 by tapping "Next" again

## Expected Result
The next step

## Actual Result
"Identity could not be accessed. Please check the rights of your TUMonline token" (Toast)

## Problem
When skipping this step and tapping "Go to TUM campus app" the following toast appears:
"Something is wrong with your chat registration", independent which boxes I check.

### Environment
* Phone: OnePlus 2
* OS version: Oxygen OS 3.6.1 (Android v. 6.0.1)
* TUM Campus App version: v1.4.9
* Language: English

pfent:
Fixes #554 

Git blames @kordianbruck in commit 2bfd63eeb1f401b8ff339a1ee70477946f490fd1 for removing the default values for the Identity, thus removing the default constructor of the model and crashing simplexml. 🎉

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -5,7 +5,7 @@ apply from: '../pmd.gradle'
 
 android {
     compileSdkVersion 27
-    buildToolsVersion '27.0.0'
+    buildToolsVersion '27.0.1'
     buildTypes {
         all {
             proguardFiles(file('../proguard').listFiles())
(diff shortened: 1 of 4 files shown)


kordianbruck:
So basically only people that pulled the source code and build themselves had this problem? Cause we didn't push this to market yet.
Wow.

PR #557, opened 2017-11-08
Fix a couple of bugs, that happened with the kotlin conversion

pfent:
* Study Rooms didn't load, because the cursor for the adapter got autoclosed
* Some REST calls failed, because there were no default constructors for the models
* Employees without an image would crash, because the image would be null, where kotlin expected a non-null value


kordianbruck:
thanks for fixing this.

PR #566, opened 2017-11-09
Use roboelectric for testing

kordianbruck:
We really do need tests. The amount of bugs we have in our code is no longer tolerable.


kordianbruck:
@pfent any idea how to get jenkins to actually run the tests? There should be output in the console in that case.

kordianbruck:
I think once the build has ran through once, subsequent builds do not really rebuild for some reason.

pfent:
There should be a gradle target for the tests, e.g. `gradle test`
However UI tests, that need to be run on an emulator probably won't work. Last time I tested it they needed around 6GB RAM, which is way too much for our server...

kordianbruck:
Ive added the target to the config on jenkins. It still seems it cant find the source files for testing altough locally everything works

pfent:
hm, maybe the pull request builder is wonky again...

PR #577, opened 2017-11-11
Fix ValueRequiredException in startup wizard

### LINKED ISSUE #576: Can't setup tumonline account in startup wizard
## Steps to Reproduce

1. Fresh install
2. Create token and grant all rights in tumonline
3. Click next

## Expected Result

Get to next screen

## Actual Result

Error: Identity could not be accessed. Please check the rights of your TUMonline token

## Problem

I can't get past step 2 in the startup wizard

See [stacktrace.txt](https://github.com/TCA-Team/TumCampusApp/files/1461010/stacktrace.txt)

### Environment
* Phone: HTC 10
* OS version: Android 7.0
* TUM Campus App version: v1.5
* Language: English

screenshot_20171110-112407
screenshot_20171110-112421

anselm92:
Set all obfuscated id fields to not being required to prevent
a VRE in startup wizard when fields bedienstete and extern aren't
set in the response.

Fixes #576

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/models/tumo/ObfuscatedIds.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/models/tumo/ObfuscatedIds.kt
@@ -4,6 +4,6 @@ import org.simpleframework.xml.Element
 import org.simpleframework.xml.Root
 
 @Root(name = "row")
-data class ObfuscatedIds(@field:Element var studierende: String = "",
-                         @field:Element var bedienstete: String = "",
-                         @field:Element var extern: String = "")
+data class ObfuscatedIds(@field:Element(required = false) var studierende: String = "",
+                         @field:Element(required = false) var bedienstete: String = "",
+                         @field:Element(required = false) var extern: String = "")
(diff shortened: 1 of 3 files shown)


app/src/main/java/de/tum/in/tumcampusapp/models/tumo/ObfuscatedIds.kt
```diff
@@ -4,6 +4,6 @@ import org.simpleframework.xml.Element
 import org.simpleframework.xml.Root
 
 @Root(name = "row")
```
kordianbruck: Why not just use `@Root(strict=false)` instead of adding it to each of the elements?

anselm92: This seemed to not have an effect. Exception was still thrown

kordianbruck: Wow you are right. https://stackoverflow.com/questions/40487944/parsing-xml-kotlin-android
@pfent this is really dumb.

@anselm92 could you write a unit test for this as well then? Just to make sure this won't break again in the future.

anselm92:
@kordianbruck Jenkins does not run the new test, if i run it locally it is executed though.

kordianbruck:
@anselm92 did you force push or something?

anselm92:
Yes to update my commit

kordianbruck:
@anselm92 any new commit to a branch will automatically update the Pullrequest. No need to force push.

PR #578, opened 2017-11-11
Add settings to navigation drawer

### LINKED ISSUE #575: Move Settings to Drawer
So they are reachable from everywhere

anselm92:
Fixes #575
screenshot_20171111-142248

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/auxiliary/DrawerMenuHelper.java
+++ b/app/src/main/java/de/tum/in/tumcampusapp/auxiliary/DrawerMenuHelper.java
@@ -27,6 +27,7 @@ import de.tum.in.tumcampusapp.activities.StudyRoomsActivity;
 import de.tum.in.tumcampusapp.activities.SurveyActivity;
 import de.tum.in.tumcampusapp.activities.TransportationActivity;
 import de.tum.in.tumcampusapp.activities.TuitionFeesActivity;
+import de.tum.in.tumcampusapp.activities.UserPreferencesActivity;
 
 public class DrawerMenuHelper implements NavigationView.OnNavigationItemSelectedListener {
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/de/tum/in/tumcampusapp/auxiliary/DrawerMenuHelper.java
```diff
@@ -94,6 +99,12 @@ public void populateMenu(Menu navigationMenu) {
                              .setIntent(new Intent(mContext, item.activity));
             }
         }
+        SubMenu appMenu = navigationMenu.addSubMenu("App");
```
kordianbruck: This needs to be translated using the Strings.xml.

pfent: I'd just leave the submenu unnamed...

kordianbruck: Just checked the google apps and others: neither use a submenu name, so we might as well leave it out then.

anselm92: I agree, i first tried it that way, however leaving the subname empty will create a nasty space between the divider and the settings submenu entry. 
Already googled to find a solution for that, unfortunately this seems to be possible only by inflating a menu.xml and using a group.
We could move the submenus into a xml file, inflate it and add the entries programmatically. What do you think?

kordianbruck: Is it not possible to directly add an entry to the menu without a submenu? 
If not, then a inflated .xml should be fine

pfent: I think you can just [add](https://developer.android.com/reference/android/view/Menu.html#add(int,%20int,%20int,%20java.lang.CharSequence)) the item with another groupid, which creates a divider

kordianbruck:
```
use NONE if an item should not be in a group.
```

anselm92:
@pfent you were right, worked by adding it directly to the navmenu

app/src/main/java/de/tum/in/tumcampusapp/auxiliary/DrawerMenuHelper.java
anselm92: @pfent you were right, worked by adding it directly to the navmenu

kordianbruck:
@anselm92 did you force push here as well? No need, I squash commits when I merge them.

anselm92:
@kordianbruck yes i did. ok i'll do that for future commits :)

PR #579, opened 2017-11-11
Redirect single person search result to activity

OliHalu:
## Issue

This fixes the following issue(s): 558 -Single results in PersonSearch should proceed to PersonDetails
Now if there's only one search result it will directly proceed to PersonDetails

Because its more comy

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/activities/PersonsSearchActivity.java
+++ b/app/src/main/java/de/tum/in/tumcampusapp/activities/PersonsSearchActivity.java
@@ -122,6 +122,15 @@ public class PersonsSearchActivity extends ActivityForSearchingTumOnline<PersonL
         requestFetch();
     }
 
+    private void proceedToPersonDetails(PersonList response )    {
+        lvPersons.setAdapter(null);
+        Bundle bundle = new Bundle();
+        bundle.putSerializable("personObject", response.getPersons().get(0));
+        Intent intent = new Intent(this, PersonsDetailsActivity.class);
+        intent.putExtras(bundle);
+        startActivity(intent);
+    }
+
     /**
      * Handles the XML response from TUMOnline by de-serializing the information
      * to model entities.
@@ -130,8 +139,10 @@ public class PersonsSearchActivity extends ActivityForSearchingTumOnline<PersonL
      */
     @Override
     public void onLoadFinished(PersonList response) {
-        if (response.getPersons() == null) {
+        if (response.getPersons() == null || response.getPersons().isEmpty()) {
             lvPersons.setAdapter(new NoResultsAdapter(this));
+        } else if (response.getPersons().size() == 1){
+            proceedToPersonDetails(response);
         } else {
             ListAdapter adapter = new ArrayAdapter<>(this, android.R.layout.simple_list_item_1, android.R.id.text1, response.getPersons());
             lvPersons.setAdapter(adapter);


.idea/codeStyleSettings.xml
```diff
@@ -1,264 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
kordianbruck: Why is this file deleted?

.idea/fileTemplates/includes/File Header.java
```diff
@@ -1,3 +0,0 @@
-/**
```
kordianbruck: Please don't delete those files

.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -1,115 +0,0 @@
-<component name="InspectionProjectProfileManager">
```
kordianbruck: Please don't delete those files

.idea/inspectionProfiles/profiles_settings.xml
```diff
@@ -1,7 +0,0 @@
-<component name="InspectionProjectProfileManager">
```
kordianbruck: Please don't delete those files

.idea/vcs.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="VcsDirectoryMappings">
-    <mapping directory="$PROJECT_DIR$" vcs="Git" />
+    <mapping directory="" vcs="Git" />
```
kordianbruck: ????

app/src/main/java/de/tum/in/tumcampusapp/activities/ChatRoomsActivity.java
```diff
@@ -43,7 +44,7 @@
  * This activity presents the chat rooms of user's
  * lectures using the TUMOnline web service
  */
-public class ChatRoomsActivity extends ActivityForLoadingInBackground<Void, Cursor> implements OnItemClickListener {
+public class ChatRoomsActivity extends ActivityForLoadingInBackground<Void, Cursor>  implements OnItemClickListener {
```
kordianbruck: Why are you adding a space here?

app/src/main/java/de/tum/in/tumcampusapp/activities/ChatRoomsActivity.java
```diff
@@ -63,6 +64,8 @@ public ChatRoomsActivity() {
     @Override
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
+        //create BaseActivity for Drawer/up button
+
```
kordianbruck: What is this?

app/src/main/java/de/tum/in/tumcampusapp/activities/PersonsSearchActivity.java
```diff
@@ -130,8 +129,21 @@ public void onStartSearch(String query) {
      */
     @Override
     public void onLoadFinished(PersonList response) {
-        if (response.getPersons() == null) {
+        //String "no search result" not shown
+        //change response.getPersons() == null to response.getPersons().isEmpty() or response.getPersons().size() == 0
```
kordianbruck: This "comment" makes no sense?

app/src/main/java/de/tum/in/tumcampusapp/activities/PersonsSearchActivity.java
```diff
@@ -130,8 +129,21 @@ public void onStartSearch(String query) {
      */
     @Override
     public void onLoadFinished(PersonList response) {
-        if (response.getPersons() == null) {
```
kordianbruck: You still need to catch null values. Otherwise it might through a NullPointerException

app/src/main/java/de/tum/in/tumcampusapp/activities/PersonsSearchActivity.java
```diff
@@ -130,8 +129,21 @@ public void onStartSearch(String query) {
...
+        //String "no search result" not shown
+        //change response.getPersons() == null to response.getPersons().isEmpty() or response.getPersons().size() == 0
+        if (response.getPersons().isEmpty()) {
             lvPersons.setAdapter(new NoResultsAdapter(this));
+        } else if (response.getPersons().size() == 1){
+            ListAdapter adapter = new ArrayAdapter<>(this, android.R.layout.simple_list_item_1, android.R.id.text1, response.getPersons());
+            lvPersons.setAdapter(adapter);
+            Person person = (Person) lvPersons.getItemAtPosition(0);
```
kordianbruck: Why are you adding the item to the layout and then directly moving to another activity? Couldn't you just directly open the new activity, without this detour?

app/src/main/java/de/tum/in/tumcampusapp/tumonline/TUMOnlineRequest.java
```diff
@@ -105,7 +105,7 @@ public void cancelRequest(boolean mayInterruptIfRunning) {
             backgroundTask.cancel(mayInterruptIfRunning);
         }
     }
-
+    //added Utils.setSetting(c, Const.TUMO_DISABLED, false);
```
kordianbruck: ????

.idea/codeStyleSettings.xml
OliHalu: Actually I haven't deleted anything at all, as far as I know.

app/src/main/java/de/tum/in/tumcampusapp/activities/PersonsSearchActivity.java
OliHalu: Ok, but then I don't get the "no search result" prompt if there is no result.

app/src/main/java/de/tum/in/tumcampusapp/tumonline/TUMOnlineRequest.java
OliHalu: Oh sorry, that was a (bad) try to get rid of the token identification bug to solve one of the chatroom issues

.idea/codeStyleSettings.xml
kordianbruck: Clearly the diff says differently :)

app/src/main/java/de/tum/in/tumcampusapp/activities/PersonsSearchActivity.java
kordianbruck: You can do both. `if(null or isEmpty()`

app/src/main/java/de/tum/in/tumcampusapp/tumonline/TUMOnlineRequest.java
kordianbruck: Ahhh, right. The diff catches all of those forgotten pieces of debugging attempts. 
(Also: theres a Debug button in your IDE for such cases :+1: )

app/src/main/java/de/tum/in/tumcampusapp/activities/PersonsSearchActivity.java
```diff
@@ -130,8 +129,21 @@ public void onStartSearch(String query) {
...
+            // store selected person ID in bundle to get in in StaffDetails
+            Bundle bundle = new Bundle();
+            bundle.putSerializable("personObject", person);
+            // show detailed information in new activity
+            Intent intent = new Intent(this, PersonsDetailsActivity.class);
+            intent.putExtras(bundle);
+            startActivity(intent);
```
pfent: This looks like copypasta'd 🍝  code to me.

I'd just reuse the onItemClick method here, but moving the duplicated code to a separate method would probably be better.

Issue #565, opened 2017-11-09
Fix TU Kino


kordianbruck:
Currently the activity is empty and shows no information. Check why, fix and write unit tests.

psukys:
@kordianbruck , [KinoManager's sync request](https://github.com/TCA-Team/TumCampusApp/blob/548ed15f30e83db6c3fe509dd1332d1874bf4d5a/app/src/main/java/de/tum/in/tumcampusapp/managers/KinoManager.java#L66) returns movies up until 2017 February, is this intended?

pfent:
@psukys This can't be. I see a movie for 14.11.2017 right on the start page.

The problem is rather, that the KinoActivity does not show anything when clicking on that card on the MainActivity.

kordianbruck:
@pfent there are two endpoints. TU Kino gets pull from the news endpoint plus there is a separate TU Kino endpoint. I think @psukys is talking about the latter right?

pfent:
Oh my... 🤦‍

I think the bug is, that we have two separate endpoints, that might be out of sync to begin with...

kordianbruck:
Yes they are out of sync, as the second endpoint uses the data from the news to fetch additional data from imdb and stuff. @psukys do you want to have a look at the backend?

psukys:
@kordianbruck sure, tho I'm still figuring out card's endpoint retrieval workflow

psukys:
Unfortunately, I wasn't able to trace the second movie endpoint, tho it certainly exists.
Database-wise, the newer movies are added to *news*, the old ones (~70 movies) are added to *kino* table and later [pruned](https://github.com/TCA-Team/TumCampusApp/blob/master/app/src/main/java/de/tum/in/tumcampusapp/managers/KinoManager.java#L48)

@kordianbruck can you point me into how/where the second endpoint is functioning?

pfent:
@psukys the second endpoint is probably referenced in the NewsManager

All other endpoints should be referenced in the [TumCabeClient](https://github.com/TCA-Team/TumCampusApp/blob/master/app/src/main/java/de/tum/in/tumcampusapp/api/TUMCabeClient.java#L58), but we never finished porting the News / Kino to retrofit...

kordianbruck:
No, it should be directly in the KinoManager or KinoDetailsActivity.

psukys:
@pfent @kordianbruck , so my initial guess on this is:

1. Kino cronjob is not running

[...]

PR #584, opened 2017-11-21
Kino Activity fixes and unit tests

psukys:
## Issue

- #565 no movies layout
- *unit* tests for Kino activity, manager.

## Screenshot

screenshot from 2017-11-21 14-57-04

Having a message is more verbose than having a blank page, when there are no movies.


psukys:
In general, no rooms layout is never used, thus just finding a nice way (through checking other activities that had a similar layout change) to implement and wrapping with tests, after reading [testing guide](https://developer.android.com/training/testing/fundamentals.html) should work

psukys:
@kordianbruck , @pfent , also question regarding accessibility - currently Kino can be accessed only through clicking the event, maybe it would be nice to have it in menu as well? 

Also, should **no movies layout** have return button with "TU Kino" title in top bar? , currently there's no top bar:
screenshot from 2017-11-21 14-57-04

kordianbruck:
@psukys yea, we opted not to have it in the menu, as its already cramped. If you think its a nice addition / useful, feel free to have it in there. 

Yes: it should definitively have the bar etc. Ideally you would use the generic Activities we have to display the correct layout.

psukys:
@kordianbruck currently menu has 17 items (which is sort of a lot as well), grouped by menu subgroups. Since Kino is sort of of a distinct functionality (not news cards that redirect to a webpage), I believe that it is menu-worthy.

kordianbruck:
@psukys go for it!

psukys:
Some notes on trying to write unit tests for KinoActivity:

1. `onCreate` creates `KinoManager` instance which has to be mocked (data supplier)
2. `Intent` allows only sending key-value pairs to KinoActivity
3. Adding an additional constructor, which provides `KinoManager` is a solution, but IMO a bit ugly.

[SO](https://stackoverflow.com/a/27559695) offers using [PowerMock](https://github.com/powermock/powermock), which I guess plays around with class resolution order.
@kordianbruck, @pfent is there a preferred path?

kordianbruck:
@psukys yes, mockito and its extension powermock are basically the industry standard. I think I already added the dependency for one of those. If not: go for it!

psukys:
@kordianbruck tried various attempts at mocking, yet failed every time. At this point, I do not even know  how to approach the problem.

pfent:
@psukys where exactly did you fail?

IMHO, you can just replay old data, which you need to extract once.

The only method in KinoManager, that seems like a bit of code is the `getAllFromDb()`, where you could return a new Matrixcursor with some prerecorded data...

psukys:
@pfent, I guess I'm not using {Mockito, PowerMock, Robolectric} the *right* way. I cannot mock KinoManager's constructor for KinoActivity, thus I cannot control the control-flow for testing.

I've as well checked for record-replay, but it seems that [Mockito doesn't use it](https://github.com/mockito/mockito/wiki/Mockito-vs-EasyMock#differences)

pfent:
Why can't you mock the KinoManager constructor? We can talk about concrete examples and problems, you might encounter, but please always include code examples, what you don't understand.

I did a quick google search, generally mocking constructors with arguments is [easily possible with Mockito](https://stackoverflow.com/a/13365710). I don't get where you don't use it the *right* way, honestly not even where you use it at all 😅

Also, you probably need to implement the record-replay of data yourself. E.g. create a sample dataset or even a static SQLite database file...

psukys:
@pfent , [I have it here](https://github.com/TCA-Team/TumCampusApp/pull/584/files#diff-83924f9b1400607d79eeeae600fbbf18R25). I think, that I use the same way, that is described in the SO answer, yet I have to as well use Robolectric for Activity compliance.

pfent:
Uhm. your `mockedDbCursor` does not contain any data and lies about that.
You can't just `when(mockedDbCursor.getCount()).thenReturn(1)`, that breaks the invariant of a Cursor...

You either need to do something like [this](https://github.com/jameskbride/grocery-reminder/blob/master/app/src/test/java/com/groceryreminder/views/reminders/RemindersActivityTest.java#L86), or as I suggested earlier just [create a dummy matrix cursor](https://stackoverflow.com/a/18290921) that always returns example data

psukys:
@pfent, for that specific unit test, I don't need data, just need to know that a specific layout is used. Will do more thorough unit tests, yet the problem stands - I cannot mock KinoManager.

pfent:
@psukys I was just trying to run your code, but it does not even compile, because you didn't include the gradle dependencies...

psukys:
@pfent, woops sorry for that, just pushed the `build.gradle`

app/src/test/java/de/tum/in/tumcampusapp/activities/KinoActivityTest.java
```diff
@@ -1,22 +1,58 @@
...
+    public void mainComponentDisplayedTest() throws Exception{
+        PowerMockito.when(mockedDbCursor.getCount()).thenReturn(1);
+        PowerMockito.when(mockedKinoManager.getAllFromDb()).thenReturn(mockedDbCursor);
+
+        PowerMockito.whenNew(KinoManager.class)
+                    .withArguments(Mockito.any())
+                    .thenReturn(mockedKinoManager);
+
+        kinoActivity = Robolectric.buildActivity(KinoActivity.class).create().start().get();
+        PowerMockito.verifyNew(KinoManager.class).withNoArguments();
```
pfent: There's your error.

KinoActivity creates an KinoManager with a Context argument, but you specified `withNoArguments()`

pfent: remove that and the test passes...

psukys: Thanks! I guess, I needed a fresh look into that

psukys:
@pfent, Alright, after fiddling even more around, Robolectric doesn't seem to compatible with PowerMock [1]. The workaround suggested is not compliant with KinoActivity, as we need Robolectric to call KinoActivity from Android SDK (as well verify if specific layout was used).

After a lot of confusion, I've tried as well debugging what the activity receives and `KinoManager` is not mocked at all. This blocks the activity's unit test (and all unit tests that need mocking imports with Robolectric's support)

I guess, what I can do in this case would just try to unittest other *Kino** parts, whichever are not affected by the problem.

[1] - https://github.com/robolectric/robolectric/issues/3549

app/src/test/java/de/tum/in/tumcampusapp/activities/KinoActivityTest.java
psukys: Seems that verifyNew needs to have a following method that specifies arguments (or lack of them)

pfent:
Mocking static methods should still work..?

What about adding a new:
```java
public static KinoManager getNewInstace(Context context) {
    return new KinoManager(context);
}
```
And then mock that instead of using `PowerMockito.whenNew()`

psukys:
A **big** maybe approach would be Robolectric's  shadowing, but I have to read more about it whether it's accomplishable.

psukys:
@kordianbruck @pfent using Robolectric's shadowing (previous comment's link) is viable. The KinoManager class shadow gets pushed and used within KinoActivity.

The problem withi this implementation is, that class's shadow does not have (or at least I haven't found a way) mocking interaction, thus I believe shadow cannot be manipulated directly. One way of resolving this - test case and class's shadow use some persistence ([ObjectBox?](https://github.com/objectbox/objectbox-java)) storage, such that test case sets the what shadow's method has to return.

What are your thoughts on this?

pfent:
Seems like you already got something working. Code examples for that would be nice, so maybe we can directly talk code.

I don't actually know about that persistency... Somewhere you always have to define, what the testcases should return, again code would probably highlights the missing pieces

kordianbruck:
@psukys I really don't get what you are trying to do. This code works just fine for me:
```
kinoActivity = Robolectric.buildActivity(KinoActivity.class).create().start().get();
assertThat(kinoActivity.findViewById(R.id.drawer_layout).getVisibility()).isEqualTo(View.VISIBLE);
```

No need for Mockito/PowerMock. The Context gets injected just fine and the KinoManager started. I really don't see the problem.

When are you actually going to start to implement the layout and stuff? Nothing apart from those five lines of "test" have been written so far?

psukys:
I'm sorry, @kordianbruck , but I don't know how to mock something that KinoManager returns via Context. So far, what I've found is mocking via PowerMock ([which does not *fully* comply with Robolectric](https://github.com/robolectric/robolectric/issues/3549)) and "mock" through Robolectric's shadows - tried to set up [ObjectBox](https://github.com/objectbox/objectbox-java) as middleware between test case and shadow class, but it project build fails (tried many variations for usage), finally threw it away and decided to use static variables in shadow class to be set from test case and returned from non static method.

psukys:
Test wraps up KinoActivity, by the following technique I'll write unit tests for Kino{Manager, Adapter} as well.

kordianbruck:
@psukys last try: we've just merged the room implementation by philipp for the kinoactivity. Can you check if this changed anything?

kordianbruck:
@psukys I already merged the master - you just need to fix one little thing in the KinoActivity.

psukys:
@kordianbruck , @pfent how is [TcaDb](https://github.com/TCA-Team/TumCampusApp/blob/master/app/src/main/java/de/tum/in/tumcampusapp/database/TcaDb.java#L57) managed? I haven't found where it's closed, is there a different workflow that does that?

Problem with tests is: Robolectric does some SQLite shadowing (don't ask me why, at this point, I just want the tests to work) which requires database closing before each test case, else it triggers [Illegal connection pointer exception](https://www.google.de/search?q=robolectric+Illegal+connection+pointer). For Room, I haven't found a legit workaround, but my raw guess is that TcaDb just checks if instance is null (initialization), but doesn't check (does it need to?) whether the database is closed.
Should TcaDb's `getInstance` have more checking than whether the instance is null, but also if it's closed?

psukys:
Also, will have to check why CI is not running the test (not even sure what it's running for KinoActivityTest, to be honest)

psukys:
- Fixed KinoActivity to show no movies layout when there are no movies
- Adjusted to DAO
- Added covering tests for the function

PR #589, opened 2017-11-23
Improve MVV detail panel

lnsp:
## This PR fixes the following
- the detail view of the departures in MVV is way too condensed
- the departure times are inaccurate since the detail view gets its data from the card view (without refresh)
- very "light" colors that are not very well distinguishable in detail view (in my opinion, if you don't think so, I can revert it)
- long destination strings are cutoff
- fade-out animation do not work

## Screenshot
screenshot_1511468101
screenshot_1511468098

- way more visually expressive
- fixed departure times
- slick animations

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/auxiliary/DepartureView.java
+++ b/app/src/main/java/de/tum/in/tumcampusapp/auxiliary/DepartureView.java
@@ -175,7 +175,6 @@ public class DepartureView extends LinearLayout {
         public void onAnimationUpdate(ValueAnimator animator) {
             float value = (Float) animator.getAnimatedValue();
             if (getLayoutParams() != null) {
-                System.out.println("Animating" + value);
                 setTranslationX(value * getWidth());
                 getLayoutParams().height = (int) ((1.0f - value) * getHeight());
                 setAlpha(1.0f - value);
--- a/app/src/main/java/de/tum/in/tumcampusapp/auxiliary/DepartureView.java
+++ b/app/src/main/java/de/tum/in/tumcampusapp/auxiliary/DepartureView.java
@@ -2,10 +2,12 @@ package de.tum.in.tumcampusapp.auxiliary;
 
 import android.animation.ValueAnimator;
 import android.content.Context;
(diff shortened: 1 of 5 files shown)


app/src/main/java/de/tum/in/tumcampusapp/auxiliary/DepartureView.java
```diff
@@ -155,7 +163,27 @@ private void animateOut() {
...
+            mValueAnimator.removeAllUpdateListeners();
+        }
+    }
+
+    private class SlideOutAnimator implements ValueAnimator.AnimatorUpdateListener {
+        @Override
+        public void onAnimationUpdate(ValueAnimator animator) {
+            float value = (Float) animator.getAnimatedValue();
+            if (getLayoutParams() != null) {
+                System.out.println("Animating" + value);
```
pfent: no `println`s please :)

app/src/main/java/de/tum/in/tumcampusapp/auxiliary/DepartureView.java
```diff
@@ -122,16 +126,17 @@ public void setLine(CharSequence line) {
     /**
      * Sets the departure time
      *
-     * @param countDown Minutes, until this line leaves
+     * @param departureTime Timestamp in milliseconds, when transport leaves
```
pfent: that timestamp instead of a countdown makes much mure sense, well done 🎉

app/src/main/java/de/tum/in/tumcampusapp/auxiliary/DepartureView.java
```diff
@@ -140,13 +145,16 @@ private void updateDepartureTime() {
         // Keep countDown approximately in sync.
         if (mHandler != null) {
             mHandler.postDelayed(() -> {
-                mCountDown--;
                 updateDepartureTime();
```
pfent: a method reference for a single function lambda would probably be better style

lnsp: good idea, fixed it

Issue #532, opened 2017-10-17
Check if eduroam was configured by another app


xsrf:
Android won't allow you to configure eduroam if it was already configured by another app (e.g. eduroamCAT). You should check for that _before_ you ask for the users credentials and instead tell him how to delete the profile manually.

Issue #531, opened 2017-10-17
Missing eduroam realm treated as insecure configuration


xsrf:
TCA treats any eduroam configuration with missing realm as insecure configuration and alarms the user. A missing realm (username "go42tum" instead of "go42tum@eduroam.mwn.de") is not a security threat and shouldn't be treated as such.

pfent:
A missing realm is still a wrongly configured connection, that doesn't work when roaming.
While this is not a security threat, we should still prompt the user to update his configuration.
Probably prompt the user with a less intimidating message in that case?

kordianbruck:
Yea, I would vouch for a different text maybe in that case, but its still not 'correct' per se.

PR #592, opened 2017-12-01
enhance Eduroam configuration (PWD, error messages)

jacqueline8711:
## Issue

This fixes the following issue(s): #530 #531 #532

## Screenshot
screenshot_20171201-155940
screenshot_20171201-160000
screenshot_20171201-160019

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/activities/SetupEduroamActivity.java
+++ b/app/src/main/java/de/tum/in/tumcampusapp/activities/SetupEduroamActivity.java
@@ -41,10 +41,8 @@ public class SetupEduroamActivity extends BaseActivity {
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         ImplicitCounter.count(this);
-        setContentView(R.layout.activity_setup_eduroam);
 
         // Enable 'More Info' links
-        ((TextView) findViewById(R.id.text_with_link_1)).setMovementMethod(LinkMovementMethod.getInstance());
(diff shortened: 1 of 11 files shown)


jacqueline8711:
Can somebody check if the PWD authentication works for SDK < 18 ? I don't have a phone with such an old operating system and the emulator doesn't seem to support Wifi

jacqueline8711:
I haven't found a way to check if the profile wasn't created by the TCA in advance.
Well there is an attribute in the WifiConfiguration class that seems to show exactly what we want to know but it seems like it isn't accessible to a non-system application

For example here it shows that eduroam was configured by the eduroamCAT app:
screen shot 2017-12-01 at 13 12 54

app/src/main/java/de/tum/in/tumcampusapp/activities/SetupEduroamActivity.java
```diff
@@ -33,6 +34,8 @@
     private EditText lrz;
     private EditText password;
 
+    public static final String EXTRA_FOREIGN_CONFIGURATION_EXISTS = "CONFIGURED_BY_OTHER_APP";
```
kordianbruck: This should be in `de.tum.in.tumcampusapp.auxiliary.Const`

kordianbruck:
@jacqueline8711 I don't think its really possible. You can only try to set the config and then realise that it won't let you. 

As for your branch: in the future please directly create the feature branch in this repo, not in your fork, so its easier to check it out. Thanks! (Naming: jacqueline/<issueid>-<fancyname>)

PR #593, opened 2017-12-04
Disable/Enable background updates based on location

### LINKED ISSUE #460: Stop Background Services if user is more than 50KM outside of Munich
So Students love to travel. 

Lets disable any background services once they are no longer in Munich so that we don't update data they potentially don't need.

anselm92:
Use Geofencing to get location updates when the device enters or leaves
munich. Based on that disables or enables the background services.

Fixes #460

## Issue

This fixes the following issue(s): #460 

It reduces background data usage and of course battery usage when users aren't in munich and thus don't need regular updates


kordianbruck:
Hey,

so this code looks solid, but I'm not sure we need all that geofencing. Especially if you are doing a circle around a point you could as well do the following:
```java
public static float distFrom(float lat1, float lng1, float lat2, float lng2) {
    double earthRadius = 6371000; //meters
    double dLat = Math.toRadians(lat2-lat1);
    double dLng = Math.toRadians(lng2-lng1);
    double a = Math.sin(dLat/2) * Math.sin(dLat/2) +
               Math.cos(Math.toRadians(lat1)) * Math.cos(Math.toRadians(lat2)) *
               Math.sin(dLng/2) * Math.sin(dLng/2);
    double c = 2 * Math.atan2(Math.sqrt(a), Math.sqrt(1-a));
    float dist = (float) (earthRadius * c);

    return dist;
}
LocationManager lm = (LocationManager) getSystemService(Context.LOCATION_SERVICE); 
Location location = lm.getLastKnownLocation(LocationManager.GPS_PROVIDER);
float meters = distFrom(location.getLongitude(), location.getLatitude(), marienPlatzLon, marienPlatzLat);
```
Copied from:
* https://stackoverflow.com/questions/837872/calculate-distance-in-meters-when-you-know-longitude-and-latitude-in-java
* https://stackoverflow.com/questions/2227292/how-to-get-latitude-and-longitude-of-the-mobile-device-in-android

In addition I'm not sure you are actually stopping the service by setting the shared preference. In addition you would need to do something like:
```java
Intent service = new Intent(mContext, BackgroundService.class);
            if (sharedPreferences.getBoolean(key, false)) {
                mContext.startService(service);
            } else {
                mContext.stopService(service);
            }
```
(Copied from SettingsFragment)

anselm92:
The reason why I chose this solution over two others ( first one having a location listener, second service asking periodically for location) is that no polling is involved. We only receive an intent when an user leaves/enters Munich and on GPS on/off

app/src/main/java/de/tum/in/tumcampusapp/services/GeofencingRegistrationService.kt
```diff
@@ -0,0 +1,88 @@
...
+    override fun onCreate() {
+        super.onCreate()
+        locationClient = LocationServices.getGeofencingClient(baseContext)
+        val intent = Intent(this, GeofencingUpdateReceiver::class.java)
+        mGeofencePendingIntent = PendingIntent.getBroadcast(this, 0,
+                intent, PendingIntent.FLAG_UPDATE_CURRENT);
+        Utils.logwithTag(TAG, "Service started")
+    }
+
+    @SuppressLint("MissingPermission")
```
pfent: could you please add a comment, why it's ok to suppress this lint?

app/src/main/java/de/tum/in/tumcampusapp/services/GeofencingRegistrationService.kt
```diff
@@ -0,0 +1,88 @@
...
+        super.onStartCommand(intent, flags, startId)
+        return Service.START_STICKY
+    }
+
+    override fun onCreate() {
+        super.onCreate()
+        locationClient = LocationServices.getGeofencingClient(baseContext)
+        val intent = Intent(this, GeofencingUpdateReceiver::class.java)
+        mGeofencePendingIntent = PendingIntent.getBroadcast(this, 0,
+                intent, PendingIntent.FLAG_UPDATE_CURRENT);
```
kordianbruck: Really no need for this to be on a new line.

app/src/main/java/de/tum/in/tumcampusapp/services/GeofencingRegistrationService.kt
kordianbruck: @anselm92 any feedback on this?

app/src/main/java/de/tum/in/tumcampusapp/services/GeofencingRegistrationService.kt
```diff
@@ -0,0 +1,88 @@
...
+        Utils.logwithTag(TAG, "Service started")
```
kordianbruck: Can you just use `Utils.log()` without the tag? The class name is derived from the stacktrace with that function. Or was there a special reason for using tags?

app/src/main/java/de/tum/in/tumcampusapp/services/GeofencingRegistrationService.kt
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @SuppressLint("MissingPermission")
+    override fun onHandleWork(intent: Intent) {
+        if (!isLocationPermissionGranted())
```
kordianbruck: Please always use `{ }` brackets around all statements. It makes it more readable and less error prone for future developers.

app/src/main/java/de/tum/in/tumcampusapp/services/GeofencingRegistrationService.kt
```diff
@@ -0,0 +1,88 @@
...
+import de.tum.`in`.tumcampusapp.auxiliary.Const.GEOFENCING_SERVICE_JOB_ID
+import de.tum.`in`.tumcampusapp.auxiliary.Utils
+
+
+/**
+ * Service that receives Geofencing requests and registers them.
+ */
+class GeofencingRegistrationService : JobIntentService() {
+
+    private lateinit var mGeofencePendingIntent: PendingIntent
```
kordianbruck: why the `m` prefix? You are mixing it here. One member has it the other one does not. Please be consistent and remove it everywhere. The IDE colors members in a different way in any case - no added benefit of using it.

app/src/main/java/de/tum/in/tumcampusapp/services/GeofencingRegistrationService.kt
```diff
@@ -0,0 +1,88 @@
...
+    @SuppressLint("MissingPermission")
+    override fun onHandleWork(intent: Intent) {
+        if (!isLocationPermissionGranted())
+            return
+        val mRequest = intent.getParcelableExtra<GeofencingRequest>(ADD_GEOFENCE)
+        if (mRequest != null) {
+            locationClient.addGeofences(
+                    mRequest,
+                    mGeofencePendingIntent
+            )
```
kordianbruck: This can be a single line.

app/src/main/java/de/tum/in/tumcampusapp/services/GeofencingStartupReceiver.kt
```diff
@@ -0,0 +1,35 @@
...
+
+import android.content.BroadcastReceiver
+import android.content.Context
+import android.content.Intent
+import de.tum.`in`.tumcampusapp.auxiliary.Utils
+
+
+class GeofencingStartupReceiver : BroadcastReceiver() {
+
+    override fun onReceive(p0: Context?, p1: Intent?) {
```
kordianbruck: It wouldn't hurt to properly name these variables.

app/src/main/java/de/tum/in/tumcampusapp/services/GeofencingStartupReceiver.kt
```diff
@@ -0,0 +1,35 @@
...
+        if (!isValidIntent(p1))
```
kordianbruck: Brackets

app/src/main/java/de/tum/in/tumcampusapp/services/GeofencingStartupReceiver.kt
```diff
@@ -0,0 +1,35 @@
...
+        return intent != null && (intent.action == "android.intent.action.BOOT_COMPLETED" ||
+                intent.action == "android.location.MODE_CHANGED" ||
+                intent.action == "android.intent.action.QUICKBOOT_POWERON" ||
+                intent.action == "android.location.PROVIDERS_CHANGED")
+    }
+
+    companion object {
+        const val TAG = "GeofencingStartupReceiver"
+        const val DISTANCE_IN_METER = 50 * 1000f
+        const val MUNICH_GEOFENCE = "geofence_munich_id"
```
kordianbruck: Please move these constants to the `Const.kt` file

app/src/main/java/de/tum/in/tumcampusapp/services/GeofencingRegistrationService.kt
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+    }
+
+    private fun isLocationPermissionGranted(): Boolean {
+        return ContextCompat.checkSelfPermission(applicationContext, android.Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED;
+    }
+
+    companion object {
+        const val TAG = "GeofencingRegistrationService"
```
kordianbruck: Please move these constants to the `Const.kt` file

PR #601, opened 2017-12-12
Feedback

jacqueline8711:
… and email

## Issue

This fixes the following issue(s): #591 

## Screenshot
screenshot_20171212-095143

Students can directly create feedback directly in the app and can send feedback about more than one topic (different recipients)


app/src/main/java/de/tum/in/tumcampusapp/activities/FeedbackActivity.java
```diff
@@ -0,0 +1,363 @@
...
+
+    private void saveLocation(){
+        location = new LocationManager(this).getLastLocation();
+        if(location == null){
+            includeLocation.setChecked(false);
+        }
+    }
+
+    private void initIncludeEmail(Bundle savedInstanceState){
+        if(!"".equals(lrzId)){
```
pfent: Guava has a Strings.isNullOrEmpty()

app/src/main/java/de/tum/in/tumcampusapp/activities/FeedbackActivity.java
```diff
@@ -0,0 +1,363 @@
...
+                email = savedInstanceState.getString(Const.FEEDBACK_EMAIL, null);
+            }
+        }
+
+        if(email != null){
+            includeEmail.setText(getResources().getString(R.string.feedback_include_email_tum_id, email));
+            includeEmail.setChecked(true);
+        } else {
+            includeEmail.setChecked(false);
+        }
```
pfent: I think some of that `email != null` logic can be collapsed with the previous else branch...

app/src/main/java/de/tum/in/tumcampusapp/activities/FeedbackActivity.java
```diff
@@ -0,0 +1,363 @@
...
+        thumbAdapter = new FeedbackThumbAdapter(picPaths);
+        pictureList.setAdapter(thumbAdapter);
+
+        pattern = Pattern.compile(EMAIL_PATTERN);
+    }
+
+    private void initIncludeLocation(Bundle savedInstanceState){
+        includeLocation.setOnClickListener(view -> {
+            if(includeLocation.isChecked() && location == null){
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
```
pfent: I'd suggest putting code that only works on specific versions in an extra method and annotate it with [`@RequiresApi`](https://developer.android.com/reference/android/support/annotation/RequiresApi.html)

app/src/main/java/de/tum/in/tumcampusapp/activities/FeedbackActivity.java
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    public void addPicture(View view){
+        AlertDialog.Builder dialog = new AlertDialog.Builder(this);
+        dialog.setTitle(R.string.feedback_add_picture);
+        dialog.setIcon(R.drawable.ic_add_photo_colored);
+        String[] options = {getString(R.string.feedback_take_picture), getString(R.string.gallery)};
+        dialog.setItems(options, (dialogInterface, i) -> {
+            if(i == 0){
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
```
pfent: [`@RequiresApi`](https://developer.android.com/reference/android/support/annotation/RequiresApi.html)

app/src/main/java/de/tum/in/tumcampusapp/activities/FeedbackActivity.java
```diff
@@ -0,0 +1,363 @@
...
+                    int permissionCheck = ContextCompat.checkSelfPermission(this, Manifest.permission.CAMERA);
+                    if( permissionCheck == PackageManager.PERMISSION_DENIED){
+                        requestPermissions(new String[]{Manifest.permission.CAMERA}, PERMISSION_CAMERA);
+                        return;
+                    }
+                }
+                startTakingPicture();
+            } else {
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
```
pfent: [`@RequiresApi`](https://developer.android.com/reference/android/support/annotation/RequiresApi.html)

app/src/main/java/de/tum/in/tumcampusapp/adapters/FeedbackThumbAdapter.java
```diff
@@ -0,0 +1,122 @@
...
+    public ViewHolder onCreateViewHolder(ViewGroup parent, int viewType) {
+        ImageView imageView = (ImageView) LayoutInflater.from(parent.getContext())
+                                                        .inflate(R.layout.feedback_thumb, parent, false);
+        return new ViewHolder(imageView);
+    }
+
+    @Override
+    public void onBindViewHolder(ViewHolder holder, int position) {
+        if(thumbs.size() < position+1 || thumbs.get(position) == null){
+            Bitmap bitmap = createThumb(holder.imageView, position);
```
pfent: `createThumb()` can return `null`. Shouldn't this be handled?

app/src/main/java/de/tum/in/tumcampusapp/activities/FeedbackActivity.java
jacqueline8711: I feel like it's more more readable the way it is (or what reason is there to do this?). Because if I put it into an extra method I'd have to surround the method call with if(Build.VERSION.SDK_INT ...) again

app/src/main/java/de/tum/in/tumcampusapp/activities/FeedbackActivity.java
```diff
@@ -0,0 +1,363 @@
...
+public class FeedbackActivity extends BaseActivity {
+
+    private static final int REQUEST_TAKE_PHOTO = 11;
+    private static final int REQUEST_GALLERY = 12;
+    private static final int PERMISSION_LOCATION = 13;
+    private static final int PERMISSION_CAMERA = 14;
+    private static final int PERMISSION_FILES = 15;
+
+    private static final String EMAIL_PATTERN =
+            "^[_A-Za-z0-9-\\+]+(\\.[_A-Za-z0-9-]+)*@[A-Za-z0-9-]+(\\.[A-Za-z0-9]+)*(\\.[A-Za-z]{2,})$";
```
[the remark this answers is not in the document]
jacqueline8711: This looks like a great solution, but somehow it doesn't compile... Not event when I change 1.4.1 to 1.5.1 or 1.6 like it was suggested on stackoverflow. I don't know how to fix that yet...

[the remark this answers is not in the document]
jacqueline8711: It only seems to work if I don't use minifyEnabled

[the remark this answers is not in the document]
pfent: Please don't disable proguard completely, it really helps to reduce the APK size...

I'd add another proguard rule [like this](https://github.com/TCA-Team/TumCampusApp/blob/master/proguard/proguard-google-play-services.pro) and add something like:
```
-keep class commons.validator.** {*;}
```

app/src/main/java/de/tum/in/tumcampusapp/activities/FeedbackActivity.java
pfent: The way it works with the support library is, you call the method unconditionally. This then feels like you always program against the latest SDK version.

The method then does the appropriate things it needs to do for each of the API versions

app/src/main/java/de/tum/in/tumcampusapp/activities/FeedbackActivity.java
```diff
@@ -226,7 +225,7 @@ public void onRequestPermissionsResult(int requestCode, String[] permissions, in
 
     private boolean isValidEmail(){
         String email = customEmailView.getText().toString();
-        boolean isValid = EmailValidator.getInstance().isValid(email);
+        boolean isValid = Patterns.EMAIL_ADDRESS.matcher(email).matches();
```
pfent: O wow. Never heard of that util class, but seems perfect 🎉

app/src/main/java/de/tum/in/tumcampusapp/adapters/FeedbackThumbAdapter.java
jacqueline8711: I'm now deleting images for which I can't create a thumbnail

kordianbruck:
@jacqueline8711 this should be done right? Label can be removed?

jacqueline8711:
I changed the way I got the location because whenever I tested the feedback the location I got was hours old which doesn't help with the feedback at all (for me it usually showed me at home)
I put it into the LocationManager because all other location stuff is there but because of the listener that doesn't look too good...

PR #604, opened 2017-12-15
improved Eduroam config checks

jacqueline8711:


app/src/main/java/de/tum/in/tumcampusapp/cards/EduroamFixCard.java
```diff
@@ -131,63 +132,61 @@ private boolean isConfigValid() {
...
+    }
 
-        return errors.isEmpty();
+    private void checkDNSName(){
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M
+            && !isValidSubjectMatchAPI18(eduroam)) {
+            errors.add(mContext.getString(R.string.wifi_dns_name_not_set));
+        }
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
```
pfent: I think this should be an `else` :wink:

Issue #560, opened 2017-11-08
Use Googles Room Persistency Library


pfent:
We currently don't have an ORM, but write all SQL queries / mappings by hand.

Google recently released the [Android Architecture Components](https://developer.android.com/topic/libraries/architecture/index.html) with the [Room Persistency Library](https://developer.android.com/topic/libraries/architecture/room.html). We should probably use this, which should significantly reduce the complexity of our various "Managers"

kordianbruck:
@pfent allowMainThreadQueries() :+1:

kordianbruck:
Overview: https://speakerdeck.com/hugovisser/room-the-blessed-object-mapper-devfest-bayern

pfent:
Migrate Managers to use entities and DAOs:
- [x] CafeteriaManager
- [x] CafeteriaMenuManager
- [x] KinoManager
- [x] LocationManager
- [x] OpenHoursManager
- [x] SyncManager
- [x] TumManager
- [ ] Calendar Manager
- [ ] CardManager
- [ ] ChatMessageManager
- [ ] ChatRoomsManager
- [ ] NewsManager
- [ ] NotificationManager
- [ ] RecentsManager
- [ ] StudyRoomGroupManager
- [ ] SurveyManager
- [ ] TransportManager
- [ ] WifiMeasurementManager

Misc:
- [ ] Migrations / replacement for `AbstractManager.resetDb()`
- [ ] Use a proper DB abstraction for `EnhancedSearchRecentSuggestionsProvider`
- [ ] Strategy to replace MainThread queries with LiveData / Background Thread

OliHalu:
@anselm92 @psukys I allready work on ChatMessageManager, ChatRoomsManager and CalenderManager

anselm92:
@pfent @kordianbruck 
Regarding the strategy to replace main thread queries, i'd suggest using rxandroid (or rxjava in case of persistence api), I've used it many times before and love it :)

See: https://medium.com/google-developers/room-rxjava-acb0cd4f3757

Advantages: 
- It's included in the room persistence api
- It's quite easy to use. 

I'll implement it for one dao, to show how to use it.

kordianbruck:
@anselm92 what are the benefits over using livedata?

anselm92:
Basically:
+ Transformations are done on the background thread
+ Provides way more functional operations

Are we going to need it? Idk, i thinks it's more a matter of taste

See:
[...]

PR #613, opened 2018-01-15
Calendar Manager with Room persistence lib

psukys:
## Issue

Partly solves #560 by converting CalendarManager queries into Room persistence ORM

## Screenshot

Not applicable

Because having #560 solved (even though step by step) helps to maintain codebase that uses ORM without wild SQL queries not in DAOs


app/src/main/java/de/tum/in/tumcampusapp/auxiliary/calendar/IntegratedCalendarEvent.java
```diff
@@ -77,19 +78,19 @@ private static String getEventLocationFromCursor(Cursor cEvents) {
...
+    private static Calendar getEventEndFromCursor(CalendarItem cEvents) {
+        String eventEnd = cEvents.getDtend();
         Calendar result = Calendar.getInstance();
         result.setTime(Utils.getDateTime(eventEnd));
         return result;
     }
 
-    private static Calendar getEventStartFromCursor(Cursor cEvents) {
-        String eventStart = cEvents.getString(5);
+    private static Calendar getEventStartFromCursor(CalendarItem cEvents) {
```
pfent: The `*FromCursor()` methods don't make any sense anymore, since they don't take a Cursor...
This should be refactored to instance methods of `CalendarItem`

app/src/main/java/de/tum/in/tumcampusapp/cards/NextLectureCard.java
```diff
@@ -162,28 +163,27 @@ protected Notification fillNotification(NotificationCompat.Builder notificationB
         return notificationBuilder.build();
     }
 
-    public void setLectures(Cursor cur) {
-        do {
+    public void setLectures(List<de.tum.in.tumcampusapp.models.tumo.CalendarItem> calendarItems) {
+        for (de.tum.in.tumcampusapp.models.tumo.CalendarItem calendarItem: calendarItems) {
```
pfent: Is there a reason for this overly long type? `CalendarItem` should do?

app/src/main/java/de/tum/in/tumcampusapp/cards/NextLectureCard.java
```diff
@@ -162,28 +163,27 @@ protected Notification fillNotification(NotificationCompat.Builder notificationB
...
+            item.start = Utils.getDateTime(calendarItem.getDtstart());
+            item.end = Utils.getDateTime(calendarItem.getDtend());
 
             // Extract course title
-            item.title = cur.getString(0);
+            item.title = calendarItem.getTitle();
             item.title = item.title.replaceAll("[A-Z, 0-9(LV\\.Nr)=]+$", "");
             item.title = item.title.replaceAll("\\([A-Z]+[0-9]+\\)", "");
             item.title = item.title.replaceAll("\\[[A-Z]+[0-9]+\\]", "");
             item.title = item.title.trim();
```
pfent: Also: this CalendarItem to CalendarItem logic probably fits better into a member function of CalendarItem

app/src/main/java/de/tum/in/tumcampusapp/database/dao/CalendarDao.java
```diff
@@ -0,0 +1,62 @@
...
+           + "AND STATUS != 'CANCEL'"
+           + "AND NOT EXISTS (SELECT * FROM widgets_timetable_blacklist WHERE widget_id = :widgetId"
+           + "                AND lecture_title = c.title)"
+           + "ORDER BY dtstart ASC")
+    List<CalendarItem> getNextDays(String from, String to, String widgetId);
+
+    @Query("SELECT * FROM calendar WHERE datetime('now', 'localtime') BETWEEN dtstart AND dtend AND status != 'CANCEL'")
+    List<CalendarItem> getCurrentLectures();
+
+    @Query("SELECT COUNT(*) FROM calendar LIMIT 1")
```
pfent: I think the `LIMIT 1` is unnecessary

app/src/main/java/de/tum/in/tumcampusapp/database/dao/CalendarDao.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    void insert(CalendarItem cal);
+
+    @Query("SELECT * " +
+           "FROM calendar c LEFT JOIN room_locations r ON " +
+           "c.location=r.title " +
+           "WHERE r.latitude = '' " +
+           "GROUP BY c.location")
+    List<CalendarItem> getLecturesWithoutCoordinates();
```
pfent: since you're doing a `LEFT JOIN`, you probably need a check for `r.latitude IS NULL`.
I think a `WHERE coalesce(r.latitude, '') = ''` should do

app/src/main/java/de/tum/in/tumcampusapp/managers/CalendarManager.java
```diff
@@ -89,55 +92,31 @@ private static void addEvents(Context c, Uri uri) {
...
+
+            // Put the received values into a contentResolver to
+            // transmit the to Google Calendar
+            values.put(CalendarContract.Events.DTSTART, Utils.getDateTime(calendarItem.getDtstart()).getTime());
+            values.put(CalendarContract.Events.DTEND, Utils.getDateTime(calendarItem.getDtend()).getTime());
+            values.put(CalendarContract.Events.TITLE, calendarItem.getTitle());
+            values.put(CalendarContract.Events.DESCRIPTION, calendarItem.getDescription());
+            values.put(CalendarContract.Events.CALENDAR_ID, id);
+            values.put(CalendarContract.Events.EVENT_LOCATION, calendarItem.getLocation());
+            values.put(CalendarContract.Events.EVENT_TIMEZONE, R.string.calendarTimeZone);
```
pfent: This should probably be a method: `CalendarItem.toContentValues()`

app/src/main/java/de/tum/in/tumcampusapp/managers/CalendarManager.java
```diff
@@ -89,55 +92,31 @@ private static void addEvents(Context c, Uri uri) {
...
-    }
-
-    public Cursor getFromDbForDate(Date date) {
+    public List<CalendarItem> getFromDbForDate(Date date) {
         // Format the requested date
         String requestedDateString = Utils.getDateString(date);
 
         // Fetch the data
-        return db.rawQuery("SELECT * FROM calendar WHERE dtstart LIKE ? AND status!='CANCEL' ORDER BY dtstart ASC", new String[]{"%" + requestedDateString + "%"});
+        return calendarDao.getAllByDateNotCancelled("%" + requestedDateString + "%");
```
pfent: The SQL `%` wildcards should move to the DAO.

Also: I think we have a Date to String mapper for the database, so it should be possible to call `getAllByDateNotCancelled()` directly with a Date?

app/src/main/java/de/tum/in/tumcampusapp/managers/CalendarManager.java
```diff
@@ -180,13 +157,9 @@ public Cursor getCurrentFromDb() {
...
-        boolean result = false;
-        try (Cursor c = db.rawQuery("SELECT nr FROM calendar", null)) {
-            if (c.moveToNext()) {
-                result = true;
-            }
-        }
-        return result;
+        int result = calendarDao.lectureCount();
+        Utils.logv("Has " + result + " lectures");
+        return result > 0;
```
pfent: that logic can be moved to the dao

app/src/main/java/de/tum/in/tumcampusapp/managers/CalendarManager.java
```diff
@@ -219,10 +188,20 @@ public void deleteLectureFromBlacklist(int widgetId, String lecture) {
...
-        return db.rawQuery("SELECT DISTINCT c.ROWID as _id, c.title, EXISTS (" +
-                           "SELECT * FROM widgets_timetable_blacklist WHERE widget_id=? AND lecture_title=c.title" +
-                           ") as is_on_blacklist from calendar c GROUP BY c.title", new String[]{String.valueOf(widgetId)});
+    public List<CalendarItem> getLecturesForWidget(int widgetId) {
+        List<CalendarItem> lectures = new ArrayList<>();
+        List<CalendarItem> distinctLectures = calendarDao.getDistinctLectures();
+        List<CalendarItem> blacklistedLectures = calendarDao.getLecturesWithBlacklist(Integer.toString(widgetId));
+        for (CalendarItem distinctLecture: distinctLectures) {
+            for (CalendarItem blacklistedLecture: blacklistedLectures) {
+                if (distinctLecture.getTitle().equals(blacklistedLecture.getTitle())) {
```
pfent: This logic looks like a manual `JOIN` to me... Is there a good reason doing this in Java vs. just SQL?

app/src/main/java/de/tum/in/tumcampusapp/managers/CalendarManager.java
```diff
@@ -263,38 +242,27 @@ void replaceIntoDb(CalendarRow row) {
...
         }
 
-        db.execSQL("REPLACE INTO calendar (nr, status, url, title, "
-                   + "description, dtstart, dtend, location) VALUES (?, ?, ?, ?, ?, ?, ?, ?)",
-                   new String[]{row.getNr(), row.getStatus(), row.getUrl(),
-                                row.getTitle(), row.getDescription(),
-                                row.getDtstart(), row.getDtend(), row.getLocation()});
+        calendarDao.insert(new CalendarItem(row.getNr(), row.getStatus(), row.getUrl(),
+                                            row.getTitle(), row.getDescription(),
+                                            row.getDtstart(), row.getDtend(), row.getLocation(), false));
```
pfent: This logic should move either to `CalendarItem` or `CalendarRow`

app/src/main/java/de/tum/in/tumcampusapp/managers/CalendarManager.java
```diff
@@ -263,38 +242,27 @@ void replaceIntoDb(CalendarRow row) {
...
-                                      "datetime('now','localtime') < c.dtend AND r.title == c.location AND c.status!='CANCEL'" +
-                                      "ORDER BY dtstart LIMIT 1", null)) {
-            geo = null;
-            if (cur.moveToFirst()) {
-                geo = new Geo(cur.getDouble(0), cur.getDouble(1));
-            }
+        Geo geo = null;
+        RoomLocations roomLocation = roomLocationsDao.getNextLectureCoordinates();
+        if (roomLocation != null) {
+            geo = new Geo(roomLocation.getLatitude(), roomLocation.getLongtitude());
```
pfent: This logic should move to `RoomLocations.toGeo()`

app/src/main/java/de/tum/in/tumcampusapp/managers/CalendarManager.java
```diff
@@ -324,29 +292,19 @@ public QueryLocationsService() {
...
+            List<CalendarItem> calendarItems = calendarDao.getLecturesWithoutCoordinates();
+            for (CalendarItem calendarItem: calendarItems) {
+                String location = calendarItem.getLocation();
+                if (location == null || location.isEmpty()) {
+                    continue;
+                }
+                Optional<Geo> geo = locationManager.roomLocationStringToGeo(location);
+                if (geo.isPresent()) {
+                    Utils.logv("inserted " + location + ' ' + geo);
+                    roomLocationsDao.insert(new RoomLocations(location, geo.get().getLatitude(), geo.get().getLongitude()));
```
pfent: RoomLocations probably should have a constructor `RoomLocations(String, Geo)`

app/src/main/java/de/tum/in/tumcampusapp/services/SilenceService.java
```diff
@@ -96,47 +98,44 @@ protected void onHandleWork(@NonNull Intent intent) {
...
+                Utils.log("set ringer mode to old state");
+                am.setRingerMode(Integer.parseInt(
+                        Utils.getSetting(this, Const.SILENCE_OLD_STATE,
+                                         Integer.toString(AudioManager.RINGER_MODE_NORMAL))));
+                Utils.setInternalSetting(this, Const.SILENCE_ON, false);
+
+                List<CalendarItem> nextCalendarItem = calendarManager.getNextCalendarItem();
+                if (nextCalendarItem.size() != 0) { //Check if we have a "next" item in the database and update the refresh interval until then. Otherwise use default interval.
+                    // refresh when next event has started
+                    waitDuration = getWaitDuration(nextCalendarItem.get(0).getDtstart());
```
pfent: for this logic we could probably use a `getNextCalendarItem()` method that actually only returns an Optional of the very next item, instead of a List...

app/src/main/java/de/tum/in/tumcampusapp/services/SilenceService.java
```diff
@@ -96,47 +98,44 @@ protected void onHandleWork(@NonNull Intent intent) {
...
+            String mode = Utils.getSetting(this, "silent_mode_set_to", "0");
+            if ("0".equals(mode)) {
+                Utils.log("set ringer mode: vibration");
+                am.setRingerMode(AudioManager.RINGER_MODE_VIBRATE);
+            } else {
+                Utils.log("set ringer mode: silent");
+                am.setRingerMode(AudioManager.RINGER_MODE_SILENT);
             }
+            // refresh when event has ended
+            waitDuration = getWaitDuration(currentLectures.get(0).getDtstart());
```
pfent: And this probably is handled better with an Optional of the current lecture instead of a List

app/src/main/java/de/tum/in/tumcampusapp/managers/CalendarManager.java
psukys: This would require an additional entity just for that specific use case, I decided to re-use calendar-item with non-persisted field, since the blacklist contains only blacklisted items.

pfent: With a second thought:
Can't you just query all non-blacklisted lectures (inverse of `getLecturesWithBlacklist()`) and merge those two lists?
I.e.
```java
List<CalendarItem> lectures = getBlacklistedLectures(widgetId);
for (CalendarItem i : lectures) {
    i.setBlacklisted(true);
}
lectures.addAll(getNonBlacklistedLectures(widgetId))
```

I really dislike the idea performing an O(n²) operation for a potentially large amount of lectures...

psukys: That should work!

app/src/main/java/de/tum/in/tumcampusapp/cards/NextLectureCard.java
psukys: NextLectureCard has a static class with public fields to *conveniently* store some values that are used in the card. I would use CalendarItem entity, but some fields types are different (dates), but I will try to change them and see for back-compatibility

app/src/main/java/de/tum/in/tumcampusapp/managers/CalendarManager.java
psukys: Will leave out CALENDAR_ID and EVENT_TIMEZONE as they seem to be out of CalendarItem's scope

app/src/main/java/de/tum/in/tumcampusapp/services/SilenceService.java
psukys: Yeah, the naming is misleading, the query originally returned 4 items and the same query is used at `onRequestCard` function which uses the list.

psukys:
@pfent, updated and adjusted where I could

app/src/main/java/de/tum/in/tumcampusapp/managers/CalendarManager.java
psukys: Actuallly, I remembered that converter converts to minute accuracy, whereas it is not the case for this query, which typically is used to retrieve day's events. That is: converter would make a string "2018-01-18 10:11", whereas previously it was expected that the string would be "2018-01-18" which would return all the calendar items for the day.

I've restored the version where a string is passed, but moved the `%` into the query

pfent: hmm, I think we'd should fix the query then to work with full-accuracy datetimes, e.g. by surrounding the parameter with [`date()`](https://www.sqlite.org/lang_datefunc.html), so we don't need to do the string conversion

pfent:
The `kinoAdapterUsedTest` shouldn't be related to this PR...

Probably open another PR to fix the test...

psukys:
@pfent yeah likely. Recently, it was converted to rxjava solution and it's questionable on how to test it with robolectric, but I'll open an issue on this

PR #614, opened 2018-01-16
#560 Implement Wifimeasurements

kordianbruck:
Please review <3

#560

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/managers/WifiMeasurementManager.java
+++ b/app/src/main/java/de/tum/in/tumcampusapp/managers/WifiMeasurementManager.java
@@ -1,14 +1,15 @@
 package de.tum.in.tumcampusapp.managers;
 
 import android.content.Context;
-import android.database.Cursor;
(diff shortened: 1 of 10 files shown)


app/src/main/java/de/tum/in/tumcampusapp/managers/WifiMeasurementManager.java
```diff
@@ -125,6 +85,10 @@ public void uploadMeasurementsToRemote(int maxRetries, int waitTimeInMillis) {
             }
             currentAttempts++;
         }
-        db.execSQL("DELETE FROM wifi_measurement;");
+
+        //Only remove them if we were successful
+        if (successful) {
```
psukys: Good update

detekt-config.yml
```diff
@@ -0,0 +1,366 @@
+autoCorrect: true
```
psukys: What is this file for?

psukys: > codacity configuration

PR #616, opened 2018-01-16
Add rxjava for kino

anselm92:


app/src/main/java/de/tum/in/tumcampusapp/models/tumcabe/Kino.kt
```diff
@@ -34,7 +36,7 @@ data class Kino(@PrimaryKey
                 var rating: String = "",
                 var description: String = "",
                 var cover: String = "",
-                var trailer: String = "",
+                var trailer: String? = "",
```
pfent: why does this field need to be nullable?

PR #618, opened 2018-01-16
Room - Refactor notification manager

kordianbruck:
#560 - please review


app/src/main/java/de/tum/in/tumcampusapp/database/Converters.kt
```diff
@@ -0,0 +1,29 @@
...
+import android.arch.persistence.room.TypeConverter
+import com.google.gson.Gson
+import de.tum.`in`.tumcampusapp.auxiliary.Utils
+import de.tum.`in`.tumcampusapp.models.gcm.GCMNotificationLocation
+import java.util.*
+
+class Converters {
+    @TypeConverter
+    fun fromIso(str: String): Date {
+        return Utils.getDateTime(str)
```
pfent: please use `fun fromIso(str) = Utils.getDateTime(str)` for [single expression functions](https://www.packtpub.com/mapt/book/application_development/9781787126367/4/ch04lvl1sec49/single-expression-functions)

app/src/main/java/de/tum/in/tumcampusapp/services/GcmReceiverService.java
```diff
@@ -97,8 +97,10 @@ public void onMessageReceived(RemoteMessage message) {
                     Utils.log(e);
                 }
 
-                //de.tum.in.tumcampusapp.managers.NotificationManager man = new de.tum.in.tumcampusapp.managers.NotificationManager(this);
-                //@todo save to our notificationmanager
+                //TODO
+                //de.tum.in.tumcampusapp.managers.NotificationManager notificationManager
+                //        = new de.tum.in.tumcampusapp.managers.NotificationManager(this);
+                //notificationManager.replaceInto(n);
```
pfent: Is this still relevant?

kordianbruck: yes, but as I don't have the full GCMNotification but only Key/Value List, its kinda tricky to insert it. Will handle it in a separate PR in the next weeks.

PR #623, opened 2018-01-17
Allow sorting of cards

anselm92:
Cards that are of a different type can be drag and dropped
up and down the list to change their location (persistent).

## Issue

This fixes the following issue(s): #228

## Screenshot

screenshot_20180118-004710

Users can change positions of cards according to their needs

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/cards/generic/Card.java
+++ b/app/src/main/java/de/tum/in/tumcampusapp/cards/generic/Card.java
@@ -6,6 +6,7 @@ import android.content.Intent;
 import android.content.SharedPreferences;
 import android.content.SharedPreferences.Editor;
 import android.preference.PreferenceManager;
+import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.support.v4.app.ActivityOptionsCompat;
 import android.support.v4.content.ContextCompat;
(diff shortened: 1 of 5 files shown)


app/src/main/java/de/tum/in/tumcampusapp/cards/generic/Card.java
```diff
@@ -209,6 +210,28 @@ public void hideAlways() {
...
+        e.apply();
+    }
+
+    @Override
+    public int compareTo(@NonNull Card card) {
+        if(card.getPosition()>getPosition()) {
+            return -1;
+        } else {
+            return 1;
+        }
```
pfent: I'd do an `Integer.compare(card.getPosition(), getPosition())` or vice versa

anselm92:
Ok (y) 
I've two questions though:
1) On restore cards, should the position be reset too?
2) Should sticky cards still be on top or moveable too?

psukys:
@anselm92 if restore does some kind of resetting, then I think the position should change as well.

kordianbruck:
@anselm92 we only have one sticky card, so make these persistent at the top. Once you dismiss them, they usually never reappear.

How does this interact with swiping the cards right/left? Is that still possible?

anselm92:
@kordianbruck 
alright.
yes you can still dismiss cards.

pfent:
@kordianbruck This one can also be merged

PR #626, opened 2018-01-19
room persistency transport manager

OliHalu:
## Issue

 Partly solves #560 by converting TransportManager queries into Room persistence ORM

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/managers/TransportManager.java
+++ b/app/src/main/java/de/tum/in/tumcampusapp/managers/TransportManager.java
@@ -2,8 +2,6 @@ package de.tum.in.tumcampusapp.managers;
 
 import android.content.ContentValues;
 import android.content.Context;
-import android.database.Cursor;
-import android.database.MatrixCursor;
 import android.util.Pair;
 import android.util.SparseArray;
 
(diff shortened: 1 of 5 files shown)


app/src/main/java/de/tum/in/tumcampusapp/managers/TransportManager.java
```diff
@@ -148,9 +148,8 @@ public TransportManager(Context context) {
      * @return True, if favorite
      */
     public boolean isFavorite(String symbol) {
-        try (Cursor c = db.rawQuery("SELECT * FROM transport_favorites WHERE symbol = ?", new String[]{symbol})) {
-            return c.getCount() > 0;
-        }
+        List<TransportFavorites> favorites = transportDao.getFavorites(symbol);
+        return favorites.size() > 0;
```
pfent: This logic should be handled completely in the DAO.
I think you could write a query like: 
```SQL
SELECT EXISTS(SELECT * FROM transport_favorites WHERE symbol = :symbol);
```
directly returning a boolean

app/src/main/java/de/tum/in/tumcampusapp/models/transport/TransportFavorites.kt
```diff
@@ -0,0 +1,9 @@
+package de.tum.`in`.tumcampusapp.models.transport
+
+import android.arch.persistence.room.Entity
+import android.arch.persistence.room.PrimaryKey
+
+@Entity(tableName = "transport_favorites")
+data class TransportFavorites(@PrimaryKey(autoGenerate = true)
+                              var id: Int = 0,
+                              var symbol: String = "")
```
pfent: shouldn't the symbol be unique? can't we even set it as primarykey?

OliHalu: Autogenerate was set to true for the PrimaryKey "id". If we want to set "symbol" as PrimaryKey as well, we should use @Entity(primaryKeys = ["id","symbol"] (proper way to declare composite PrimaryKeys in Room otherwise there will be an error: You cannot have multiple primary keys defined in an Entity...). But then the unique ids for the column "id" are not generated automatically anymore.

pfent: I thought more of a UNIQUE column...
https://stackoverflow.com/a/46035385

Composite primary keys wouldn't help much, you're right...

Issue #624, opened 2018-01-18
Failing test: kinoAdapterUsedTest


psukys:
## Steps to Reproduce

1. Run test suite `./gradlew test`
2. Verify unittest `kinoAdapterUsedTest` fails **after several runs**

Sample problem, as seen in the builds, [#223 failed build](https://server4.wdkk.de/job/TCA-Android/223/testReport/junit/de.tum.in.tumcampusapp.activities/KinoActivityTest/kinoAdapterUsedTest/) from #613  PR

This is likely only problematic within the test suite.

## Expected Result
`de.tum.in.tumcampusapp.activities.KinoActivityTest > kinoAdapterUsedTest PASSED` :100:

## Actual Result
`de.tum.in.tumcampusapp.activities.KinoActivityTest > kinoAdapterUsedTest FAILED`

## Problem
Usage of rxjava and robolectric. There is likely a race condition, as the test *sometimes* fails.

## Suggested Enhancement :sparkles:
Finding/creating a guide if we intend to convert and add test functionality in the future.

### Environment
* Phone: not applicable
* OS version: not applicable
* TUM Campus App version: 156b006 
* Language: not applicable

psukys:
Resolved by #627

PR #627, opened 2018-01-19
Fix failing kino test

anselm92:
- Add delay until ui's been updated
- Remove obsolete code

## Issue

This fixes the following issue(s): #624

However this does not really fix the issue, it's just a workaround for now. 
We should discuss and find a proper way how to test activities that are updated asynchronously.


psukys:
@anselm92 I think the other test I've written in a broken fashion, else should be good, I think

pfent:
@kordianbruck plz merge dis

PR #629, opened 2018-01-21
Fix tests that manipulate DateTime

psukys:
## Issue

Sometimes time based tests fail, it would be good to have them either always fail or always pass.

### DIFF
--- a/app/src/test/java/de/tum/in/tumcampusapp/database/dao/CalendarDaoTest.java
+++ b/app/src/test/java/de/tum/in/tumcampusapp/database/dao/CalendarDaoTest.java
@@ -68,10 +68,11 @@ public class CalendarDaoTest {
      */
     @Test
     public void getAllNotCancelledAll() {
-        dao.insert(createCalendarItem("OPEN", DateTime.now()));
-        dao.insert(createCalendarItem("OTHER", DateTime.now()));
-        dao.insert(createCalendarItem("OTHER", DateTime.now()));
-        dao.insert(createCalendarItem("OPEN", DateTime.now()));
+        DateTime now = DateTime.now();
+        dao.insert(createCalendarItem("OPEN", now));
+        dao.insert(createCalendarItem("OTHER", now));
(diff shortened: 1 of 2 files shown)


psukys:
I believe that parts of code use [different types of locales](https://github.com/TCA-Team/TumCampusApp/search?utf8=%E2%9C%93&q=Locale.). Any thoughts on how this should be approached, @pfent, @kordianbruck ?

psukys:
Not sure, but there might be timezone problems?

app/src/test/java/de/tum/in/tumcampusapp/database/dao/CalendarDaoTest.java
```diff
@@ -68,10 +68,11 @@ private CalendarItem createCalendarItem(String status, DateTime startDate, DateT
      */
     @Test
     public void getAllNotCancelledAll() {
-        dao.insert(createCalendarItem("OPEN", DateTime.now()));
-        dao.insert(createCalendarItem("OTHER", DateTime.now()));
-        dao.insert(createCalendarItem("OTHER", DateTime.now()));
-        dao.insert(createCalendarItem("OPEN", DateTime.now()));
+        DateTime now = DateTime.now();
```
pfent: I'd make this DateTime static, i.e. a hardcoded time, that is known to be ok

psukys: The issue with test failing was `cleanUp` function that used SQL's localtime time function (and subtracting 3 months from it to compare). Yes, DateTime.now() generates a new time, but with current commit I've isolated each test case to expected times, so that exact behaviour is expected. Hardcoding time would decrease test stability as they're supposed to fail if something not under behaviour happens.

I'll try to investigate what timezone SQL database uses.

pfent: Hm, seems reasonable.

pfent:
@kordianbruck plz merge

PR #639, opened 2018-01-27
Fix clearing cache makes app unusable

### LINKED ISSUE #636: Clearing the Cache through app makes it unusable
The SQL DB is not recreated by room automatically. Thats why we get multiple exceptions.

Please refine that action, to allow resetting the app without deleting the token.

anselm92:
- Do not drop database tables but instead clear them.
Room can't handle recreation of tables yet and thus trows exceptions
when tables were deleted and not recreated.

Fixes #636

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/managers/AbstractManager.java
+++ b/app/src/main/java/de/tum/in/tumcampusapp/managers/AbstractManager.java
@@ -7,6 +7,7 @@ import android.database.sqlite.SQLiteDatabase;
 import java.io.File;
 
 import de.tum.in.tumcampusapp.auxiliary.Const;
+import de.tum.in.tumcampusapp.database.TcaDb;
 import de.tum.in.tumcampusapp.services.BackgroundService;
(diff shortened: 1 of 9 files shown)


app/src/main/java/de/tum/in/tumcampusapp/managers/AbstractManager.java
```diff
@@ -64,32 +65,42 @@ public static void resetDb(Context c) {
...
+        tdb.facultyDao().flush();
+        tdb.transportDao().removeCache();
+        tdb.studyRoomDao().removeCache();
+        tdb.studyRoomGroupDao().removeCache();
+        tdb.kinoDao().flush();
+        tdb.widgetsTimetableBlacklistDao().flush();
+        tdb.notificationDao().cleanup();
+        tdb.favoriteDishDao().removeCache();
+        tdb.buildingToGpsDao().removeCache();
+        tdb.wifiMeasurementDao().cleanup();
```
pfent: We could probably use an Interface for this, but for now it's fine

Issue #573, opened 2017-11-09
Chat Room Activity does not show history


kordianbruck:
Whenever you leave the room and reenter it, it now longer displays the previous messages

OliHalu:
It looks like the Cursor we create in the `getNextHistoryFromServer(final boolean newMsg)` method is closed before the message history is retrieved. Deleting `cur.close()` will solve the problem that the message history is not shown. But I think the cursor needs to be closed somewhere to avoid memory problems right?

kordianbruck:
@OliHalu ok, if that really the problem (I couldn't verify this in a quick debug session) please start migrating the Chat to the room persistency library and merge that into philipps branch (Create a PR for this). 
You guys really need to get going! :alarm_clock:

PR #642, opened 2018-02-04
Chat bug fix

OliHalu:
## Issue
Sometimes the message wasn't send. Lettng the thread sleep for 1.5 seconds after sending the message fixes this bug.

#573 

Whenever you reenter a chat room the chat history is shown


kordianbruck:
@OliHalu is this ready for review?

Issue #502, opened 2017-07-25
Implement NotificationChannels


pfent:
Android O introduces NotificationChannels for better user-control of notifications:
https://developer.android.com/preview/features/notification-channels.html

I implemented a default / dummy version in bbf15cf already, but my implementation does not do anything useful yet.

Also, we should probably enable more of those Notifications by default. I didn't even remember, that we had so many Notifications for useful things.

PR #650, opened 2018-02-09
Notification Improvements

jacqueline8711:
## Issue

This fixes the following issue(s): #502 

## Screenshot
screenshot_20180209-170139

Students can change notification properties like vibration, lights etc. for different notifications (>= Android O)


pfent:
Awesome! Thanks for implementing this.

We can probably also change the defaults to show all those notifications...

jacqueline8711:
@pfent I'll do that. 

I'll also have to change the icons, since all I see is a white (or grey) circle instead of the TUM logo when I get a notification (at least for the cafeteria notification, idk about the others)

app/src/main/java/de/tum/in/tumcampusapp/notifications/Alarm.java
```diff
@@ -30,6 +30,9 @@
 import de.tum.in.tumcampusapp.models.gcm.GCMAlert;
 import de.tum.in.tumcampusapp.models.gcm.GCMNotification;
 
+/**
+ * TUM Alarmierung
```
pfent: German code :see_no_evil:

Issue #660, opened 2018-02-13
Easter Egg


kordianbruck:
https://github.com/TCA-Team/iOS/pull/224

jacqueline8711:
@kordianbruck in the iOS issue it says it changes when tapping on the tum logo, so I guess when tapping on the ActionBar (on the main screen)?

kordianbruck:
We don't have the TUM Logo in the Toolbar/ActionBar on the main screen yet. 

No, the Idea was to have it during the startup screen, where we show the offical TUM logo for a few seconds.

psukys:
Agree, with time it could build up even more variations, supposedly time-of-year theming as well

PR #665, opened 2018-02-14
colorful tum logo

jacqueline8711:
## Issue
Shows colorful tum logo on startup screen and drawer after tapping startup or main action bar three times (toggle for going back and forth between the logos).

This fixes the following issue(s): #660 

## Screenshot
screenshot_20180214-001127
screenshot_20180213-235201


app/src/main/java/de/tum/in/tumcampusapp/activities/StartupActivity.java
```diff
@@ -147,6 +148,34 @@ protected void onCreate(Bundle savedInstanceState) {
...
+            tumLogo.setImageResource(R.drawable.tum_logo_rainbow);
+        } else {
+            tumLogo.setImageResource(R.drawable.tum_logo);
+        }
+
+        tapCounter = 0;
+        View background = findViewById(R.id.startup_background);
+        background.setOnClickListener(view -> {
+            tapCounter++;
+            if(tapCounter == 3){
```
pfent: This OnClockListener in Startup- / MainActivity is duplicated code. Please extract the common logic

app/src/main/java/de/tum/in/tumcampusapp/activities/MainActivity.java
```diff
@@ -118,6 +125,21 @@ public void onDrawerOpened(View drawerView) {
         } else {
             initAdapter();
         }
+
+        tapCounter = 0;
+        findViewById(R.id.main_toolbar).setOnClickListener(view -> {
+            tapCounter++;
```
pfent: That `tapCounter` incremented on every tap, even after quite some time.
I feel like only tapping the logo in quick succession should toggle the logo. I'd suggest tp remember the last time of tap and reset the `tapCounter` to 1 when more than one second has passed after the last tap

kordianbruck: its a bit overkill. The start screen is only visible for a few seconds.

PR #668, opened 2018-02-15
extend thread sleep time

OliHalu:
When using the Chat with mobile data/wifi the message is not shown after sending it when the connectivity is too bad. By extending the sleep time from 1500 ms to 2000 ms it works even with a bad connection.

Because the message you send should shown immediately after sending.


kordianbruck:
@OliHalu this certainly cannot be the solution to the problem at hand. There must be some way of synchronizing/notifying other threads once the data updated. We can't simply sleep.

Issue #654, opened 2018-02-13
ToolBar too big


kordianbruck:
So testing the latest master, there are a couple bugs that where introduced.

This toolbar is double the height - should only be one line.

![screenshot_20180213-011318](https://user-images.githubusercontent.com/298860/36126994-710923b2-105b-11e8-9184-d1267e3506a5.png)

kordianbruck:
@OliHalu are you going to fix this?

kordianbruck:
@OliHalu nevermind, I've fixed it myself.

PR #673, opened 2018-02-16
Oli halu/654 tool bar too big

OliHalu:
## Issue
Chat ToolBar was to big
This fixes the following issue(s): #654

More space for chat message history


kordianbruck:
Already implemented. Thanks @OliHalu

PR #677, opened 2018-02-19
Remove EnhancedSearchRecentSuggestionsProvider

psukys:
## Issue

Use Android builtin solution instead of `EnhancedSearchRecentSuggestionsProvider` 

## Screenshot

Not applicable

Less code is probably better.


psukys:
I'll do some manual usability tests and maybe could write a test or two if applicable.

psukys:
Alright, it seems that recent search suggestions are not individual per activity, whereas previously they were. Was this the original intention with Enhanced[...], @pfent ? The documentation is pretty vague on this class.

pfent:
If that's the only difference, that's totally fine with me.

I'll take a closer look tomorrow

kordianbruck:
@psukys yes, the id parameter usually distinguished the different search values: 
```
* @param id        String identifying the SuggestionProvider. Must not contain blanks or any
-     *                  other special characters
```

kordianbruck:
I think we want to go towards a unified search activity in any case, so this is fine for any casse.

pfent:
> +16 −343

Much less complexity for basically the same functionality :tada:

Issue #659, opened 2018-02-13
Not all rooms shown till refreshed


kordianbruck:
Happens every time I open the chatrooms activity. Before the refresh only 2-3 rooms are shown. After refresh I get all the rooms (20+)

OliHalu:
@kordianbruck are 2-3 rooms of 'My Chatrooms' shown or the rooms which belong to the lecture the user attends?

kordianbruck:
of 'My Chatrooms' - all of the current semester are missing somehow.

anselm92:
I tested this on 1.5.1, happens there too. so this is an old bug

kordianbruck:
Still, this should be fixed if possible in a reasonable time

kordianbruck:
Fix in PR.

Issue #653, opened 2018-02-13
Exception when sending text in chat


kordianbruck:
So testing the latest master, there are a couple bugs that where introduced.

```
02-13 01:11:48.615 7143-7143/de.tum.in.tumcampus E/AndroidRuntime: FATAL EXCEPTION: main
                                                                   Process: de.tum.in.tumcampus, PID: 7143
                                                                   java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
                                                                       at java.util.ArrayList.get(ArrayList.java:411)
                                                                       at android.widget.HeaderViewListAdapter.isEnabled(HeaderViewListAdapter.java:164)
                                                                       at android.widget.AbsListView.setItemViewLayoutParams(AbsListView.java:2430)
                                                                       at android.widget.AbsListView.obtainView(AbsListView.java:2399)
                                                                       at android.widget.ListView.makeAndAddView(ListView.java:1970)
                                                                       at android.widget.ListView.fillUp(ListView.java:738)
                                                                       at android.widget.ListView.layoutChildren(ListView.java:1711)
                                                                       at android.widget.AbsListView.onLayout(AbsListView.java:2170)
                                                                       at android.view.View.layout(View.java:17641)
                                                                       at android.view.ViewGroup.layout(ViewGroup.java:5575)
                                                                       at android.widget.RelativeLayout.onLayout(RelativeLayout.java:1079)
                                                                       at android.view.View.layout(View.java:17641)
                                                                       at android.view.ViewGroup.layout(ViewGroup.java:5575)
                                                                       at android.widget.LinearLayout.setChildFrame(LinearLayout.java:1741)
                                                                       at android.widget.LinearLayout.layoutVertical(LinearLayout.java:1585)
                                                                       at android.widget.LinearLayout.onLayout(LinearLayout.java:1494)
                                                                       at android.view.View.layout(View.java:17641)
                                                                       at android.view.ViewGroup.layout(ViewGroup.java:5575)
                                                                       at android.support.v4.widget.DrawerLayout.onLayout(DrawerLayout.java:1172)
                                                                       at android.view.View.layout(View.java:17641)
                                                                       at android.view.ViewGroup.layout(ViewGroup.java:5575)
                                                                       at android.widget.FrameLayout.layoutChildren(FrameLayout.java:323)
                                                                       at android.widget.FrameLayout.onLayout(FrameLayout.java:261)
                                                                       at android.view.View.layout(View.java:17641)
[...]

PR #686, opened 2018-02-20
More Chat Fixes

kordianbruck:
As this is completely broken and unusable right now, here are more fixes

ToDo:
- [x] #653 
- [x] #659


app/src/main/java/de/tum/in/tumcampusapp/cards/ChatMessagesCard.java
```diff
@@ -92,7 +92,7 @@ public void setChatRoom(String roomName, int roomId, String roomIdString) {
         mRoomName = mRoomName.replaceAll("\\[[A-Z]+[0-9]+\\]", "");
         mRoomName = mRoomName.trim();
         chatMessageDao.deleteOldEntries();
-        mUnread = chatMessageDao.getLastUnread(roomId);
+        //mUnread = chatMessageDao.getLastUnread(roomId); TODO
```
pfent: `mUnread` remains uninitialized :boom:

kordianbruck: Well, there is a todo marker there :tada:

PR #693, opened 2018-02-21
#672 restructure project

### LINKED ISSUE #672: New Folder Structure
The App as it is, is hard to understand, as everything is really far apart and we have a lot of features packed into it. So I would recommend the following improved app structure:

* component
   * abstract
   * cafeteria
      * activity
      * controller
      * adapter
      * fragment
      * model
         * api
         * db
         * view
      * cards
      * widget
      * dao
      * repository
* api
* database
* notification
   * model
* service
* exception

@pfent @anselm92 ideas? thoughts?

I would like to get more to a MVVM approach of doing things. Remove any business logic. Get rid of all the managers we have and move that to repos / controllers.

kordianbruck:
@TCA-Team/android please wait till this is done - otherwise you will have a lot of conflicts.

fixes #672


pfent:
@kordianbruck I hope you don't mind me getting this to actually compile...

kordianbruck:
@pfent  feel free to continue. Im gone for couple of hours

pfent:
I also feel like the components are a little overstructured ATM:
```
component$ find . -type d | wc -l
70
component$ find . -name '*.java' | wc -l
117
component$ find . -name '*.kt' | wc -l
11
```

70 folders for 130 classes, I wouldn't call that reducing complexity :see_no_evil:

kordianbruck:
The structure within components surely can be optimized. But the main objective is to have less dependencies between our activities and introduce hard boundaries. 

Previously we had very long lists of classes and no clear boundaries. 

Surely smaller components don't need the full structure in them but can be merged to a single package.

pfent:
Sure, I see.
Now AndroidStudio complains in every other class "Access can be package-private", which is a good thing I guess :smile: 

@kordianbruck You might want to have a look at the final structure I came up with, feel free to move it around as you wish.
I'd keep the size of individual packages in range: 3 ≤ size ≤ 10, just to keep the overhead small

Issue #702, opened 2018-02-22
EduroamFixCard - getAltSubjectMatch


kordianbruck:
Probably related to: #698 

```
java.lang.NoSuchMethodError: No virtual method getAltSubjectMatch()Ljava/lang/String; in class Landroid/net/wifi/WifiEnterpriseConfig; or its super classes (declaration of 'android.net.wifi.WifiEnterpriseConfig' appears in /system/framework/framework.jar)
at de.tum.in.tumcampusapp.cards.EduroamFixCard.checkDNSName(EduroamFixCard.java:180)
at de.tum.in.tumcampusapp.cards.EduroamFixCard.isConfigValid(EduroamFixCard.java:152)
at de.tum.in.tumcampusapp.cards.EduroamFixCard.shouldShow(EduroamFixCard.java:80)
at de.tum.in.tumcampusapp.cards.generic.Card.apply(Card.java:157)
at de.tum.in.tumcampusapp.cards.generic.NotificationAwareCard.apply(NotificationAwareCard.java:88)
at de.tum.in.tumcampusapp.managers.CardManager.update(CardManager.java:122)
at de.tum.in.tumcampusapp.services.DownloadService.download(DownloadService.java:145)
at de.tum.in.tumcampusapp.services.DownloadService.lambda$onHandleWork$0$DownloadService(DownloadService.java:198)
at de.tum.in.tumcampusapp.services.DownloadService$$Lambda$0.run(Unknown Source)
at java.lang.Thread.run(Thread.java:818)
```

PR #708, opened 2018-02-22
correctly checking api (eduroam)

jacqueline8711:
## Issue
I think this should fix both #702 and #698 but I haven't tested it yet


kordianbruck:
Work in progress then till tested :+1:

pfent:
@jacqueline8711 Could you already test this?

jacqueline8711:
@pfent to be honest I don't have an old enough test device I think testing in an emulator is not possible because it doesn't have wifi. Ideas?

pfent:
Then let's merge it anyway, maybe @M-FF-M can test it proper

jacqueline8711:
I should work fine. I know what caused the crashes,  i.e. what I did wrong and I fixed that

Issue #682, opened 2018-02-20
WS2017 Project Completion


kordianbruck:
Here are a few todos, that are still open for the Android Praktikum Guys

- [x] #272 Creating a fancy video
- [ ] #674 Daoize Cache Manager
- [x] #659 Not all rooms shown till refreshed - kordian
- [x] #653 Exception when sending text in chat - kordian
- [ ] #587 Comment Bot - paulius
- [x] #586 SonarQube?
- [x] #574 Calender does not display past events - oli
- [x] #568 Code Coverage - paulius
- [x] #567 Add CheckStyle - paulius
- [ ] #356 Code Quality
- [x] #640 NewsDaoTest.cleanUpOldTest sometimes failing
- [x] Send me a copy of the presentation
- [x] Send me the visio files for all graphs you made
- [x] Send me all the source files for the fancy video
- [ ] Finish documentation - this needs to be complete and has a big grade impact.

I have to submit the Grading till the Wednesday, 07.03.2018 at the very latest. Please have everything done by then, otherwise this might impact the final outcome. Each of you gets a separate grade, but also team effort counts. So anyone feeling he has not yet lived up to their standards, now is your time to shine :star2:

kordianbruck:
Alrighty, we're done with the restructure. Let us know, what you guys think. :heart:

psukys:
Related documentation:

 - [Quality Assurance](https://github.com/TCA-Team/TumCampusApp/wiki/Quality-Assurance)
 - [Code style with Checkstyle](https://github.com/TCA-Team/TumCampusApp/wiki/Code-style-with-Checkstyle)
 - [Writing tests with Robolectric](https://github.com/TCA-Team/TumCampusApp/wiki/Writing-tests-with-Robolectric) (Working on it)

Issue #574, opened 2017-11-09
Calendar does not display past events


kordianbruck:
It should not delete those, so one can actually go back

kordianbruck:
@psukys is there any development on this? Done? Do we still purge the db on upgrade?

psukys:
Haven't checked this in a while, only on when making CalendarManager DAO and then I couldn't find clear indicators of this behaviour, will look into this after I've resolved the fishy test db date issue (#640)

kordianbruck:
As we now go two months backwards, this should no longer be a problem. Still a proper paging might be interesting in the future..

PR #715, opened 2018-02-23
calendar does not show past events

OliHalu:
## Issue
Past events were not shown in calendar. I set the number of past months shown in calendar to 2. Not sure if going much further into the past is necessary.

#574  - Calendar does not show past events
Fixes part of #682 

Because now user can check past events in their calendar.


kordianbruck:
@pfent actually this is troublesome. Really we should properpy do paging. So just load one month each time. Most users won't go into the past, so enabling all users to simply download two more months, puts a lot of stress on the TUMonline DB.

pfent:
@kordianbruck how would paging work with a calendar, that is synced to the device calendar?
Also, what's more stress? My calendar is currently 36kb for the last 3 months, which does not seem to be very tasking to me...

kordianbruck:
No clue how those synchronized cals work. But thats not implemented right now in any case.

Its not the size that matters here. TUMonline is doing like 10 union alls on different tables that have several million records. This times 20k devices that sync this at least every week, is quite much. Already the calendar API is the one using the most CPU time. 

@xsrf you wouldn't happen to have any statistics about this?

pfent:
Hrrngh :man_facepalming: 

We haven't released those changes, yet, So there's no rush fixing that.

I suppose we could do something more clever than a [`DELETE FROM calendar`](https://github.com/TCA-Team/TumCampusApp/blob/ad28cb21894c58d899cd8186171fe09b78c7462a/app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarController.java#L197), every time we sync it :man_shrugging: 
However, we potentially show wrong information for the past then (e.g. Sync Monday, event gets canceled on Tuesday, Sync Wednesday: we can't possibly know when only syncing future events). My reasoning was: Just trust TUMonline...

xsrf:
Access to "kalender" isn't a problem at the moment (I guess). The API gets called twice a week per token on average. Only few devices call it ~20 times per week. Requests are processed within 1s or less.

kordianbruck:
@xsrf cool beans - so this means we can move ahead with simply adding some months to the result set.

PR #722, opened 2018-02-25
Default to false for silence service

kordianbruck:
We need special permissions. Those are only checked, if one tries to activate the silence service. Only logical thing to do, is to default to false, so that a student has to manually activate it.


kordianbruck:
We do prompt them, just he needs to activate it himself. Defaulting to true does not check the permission.

PR #724, opened 2018-02-26
Add code coverage

### LINKED ISSUE #568: Code Coverage
Now that we have unit tests, we should also be made aware how bad the situation is.

Lets have Jenkins also calculate the code coverage.

psukys:
## Issue

 - fixes #568 

## Screenshot
https://server4.wdkk.de/job/TCA-Android/426/jacoco/

Having some heuristics to tell how well a codebase is maintained is beneficial in the long run

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -2,6 +2,7 @@ apply plugin: 'com.android.application'
 apply plugin: 'kotlin-android'
 apply from: '../findbugs.gradle'
 apply from: '../pmd.gradle'
+apply from: '../jacoco.gradle'
 
 android {
     compileSdkVersion 27
@@ -14,6 +15,7 @@ android {
             useProguard true
         }
         debug.debuggable true
+        debug.testCoverageEnabled true
         release.debuggable false
     }
     defaultConfig {
--- a/jacoco.gradle
+++ b/jacoco.gradle
@@ -3,7 +3,7 @@ apply plugin: 'jacoco'
 jacoco {
     // Use this version for upper ones are broken (with Gradle)
     // https://github.com/jacoco/jacoco/issues/288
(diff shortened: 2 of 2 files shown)


psukys:
Finally got Jenkins talking with current configuration, just need to find a way to actually parse tests (because current coverage is 0 :laughing: )

psukys:
Should be ready to review, @kordianbruck @pfent

psukys:
In short, current problem:
1. Newer jacoco has problems in discovering tests in our current system,
2. Older jacoco doesn't have problems, but is extremely (like 4-5 years old) old,
3. Jenkins uses newer jacoco plugin (for parsing) and versions are non back-compatible
4. Maybe possible to downgrade Jenkins, but then it's a just a nuissance.

psukys:
Thought: using alternative coverage tools? Jacoco has been for quite a while integrated as an official plugin into gradle, but if its acting up...

psukys:
This looks extremely promising:
https://bitbucket.org/atlassian/clover

Comparison:
https://confluence.atlassian.com/clover/comparison-of-code-coverage-tools-681706101.html

pfent:
IMHO getting a coverage percentage to display on the CI server just isn't worth the hazzle this seems to be.
Tests won't get better with that number, heck we're not even really trying to keep the lint warnings down :see_no_evil: 

Alas, if you want to have coverage numbers, just run it in AndroidStudio:
```
Run 'Tests in 'tumcampusapp'' with Coverage
```

kordianbruck:
@psukys yea, I'm not sure that works with android either. Doing Roboelectric test cases is really a weird thing.

@pfent yea, this has eaten up enough time as it is.

Issue #719, opened 2018-02-25
Roomfinder & LecturesPersonalActivity crashes app on Android 8


xsrf:
## Steps to Reproduce

1. Roomfinder
2. Search "n1190"
3. Tap first result

## Expected Result

Map of N1190

## Actual Result

App crashes (same with other rooms)
First two exceptions are thrown when you search for the room
Last exception is thrown when you click on the first result (which then crashes the app)
Also if you open the roomfinder and then minimize the app it'll throw the same exception and crash.

### Environment
* Phone: Nexus 5X
* OS version: 8.1.0
* TUM Campus App version: v1.5.3
* Language: de

```
02-25 23:29:26.837 22486-23716/de.tum.in.tumcampus E/ActivityThread: Failed to find provider info for de.tum.in.tumcampusapp.component.officialtum.roomfinder.RoomFinderSuggestionProvider
02-25 23:29:26.838 22486-23716/de.tum.in.tumcampus E/SearchSuggestions: saveRecentQuery
                                                                        java.lang.IllegalArgumentException: Unknown URL content://de.tum.in.tumcampusapp.component.officialtum.roomfinder.RoomFinderSuggestionProvider/suggestions
                                                                            at android.content.ContentResolver.insert(ContentResolver.java:1551)
                                                                            at android.provider.SearchRecentSuggestions.saveRecentQueryBlocking(SearchRecentSuggestions.java:201)
                                                                            at android.provider.SearchRecentSuggestions.-wrap0(Unknown Source:0)
                                                                            at android.provider.SearchRecentSuggestions$1.run(SearchRecentSuggestions.java:172)
02-25 23:29:26.838 22486-23716/de.tum.in.tumcampus E/ActivityThread: Failed to find provider info for de.tum.in.tumcampusapp.component.officialtum.roomfinder.RoomFinderSuggestionProvider
02-25 23:29:26.838 22486-23716/de.tum.in.tumcampus E/SearchSuggestions: truncateHistory
                                                                        java.lang.IllegalArgumentException: Unknown URL content://de.tum.in.tumcampusapp.component.officialtum.roomfinder.RoomFinderSuggestionProvider/suggestions
                                                                            at android.content.ContentResolver.delete(ContentResolver.java:1648)
                                                                            at android.provider.SearchRecentSuggestions.truncateHistory(SearchRecentSuggestions.java:243)
[...]

PR #726, opened 2018-02-27
workaround StickyListHeaders

jacqueline8711:
## Issue
This fixes the following issue(s): #719 

Android O users can use Roomfinder again


kordianbruck:
@pfent are we using this anywhere else? We have several of those autofills no? Isn't that the same things as the Suggestions?

pfent:
@kordianbruck We're using StickyListHeaders in several Places:
* GradesActivity
* LecturesPersonalActivity
* RoomFinderActivity
* BarrierFreeContectActivity
* BarrierFreeMoreInfoActivity
* ChatRoomsActivity
* CurriculaActivity

pfent:
The LecturesPersonalActivity is probably also affected by that crash, since there's a search bar in there.
@jacqueline8711 could you test this on your device?

pfent:
Oh, and maybe setting this via XML should be a bit more neat:
https://developer.android.com/reference/android/view/View.html#attr_android:importantForAutofill

jacqueline8711:
@pfent yeah, you're right, the same problem exists in the LecturesPersonalActivity

kordianbruck:
@jacqueline8711 cool - please fix that as well then. Also add a comment to the issue in our repo and a `TODO` marker, so we don't forget to remove it once its resolved in the lib

Issue #697, opened 2018-02-21
Leakcanary


kordianbruck:
cause why should not check for leaks in our app:

https://github.com/square/leakcanary

PR #728, opened 2018-02-27
Migrate all the AsyncTasks to RxJava

pfent:
There are several [pitfalls with AsyncTasks](http://blog.danlew.net/2014/06/21/the-hidden-pitfalls-of-asynctask/), most prominently the Activity leakage (see #697)

In this PR, I:
* removed all usages of AsyncTask
* bound the Activity lifecycle to the reactive observables with rxlifecycle, which ensures proper handling of device rotation etc.
* removed the ActivityAsyncTask used in the ExceptionHandler, since we used litterally 0 of the functionality it provided :man_shrugging: 

In a next step, we can probably remove a lot of the `doInBackground()` boilerplate and move everything to directly use Observables...

Tested everything and works like a charm :ok_hand:


kordianbruck:
@pfent are we sure, we want to have this in 1.5.X? I'd rather merge new features into 1.6 branch.

pfent:
I'd probably label this as concurrency bug fix, instead of new feature :wink:

Issue #723, opened 2018-02-25
Intrusive / wrong permission request


xsrf:
## Steps to Reproduce

On every start the app bothers me with "Diese App braucht die Erlaubnis auf deinen Kalender zuzugreifen, um **sie** diesen mit deinem TUM-Kalender abzugleichen." and after that it prompts to allow access to location. I obviously deny, because I don't know why it would need it. 

### Environment
* Phone: Nexus 5X
* OS version: 8.1.0
* TUM Campus App version: v1.5.3
* Language: de

kordianbruck:
@xsrf did you setup the sync with your calendar in our activity? In that case it will ask you for permissions in order to update it on app start.

xsrf:
No, I didn't ... Also the permission request "Diese App braucht die Erlaubnis auf deinen Kalender zuzugreifen, um sie diesen mit deinem TUM-Kalender abzugleichen." is clearly generated by you, not by the system, so the app is not really asking for permission anyway.

kordianbruck:
Well somehow that got enabled. No clue how - I will have to check on your phone on friday, but I've never received that request until I opted in to sync the calendar.

Yes, its our notification, that gets shown once you decline.

xsrf:
Refering to 8db4c54 you must have gotten the request at least on every new install, since it is (was) actually the explanatory text for the location request. You don't read your own annoying popups, do you? 🤣

kordianbruck:
@xsrf are you talking about location or calendar permission? Those are two pairs of shoes. Yes, we ask for location permission on each startup as we base some shown cards around that. (e.g. Mensa)

The Calendar permission only gets asked after you setup the sync.

kordianbruck:
I see - I think I get it now :dancer: 
Nevermind then - just the text was wrong in this case, if I understand correctly.

jacqueline8711:
@kordianbruck Yes, they were just switched around. I never noticed because I always use the app in english and those strings were always correct

kordianbruck:
@jacqueline8711 yea, me too :man_shrugging:

PR #729, opened 2018-02-27
Improve german permission strings

jacqueline8711:
## Issue
This fixes the following issue(s): #723

The german calendar and locations permission strings were mixed up.


kordianbruck:
@jacqueline8711 we should also just as the calendar permission, once a user sets up the sync in the activity. Can you still move that over please?

jacqueline8711:
@kordianbruck what do you mean? I don't really get what you're trying to say...

kordianbruck:
@jacqueline8711 nevermind, I was just checking andreas his answer - I'm unsure about what he actually means. The calendar permission only gets queried if you've opted in to sync your calendar. So this should not happen for every user.

Issue #701, opened 2018-02-22
Migrations still seem not to work


kordianbruck:
```
java.lang.IllegalStateException: Migration didn't properly handle Kino(de.tum.in.tumcampusapp.models.tumcabe.Kino).
Expected:
TableInfo{name='Kino', columns={rating=Column{name='rating', type='TEXT', notNull=true, primaryKeyPosition=0}, link=Column{name='link', type='TEXT', notNull=true, primaryKeyPosition=0}, trailer=Column{name='trailer', type='TEXT', notNull=false, primaryKeyPosition=0}, date=Column{name='date', type='TEXT', notNull=true, primaryKeyPosition=0}, actors=Column{name='actors', type='TEXT', notNull=true, primaryKeyPosition=0}, year=Column{name='year', type='TEXT', notNull=true, primaryKeyPosition=0}, cover=Column{name='cover', type='TEXT', notNull=true, primaryKeyPosition=0}, runtime=Column{name='runtime', type='TEXT', notNull=true, primaryKeyPosition=0}, genre=Column{name='genre', type='TEXT', notNull=true, primaryKeyPosition=0}, title=Column{name='title', type='TEXT', notNull=true, primaryKeyPosition=0}, description=Column{name='description', type='TEXT', notNull=true, primaryKeyPosition=0}, id=Column{name='id', type='TEXT', notNull=true, primaryKeyPosition=1}, director=Column{name='director', type='TEXT', notNull=true, primaryKeyPosition=0}, created=Column{name='created', type='TEXT', notNull=true, primaryKeyPosition=0}}, foreignKeys=[], indices=[]}
Found:
[...]

PR #730, opened 2018-02-27
Add test for database migration and versionize schema

pfent:
A simple unit test which applies the migrations on the pre-room database in-order.

I also wanted to use the [MigrationTestHelper](https://medium.com/google-developers/testing-room-migrations-be93cdb0d975), but Roboelectric just decided to be a PITA and fuck up the asset management, on which the helper relies. 

Now I don't automatically verify the schema, but just test, if the migrations succeed without exceptions :man_shrugging:


pfent:
@kordianbruck turns out, that all the room tests would had been Instrumentation tests, which can only be run on an emulator. Roboelectric simply can't possibly work with that...

I now simply instantiate our database, which correctly triggers migrations and validations of the current schema. I didn't reproduce #701 with that approach, yet. However that seems to be a matter of inserting the right test data into the database

kordianbruck:
@pfent yea, while trying it out yesterday it seemed waaaay too complicated to get it working for what we want it to do.

The Bug really only occures, if you install an app version after one of the first migrations so lets say version 3 and then execute two migrations on it. Somehow when setting it up, it is not using the migrations, but generating the schema from the models. So I really have no clue how to test this, as we cannot do intermediate migration tests

pfent:
@kordianbruck sure we can. Just extract the schema and set the version in `SQLiteOpenHelper`

psukys:
The unittest failure is quite cryptic, but in general I believe that it might be related to persisting TcaDb, where it has to be closed.
Reference test: https://github.com/TCA-Team/TumCampusApp/blob/master/app/src/test/java/de/tum/in/tumcampusapp/managers/SyncManagerTest.java#L37

pfent:
@psukys thanks for the hint, should be fixed now

Issue #711, opened 2018-02-22
Move to SDK 21


kordianbruck:
We currently have less than 1% using any of 4.1,4.2,4.3, about 150 alltogether. 
4.4 is still used by ~350 users. 

@pfent what do you think about increasing to API level 19 starting with our 1.6 release?

kordianbruck:
Also I've added a release branch: https://github.com/TCA-Team/TumCampusApp/tree/release/1.6

pfent:
[@minSdkVersion](https://twitter.com/minSdkVersion) agrees with you, but I don't see any new exciting features we can use with 14 -> 19

API 21 on the other hand would be more promising :rocket:

kordianbruck:
I don't really care for 4.4 users either. Those phones are like 3+ years old. 
We can keep the 1.5.X branch for those users and do bug fixes, but as for new features, sure, lets go with SDK 21.

PR #732, opened 2018-02-28
Move to sdk 21

kordianbruck:
## Issue

This fixes the following issue(s): #711 

@jacqueline8711 If you see any more simplifications that can be done with the eduroam code, go ahead.


pfent:
@kordianbruck can we do a Bugfix Release for that wifi crash prior to merging this?

kordianbruck:
@pfent I'm merging this to the 1.6 release branch - no worries. Current master will become 1.5 release branch

pfent:
ah, now we're doing release branches again... I'm certainly not a fan, since this usually creates more problems than it solves.

> Current master will become 1.5 release branch

then why don't branch release/1.5 and continue developing on master? That's how the kernel does it and what usually works best with git...

kordianbruck:
@pfent still waiting for some bug fixes. Will certainly switch it around, once I do the release.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/eduroam/EduroamFixCard.java
```diff
@@ -123,69 +125,68 @@ private boolean isConfigValid() {
...
+            checkDNSName();
+            checkAnonymousIdentity();
+            // note: checking the certificate does not seem possible
         }
+        // else: PWD or unknown authentication method (we don't know if that method is safe or not -> ignore)
+
         return errors.isEmpty();
     }
 
     @RequiresApi(api = Build.VERSION_CODES.JELLY_BEAN_MR2)
```
pfent: This RequiresApi can be removed, when we move to 21

pfent:
And since we're moving to API 21, consider removing dead code in:
- [ ] FavoriteDishAlarmScheduler
- [ ] ScanResultsAvailableReceiver
- [ ] Utils
- [ ] StartSyncReceiver
- [ ] IntegratedCalendarEvent
- [ ] EduroamController 
- [ ] SetupEduroamActivity
- [ ] TimetableWidgetService

just search for `Build.VERSION`

There are also some `@SuppressLint("NewApi")` and `@SuppressLint("InlinedApi")` Annotations, that need review

jacqueline8711:
@kordianbruck sorry this is such a mess
I was just trying to fix my username for the first three commits... 😬

kordianbruck:
@jacqueline8711 no worries :)

pfent:
Also, when moving to 21, we could also drop the PDFView dependency:

>Also, an inline display for the PDFs would be nice to have. There is a [PDFRenderer](https://developer.android.com/reference/android/graphics/pdf/PdfRenderer.html) available for API 21+, even with an exhaustive [example on GitHub](https://github.com/googlesamples/android-PdfRendererBasic). Then you can use the fileprovider as a fallback for lower APIs.

[AndroidPdfViewer](https://github.com/barteksc/AndroidPdfViewer) comes with PdfiumAndroid, which is ~16MB

PR #733, opened 2018-02-28
Trivial checkstyle fixes

psukys:
## Issue

With introduction of `checkstyle`, a lot of warnings were reported back regarding. This PR attempts to fix the trivial ones that require mostly formatting based fixes.


psukys:
Change of plans, rather do module type of fixes than warning/error type

PR #734, opened 2018-03-01
Fix tests that use Joda datetime

### LINKED ISSUE #640: NewsDaoTest.cleanUpOldTest sometimes failing
see the build server:
https://server4.wdkk.de/job/TCA-Android/269/testReport/de.tum.in.tumcampusapp.database.dao/NewsDaoTest/cleanUpOldTest/

psukys:
## Issue

Sometimes test that uses Joda DateTime fails (as far as I could tell, when tests are run around midnight). As well added intializations to parts of code that contain joda's datetime usage.

## Screenshot
Not applicable

Having rogue test at midnight is not something anyone wants.

Fixes #640

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/ui/cafeteria/details/CafeteriaDetailsSectionFragment.java
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/ui/cafeteria/details/CafeteriaDetailsSectionFragment.java
@@ -16,6 +16,8 @@ import android.view.ViewGroup;
 import android.widget.LinearLayout;
 import android.widget.TextView;
 
+import net.danlew.android.joda.JodaTimeAndroid;
+
 import org.joda.time.DateTime;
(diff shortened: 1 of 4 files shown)


kordianbruck:
@psukys which test did you fix that was running during midnight? The diff does not show anything?

psukys:
@kordianbruck the test logic was alright, the problem was with the way the test was setup - joda timezone info is not passed and at this point I'm not sure what the default behaviour is. The problem appeared with `NewsDaoTest.cleanUpOldTest`, which used Joda's `DateTime` which [needed library's initialization](https://stackoverflow.com/a/31775313/552214). I quickly looked through which act can be nicer - eradicating joda time (as it wasn't used in so many places anyway) or using initialization. 
Joda's library is nice for test, because I can nicely and understandably (i.e. `DateTime.now().minusMonths(3)`) manipulate date; and Cafeteria code used a custom format and several other dependencies from Joda, which I decided not to touch, cause it would in the end require more checking.

PR #737, opened 2018-03-03
Add Kotlin static source analysis checks

psukys:
We're moving to Kotlin, so let's have it as clean as possible.


psukys:
Integration with Jenkins is only possible through extending a plugin that we're using for Github reports. There's no jenkins plugin for detekt yet.

psukys:
@kordianbruck , @pfent , should be ready for review. I've disabled build-fail on cases where duplicates are found.

app/build.gradle
```diff
@@ -1,9 +1,14 @@
+plugins {
+    id 'io.gitlab.arturbosch.detekt' version "1.0.0.RC6-3"
```
kordianbruck: Can't this be moved to the detekt.gralde file?

detekt-config.yml
```diff
@@ -21,7 +21,6 @@ test-pattern: # Configure exclusions for test sources
 
 build:
   warningThreshold: 5
-  failThreshold: 10
```
kordianbruck: Why was this removed?

psukys: On similar pace (as other checks) that violations are only reported but do not fail the build process.

app/build.gradle
psukys: I can check, tho I remember that gradle complained before that `plugins` needs to be topmost instruction

PR #746, opened 2018-03-27
Add Crashlytics

kordianbruck:
Housekeeping our own bug reports on our server is really hard. The system works fine, but its not perfect. Firebase actually has a pretty interface to manage those bug reports/stack traces.

Also we did not get reports from background services and other threads, where we did not hook the exception handler - this has been bothering me for a long time.


pfent:
I think you're still missing some steps from the [install instructions](https://fabric.io/kits/android/crashlytics/install)

Did I miss the "Initialize Your Kit" step in the diff?

pfent:
Also: Since Crashlytics is a Google, on the privacy policy side, we're probably good to go with the PlayStore policy?

kordianbruck:
@pfent we are using firebase, not crashlytics directly - whatever the difference is: https://firebase.google.com/docs/crashlytics/get-started

PR #747, opened 2018-03-27
Merge Shared Preferences

### LINKED ISSUE #694: Merge Shared Preferences
Currently we have two shared preferences.
- Default
- Internal

I vouch to merge those two, just to have less confusion on which one to use. That way we can remove some functions from the Utils Class.

kordianbruck:
Whyever we had two shared prefs - this PR should merge them both into one unified SP.

Fixes #694


app/src/main/java/de/tum/in/tumcampusapp/utils/Utils.java
```diff
@@ -50,6 +51,43 @@ private Utils() {
...
+            }
+            if (object instanceof Integer) {
+                prefsEditor.putInt(entry.getKey(), (Integer) object);
+            }
+            if (object instanceof Float) {
+                prefsEditor.putFloat(entry.getKey(), (Float) object);
+            }
+            if (object instanceof Long) {
+                prefsEditor.putLong(entry.getKey(), (Long) object);
+            }
```
pfent: Ain't nobody got Double values :see_no_evil:

app/src/main/java/de/tum/in/tumcampusapp/utils/Utils.java
```diff
@@ -96,15 +134,78 @@ public static String cutText(String text, String startString, String endString)
...
+     * @return setting value, defaultVal if undefined
      */
     public static String getSetting(Context c, String key, String defaultVal) {
         SharedPreferences sp = PreferenceManager.getDefaultSharedPreferences(c);
         return sp.getString(key, defaultVal);
     }
 
     /**
-     * Get a value from the default shared preferences
+     * Get a value from the default shared preferences.
```
pfent: All those shared preference accesses probably can be replaced with [Android KTX](https://github.com/android/android-ktx) in the near future...

app/src/main/java/de/tum/in/tumcampusapp/utils/Utils.java
kordianbruck: Nope :dagger:

app/src/main/java/de/tum/in/tumcampusapp/utils/Utils.java
kordianbruck: You still have to call `.edit` with that, but yea - it still will be some time till that reaches general availability.

kordianbruck:
@pfent ready to merge then?

pfent:
sure the migration path looks good to me, we only need to think about when we should remove that for future versions

PR #753, opened 2018-04-05
Remove old caching behaviour of the curricula

kordianbruck:
Found some artifacts from ancient times...


app/src/main/java/de/tum/in/tumcampusapp/api/app/TUMCabeClient.java
```diff
@@ -234,7 +236,7 @@ public void deviceUploadGcmToken(DeviceUploadGcmToken verification, Callback<TUM
                .enqueue(cb);
     }
 
-    public void createMeasurements(List<WifiMeasurement> wifiMeasurementList, Callback<TUMCabeStatus> cb) throws IOException {
```
pfent: @kordianbruck there are still errors in RoomFinderDetailsActivity, because that IOException isn't declared anymore...

kordianbruck: Gradle incremental builds sometimes aren't fun... Thanks for catching this.

Issue #480, opened 2017-06-12
Default station "Theresienstraße" does not show any results


xsrf:
## Steps to Reproduce

1. Install TCA
2. Be around main campus
3. Item "Theresienstraße" should show up on the main screen by default
4. Tap on "Theresienstraße"

OR

1. Tap "Menu"
2. Tap "MVV"
3. Search "Theresienstraße"
4. Tap "München, Theresien"

## Expected Result

See trains departing Theresienstraße

## Actual Result

Blank screen

## Problem

No results are shown. Looks like "ß" is a really bad character ...

## Suggested Enhancement :sparkles:

Handle "ß" properly

### Environment
* Phone: Nexus 5X
* OS: Android 7.1.2
* TUMOnline App: v1.4.8
* Language: de

pfent:
Should actually have been fixed with #528, is it not?

kordianbruck:
@pfent nope, there is even an ToDo in the code :stuck_out_tongue:

PR #754, opened 2018-04-05
Fix transportation manager not finding station

kordianbruck:
## Issue
This fixes the following issue(s): #480 

## Screenshot
screenshot_20180405-142218

It just works


app/src/main/java/de/tum/in/tumcampusapp/component/other/locations/LocationManager.java
```diff
@@ -59,13 +60,18 @@
     };
     private static final StationResult[] DEFAULT_CAMPUS_STATION = {
             new StationResult("Garching-Forschungszentrum", "1000460", Integer.MAX_VALUE),
+            new StationResult("Garching-Technische Universität", "1002070", Integer.MAX_VALUE),
```
jacqueline8711: It seems like this causes an issue because the indices don't line up anymore. For me it shows 'Weihenstephan' although I am at 'Stammgelände' which has index 4 in the array CAMPUS_SHORT. In this array on index 4 ist 'Weihenstephan'

pfent: :man_facepalming: 
We should probably use a map or something, so we don't need to throw indices around?

kordianbruck: Not as easy, as you would have to map the relationsships somehow using IDs or something. Feel free to suggest a better way @pfent

pfent:
@kordianbruck have a look

Issue #742, opened 2018-03-16
Rework Wizard / App Setup


kordianbruck:
We want to make onboardinf for new users easier. In order to do that, we want to make the Wizard optional and less blocky.
The following should be implemented:

- [ ] Directly start into the App, don't show the wizard before first load
- [ ] Hide menu entries, that require the TUM token (Should already be the case with a boolean flag)
- [ ] When not logged in, show a 'Login to TUM' menu entry, that will start the wizard with the existing three step proccess
- [ ] Add a card, that is shown on the very top after installing, that additionally prompts / asks the user to login / activate the token

jacqueline8711:
- [ ] show the benefits of logging into tum (e.g. in wizard below the tum id text field)

PR #755, opened 2018-04-05
Modified App Setup

jacqueline8711:
## Issue

This fixes the following issue(s): #742 

## Screenshot
screenshot_20180405-143638
screenshot_20180405-143642
screenshot_20180405-150246

Easier/faster for users who don't want or can't log into TUMonline or for those that just don't want to enable the Token on the go


app/src/main/java/de/tum/in/tumcampusapp/component/ui/onboarding/LoginPromtCard.java
```diff
@@ -0,0 +1,61 @@
...
+import android.view.ViewGroup;
+
+import de.tum.in.tumcampusapp.R;
+import de.tum.in.tumcampusapp.component.ui.overview.CardManager;
+import de.tum.in.tumcampusapp.component.ui.overview.card.Card;
+import de.tum.in.tumcampusapp.utils.Const;
+import de.tum.in.tumcampusapp.utils.Utils;
+
+/**
+ * Card that describes how to dismiss a card
```
pfent: Nope, that's not what that card does :)

Maybe add the reasoning from #742 here?

app/src/main/java/de/tum/in/tumcampusapp/component/ui/onboarding/StartupActivity.java
```diff
@@ -88,37 +87,6 @@ private void init() {
...
-        // For compatibility reasons: big update happened with version 35
-        int prevVersion = Utils.getSettingInt(this, Const.APP_VERSION, 35);
-
-        // get current app version
-        int currentVersion = Utils.getAppVersion(this);
-        boolean newVersion = prevVersion < currentVersion;
-        if (newVersion) {
-            this.setupNewVersion();
-            Utils.setSetting(this, Const.APP_VERSION, currentVersion);
-        }
```
pfent: what happened to `setupNewVersion()`?
I think we can either completely remove it then?

app/src/main/java/de/tum/in/tumcampusapp/component/ui/onboarding/WizNavCheckTokenActivity.java
```diff
@@ -33,22 +33,9 @@ public void onCreate(Bundle savedInstanceState) {
...
-
-    /**
-     * If back key is pressed start previous activity.
-     */
-    @Override
-    public void onBackPressed() {
-        finish();
-        startActivity(new Intent(this, WizNavStartActivity.class));
-        overridePendingTransition(R.anim.fadein, R.anim.fadeout);
+    public void onBackPressed(){
```
pfent: I think this is missing an `@Override`

app/src/main/java/de/tum/in/tumcampusapp/component/ui/onboarding/StartupActivity.java
jacqueline8711: I'm not really sure, we probably can. What do you think @kordianbruck ?

kordianbruck: We can probably get rid of that as well. Just a question of if we already want to skip resetting the DB. I had to trigger a destructive migration for the current 1.5.6 release, as we were getting many crash reports from users.

PR #759, opened 2018-04-05
Add our stylish background to the main screen

kordianbruck:
screenshot_20180405-173347


kordianbruck:
Or is it 'too loud'?

pfent:
maybe add something like
```xml
android:alpha="0.5"
```
so it is a little bit more subtle. What do you think?

kordianbruck:
screenshot_20180406-020442

better?

PR #766, opened 2018-04-11
bugfix permission calendar

jacqueline8711:
## Issue
CalendarActivity.java line 281
Fatal Exception: java.lang.IllegalArgumentException
requestCode should be >= 0

occurring on Android 8


jacqueline8711:
@kordianbruck this only occurred when the permission was originally denied, but most people just give the permission. So usually it was no problem

PR #767, opened 2018-04-11
 Crashes in Android 4 caused by drawableLeft/Right 

jacqueline8711:
!! only affects Android 4, should not go into master !!

## Issue
DrawableLeft/Right with a vector drawable in an xml layout file made the app crash on Android 4
-> wrappers introduced around the drawables as suggested in https://stackoverflow.com/a/35761799 (worked on the emulator)

Students without Android 4 phones can now view the lecture list, the chat and the details of events again! And those who didn't have eduroam configured (or incorrectly configured) can use the app again without crash on startup


kordianbruck:
@jacqueline8711 cool, I'll have to cherry pick fixes then from/to this branch then - not a problem.

Issue #764, opened 2018-04-09
Kino - wrong ordering of movies


jacqueline8711:
## Steps to Reproduce
click on Kino news card (showing correct movie) to get to KinoActivity

## Expected Result
Movies should be ordered by date and match the title just below the action bar

## Actual Result
The ordering in the tab bar (below action bar) is correct but the wrong movie is displayed

### Environment
* TUM Campus App version: 1.5.8 (from Play Store)

![screenshot_20180408-164007](https://user-images.githubusercontent.com/16260112/38503203-c1a4766e-3c11-11e8-9404-abf21c3cd21c.png)
![screenshot_20180408-164055](https://user-images.githubusercontent.com/16260112/38503202-c177da14-3c11-11e8-8422-3d520bc68620.png)

PR #771, opened 2018-04-13
Fix movie ordering

jacqueline8711:
## Issue
This fixes the following issue(s): #764 

I don't really know why it doesn't work with the order by statement, but omitting it is fixing this somehow

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/ui/tufilm/KinoDao.java
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/ui/tufilm/KinoDao.java
@@ -23,13 +23,13 @@ public interface KinoDao {
     @Insert(onConflict = OnConflictStrategy.REPLACE)
     void insert(Kino kino);
 
-    @Query("SELECT * FROM kino")
+    @Query("SELECT * FROM kino ORDER BY date")
     Flowable<List<Kino>> getAll();
 
     @Query("SELECT id FROM kino ORDER BY id DESC LIMIT 1")
     Maybe<String> getLastId();
 
-    @Query("SELECT * FROM kino ORDER BY id LIMIT 1 OFFSET :position")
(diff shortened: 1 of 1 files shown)


app/src/main/java/de/tum/in/tumcampusapp/component/ui/tufilm/KinoDao.java
```diff
@@ -29,7 +29,7 @@
     @Query("SELECT id FROM kino ORDER BY id DESC LIMIT 1")
     Maybe<String> getLastId();
 
-    @Query("SELECT * FROM kino ORDER BY id LIMIT 1 OFFSET :position")
+    @Query("SELECT * FROM kino LIMIT 1 OFFSET :position")
```
pfent: `LIMIT` without `ORDER BY` doesn't specify which element it returns.
Maybe we want to `ORDER BY date` instead

jacqueline8711: @pfent Yes, ordering by date should work for now. And I finally get why ordering by id does not work (although the ids are in the right order, we get them from the api). The problem is that the ids are strings, so the Jumanji movie had id '100' which is smaller than '99' etc. 😣

app/src/main/java/de/tum/in/tumcampusapp/component/ui/tufilm/KinoDao.java
```diff
@@ -23,13 +23,13 @@
     @Insert(onConflict = OnConflictStrategy.REPLACE)
     void insert(Kino kino);
 
-    @Query("SELECT * FROM kino")
+    @Query("SELECT * FROM kino ORDER BY date")
     Flowable<List<Kino>> getAll();
 
     @Query("SELECT id FROM kino ORDER BY id DESC LIMIT 1")
```
jacqueline8711: I'm not sure if I should change this to ORDER BY DATE too

Issue #695, opened 2018-02-21
Utilize Picasso for ImageViewing/Caching


kordianbruck:
https://square.github.io/picasso/

Its huge, its great, its gonna make us great again :us:

psukys:
yes please

PR #775, opened 2018-04-20
Picasso

jacqueline8711:
## Issue
This fixes the following issue(s): #695 

Roomfinder now caches recent results


app/src/main/java/de/tum/in/tumcampusapp/component/ui/news/NewsAdapter.java
```diff
@@ -111,11 +111,16 @@ public void onError(Exception e) {
 
     @Override
     public CardViewHolder onCreateViewHolder(ViewGroup parent, int viewType) {
-        return NewsCard.inflateViewHolder(parent, viewType);
+        if(viewType == 0){
```
psukys: `return newNewsView(parent, viewType == 0)` ?

jacqueline8711: Thanks!

Issue #786, opened 2018-04-29
Same background across all activities


kordianbruck:
I would suggest a unification of the UIs / Activity views across all views. Right now we have cards & bubbles over the really nice faculty background on the main and chat screen. I would suggest to use a similar setup in all activities where possible. 

For example the feedback activity currently uses just a plain simple grey background and no white card / box. Many other activities do not follow any guideline. 

Ideas? Suggestions? Feelings?

kordianbruck:
We discussed this again and it seems that in many activities this would add too much noise. We could think of adding it to the ChatActivity maybe, but for now lets postpone this issue indefinitely, as its not a priority.

PR #800, opened 2018-05-04
feedback layout

jacqueline8711:
## Issue
#786 

## Screenshot
screenshot_20180504-104918


kordianbruck:
The margins (for text) from the screensides are different than other activities. It would be nice, if we'd follow material design guidelines here.

jacqueline8711:
@kordianbruck margins and paddings should be more material now

PR #801, opened 2018-05-04
Make person details look fancy

thellmund:
## Issue
- Design of PersonDetailsActivity looked outdated
- Touch targets (links, e-mail addresses, phone numbers) weren’t big enough
- Rooms couldn’t be opened right from PersonDetailsActivity

This gives students: 
- More intuitive access to person information
- Ability to more easily find a person’s room

## Screenshot
screenshot_20180504-125702

## Steps performed
I took the following steps: 
1. Add dependency for 3rd-party framework for circular ImageView
2. Migrate UI elements in PersonDetailsActivity to something more in line with Material Design
3. Convert PersonDetailsActivity to Kotlin
4. Add PersonContactItemsAdapter, which is responsible for displaying all ways of contacting a person
5. Add AbstractContactItem and subclasses EmailContactItem, HomepageContactItem, InformationContactItem, MobilePhoneContactItem,  OfficeHoursContactItem, PhoneContactItem, RoomContactItem to encapsulate the contact information and the associated actions (phone -> call, homepage -> open in browser, room -> open room finder)
6. Add PersonGroupsAdapter, which is responsible for displaying a person’s groups (consisting of function and org)
7. Made changes that required other changes in PersonSearchActivity. As a result, switch from ListView to RecyclerView and convert entire class to Kotlin
8. Move code to save a person to phone contacts to a dedicated ContactsManager class


app/build.gradle
```diff
@@ -122,6 +122,8 @@ dependencies {
     implementation 'com.android.support.constraint:constraint-layout:1.0.2'
     implementation 'com.crashlytics.sdk.android:crashlytics:2.9.1'
     implementation 'com.squareup.picasso:picasso:2.71828'
+    implementation 'de.hdodenhof:circleimageview:2.2.0'
```
jacqueline8711: We already had that library in the app, see line 107

thellmund: Oops, fixed

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/person/PersonDetailsActivity.kt
```diff
@@ -0,0 +1,210 @@
...
+        }
+    }
+
+    /**
+     * Adds the given employee to the users contact list
+     *
+     * @param employee Object to insert into contacts
+     */
+    private fun addContact(employee: Employee?) {
+        if (!isPermissionGranted(0)) {
```
kordianbruck: Please make the `0` a constant inside `Const.kt` so we avoid clashes and don't have magic constants throughout random classes.

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/person/PersonDetailsActivity.kt
```diff
@@ -0,0 +1,210 @@
...
+    /**
+     * Check Contacts permission for Android 6.0
+     *
+     * @param id the request id
+     * @return If the contacts permission was granted
+     */
+    private fun isPermissionGranted(id: Int): Boolean {
+        if (ContextCompat.checkSelfPermission(this, Manifest.permission.WRITE_CONTACTS) == PackageManager.PERMISSION_GRANTED) {
+            return true
+        } else {
```
kordianbruck: You don't need an `else` here, as the first part of the condition returns in case it is true.

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/person/PersonDetailsActivity.kt
```diff
@@ -0,0 +1,210 @@
...
+        }
+    }
+
+    /**
+     * Check Contacts permission for Android 6.0
+     *
+     * @param id the request id
+     * @return If the contacts permission was granted
+     */
+    private fun isPermissionGranted(id: Int): Boolean {
```
kordianbruck: This function looks like it can be generalized and move to the utils package - we have a similar setup in other classes I think. For now its fun, just as a future reference.

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/person/PersonSearchActivity.kt
```diff
@@ -0,0 +1,104 @@
...
+        if (response.persons.size == 1) {
+            showPersonDetails(response.persons.first())
+        } else {
+            val adapter = personsRecyclerView.adapter as? PersonSearchResultsAdapter
+            adapter?.update(response.persons)
+        }
+    }
+
+    companion object {
+        private const val P_SUCHE = "pSuche"
```
kordianbruck: Please move this to `Const.kt`

app/src/main/java/de/tum/in/tumcampusapp/utils/ContactsManager.kt
```diff
@@ -0,0 +1,188 @@
...
+import android.os.RemoteException
+import android.provider.ContactsContract
+import de.tum.`in`.tumcampusapp.R
+import de.tum.`in`.tumcampusapp.component.tumui.person.model.Contact
+import de.tum.`in`.tumcampusapp.component.tumui.person.model.Employee
+import java.io.ByteArrayOutputStream
+import java.io.IOException
+import java.util.*
+
+class ContactsManager {
```
kordianbruck: Could we please rename this to `ContactHelper` as we have `Managers` already used in this project for controller like classes.

thellmund: Fixed

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/person/PersonDetailsActivity.kt
thellmund: Good point. Switching to something like RxPermissions might be an option, as this removes a lot of the boilerplate code

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/person/PersonDetailsActivity.kt
thellmund: Fixed

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/person/PersonDetailsActivity.kt
thellmund: Fixed

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/person/PersonDetailsActivity.kt
kordianbruck: Feel free to open up an issue with this suggestion! Looked at it briefly and it looks nice.

Issue #768, opened 2018-04-11
Allow users to customize/filter the calendar


jacqueline8711:
- [x] whether to show moved/canceled events 
- [x] hide weekend (possible?)
- [ ] show timespan e.g. only 8am to 8pm according to user input (possible with #630 )

related: #408

kordianbruck:
![unnamed](https://user-images.githubusercontent.com/298860/39406704-377fd8d4-4bbb-11e8-9538-ed2083ece842.png)


Example for a filter option (top right)

kordianbruck:
related: #408

PR #816, opened 2018-05-14
768 user custom filters calendar

schulzalexander:
## Issue
#768 
Implemented filters in the calendar top menu for:
- show/hide canceled events
- fit the calendar to screen, i.e. show all 24 hours without scrolling
## Screenshot
bildschirmfoto 2018-05-14 um 23 15 05

## Still to be implemented
filter to only show hours from 8am to 8pm, not sure if this is possible without modifying the WeekView class...


kordianbruck:
@schulzalexander please only implement what is easy to do - modifying the weekview class sounds like a lot of additional work and maintenance complexity. We might be switching calendar libraries frequently, as none of them are really offical.

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarActivity.java
```diff
@@ -367,6 +403,7 @@ private void deleteCalendarFromGoogle() {
             calendar.set(Calendar.DAY_OF_MONTH, curDay);
             List<CalendarItem> calendarItems = calendarController.getFromDbForDate(new Date(calendar.getTimeInMillis()));
             for (CalendarItem calendarItem : calendarItems) {
+                if (!showCanceledEvents && calendarItem.getStatus().equals("CANCEL")) continue;
```
kordianbruck: There is no need to use `continue` here - just do the normal `if` with brackets and it should be all good.

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarActivity.java
```diff
@@ -483,4 +520,40 @@ protected void onResume(){
...
+        showCanceledEvents = val;
+        onResume();
+    }
+
+    protected void applyFilter8to8(boolean val) {
+        return;
+    }
+
+    protected void applyFilterFitScreen(boolean val) {
+        int hourHeight = 0;
```
kordianbruck: Please initialize this variable with `defaultHourHeight`. This way you don't need the `else` branch below.

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarActivity.java
```diff
@@ -483,4 +520,40 @@ protected void onResume(){
...
+        if (val) {
+            hourHeight = (mWeekView.getMeasuredHeight()
+                          - mWeekView.getTextSize()
+                          - (2*mWeekView.getHeaderRowPadding()))
+                         / 24;
```
kordianbruck: This calculation would be an excellent example for a useful comment to be required.

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarActivity.java
```diff
@@ -74,14 +74,26 @@
     private CalendarController calendarController;
     private final LifecycleProvider<Lifecycle.Event> provider = AndroidLifecycle.createLifecycleProvider(this);
 
+    /**
+     * Variables for filtering calendar events
+     */
+    private boolean showCanceledEvents = true;
```
kordianbruck: Why does this need to be a member variable of the class? 

Better just fetch the value from shared preferences once its needed and have less state inside the object wide access (=> usually means less bugs)

Up to you - just a suggestion.

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarActivity.java
pfent: With kotlin, this would just be `.filter().map()` :see_no_evil:

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarActivity.java
schulzalexander: Agree and changed it, I only thought that Utils would always fetch it from storage and that it would be nicer to have it in RAM.

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarActivity.java
schulzalexander: @kordianbruck agree, changed

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarActivity.java
schulzalexander: @kordianbruck agree, changed

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarActivity.java
schulzalexander: @kordianbruck I added the comments, it's basically just calculating the height of the weekView header, subtract it from the total height and div by 24 to get height per hour.

schulzalexander:
@kordianbruck ok, I removed the "8am to 8pm" filter and the "hide weekend" filter, they're as far as I experienced not possible with the weekView as is.

app/src/main/res/values/strings.xml
```diff
@@ -265,7 +268,7 @@
     <string name="cafeteria_notification_settings">Cafeteria Notification Settings</string>
 
     <!-- Cafeteria Notification Settings -->
-    <string name="invalid_notification_time">Invalid time. Please pick a time ranging from:</string>
+    <string name="invalid_notification_time">Invalid time. Please pick a ctime ranging from:</string>
```
pfent: ctime?

jacqueline8711:
@schulzalexander Showing only a certain timespan is actually very easily achievable with our new weekView fork. You can simply use 
```
mWeekView.setMinTime(8);
mWeekView.setMaxTime(24);
```
if you want the calendar to display everything between 8-24 o'clock (you could also let the user choose the hours to be displayed)

or maybe even better is the option ```mWeekView.setAutoLimitTime(true);```  which automatically sets the min and max time so that all events are displayed.
Refer to https://github.com/Quivr/Android-Week-View/wiki/Customization for more information

schulzalexander:
Just added a dialog for the user to select upper and lower bound of the timespan to be shown in the calendar.
bildschirmfoto 2018-05-18 um 14 56 45

kordianbruck:
@schulzalexander maybe "From hour" and "To hour" would be more descriptive of what it is the user is choosing?

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarActivity.java
```diff
@@ -516,6 +519,10 @@ protected void initFilterCheckboxes() {
         settings = Utils.getSettingBool(this, Const.CALENDAR_FILTER_CANCELED, true);
         menuItemFilterCanceled.setChecked(settings);
         applyFilterCanceled(settings);
+
+        int savedMin = Integer.parseInt(Utils.getSetting(this, Const.CALENDAR_FILTER_HOUR_LIMIT_MIN, "0"));
+        int savedMax = Integer.parseInt(Utils.getSetting(this, Const.CALENDAR_FILTER_HOUR_LIMIT_MAX, "24"));
```
pfent: I don't think 0 to 24 are good defaults.
I'd prefer defaulting to something like 8 to 20
Or put some intelligence in it and calculate the hours actually needed to display all lectures...

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarActivity.java
```diff
@@ -526,14 +533,43 @@ protected void applyFilterCanceled(boolean val) {
     protected void applyFilterFitScreen(boolean val) {
         Utils.setSetting(this, Const.CALENDAR_FILTER_FIT_SCREEN, val);
         int hourHeight = defaultHourHeight;
+        int minHour = Integer.parseInt(Utils.getSetting(this, Const.CALENDAR_FILTER_HOUR_LIMIT_MIN, "0"));
+        int maxHour = Integer.parseInt(Utils.getSetting(this, Const.CALENDAR_FILTER_HOUR_LIMIT_MAX, "24"));
```
pfent: And maybe avoid duplicating *magic numbers*

schulzalexander:
I've applied the requested changes, finished from my side.

kordianbruck:
@schulzalexander please ask one of your team members to review your code. Also please merge the `master` into your branch, so there are no conflicts.

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/LimitPickerDialog.java
```diff
@@ -0,0 +1,106 @@
...
+
+
+
+class LimitPickerDialog extends Dialog {
+    private Button cancelButton;
+    private Button saveButton;
+    private NumberPicker numberPickerMin;
+    private NumberPicker numberPickerMax;
+    protected Context context;
+    private List<LimitPickerDialogListener> listeners = new ArrayList<LimitPickerDialogListener>();
```
saileru: I think, we can leave out the generic type on the right side.
Since JDK 7 it should be sufficient to write:
private List\<LimitPickerDialogListener\> listeners = new ArrayList<>()

app/src/main/res/values/strings.xml
```diff
@@ -26,6 +26,13 @@
     <string name="calendar_not_existing_toast">No local calendar existing</string>
     <string name="semester_summer">Summer term</string>
     <string name="semester_winter">Winter term</string>
+    <string name="calendar_filter_title">Change Filters</string>
+    <string name="calendar_filter_canceled">Show Canceled Events</string>
+    <string name="calendar_filter_fit_screen">Fit Screen</string>
+    <string name="calendar_filter_hour_limit">Limit Shown Timespan</string>
+    <string name="calendar_filter_hour_limit_min_title">From Hour</string>
+    <string name="calendar_filter_hour_limit_max_title">To Hour</string>
+    <string name="calendar_filter_hour_limit_error_message">Minimum hour must be below maximum hour!</string>
```
saileru: We should probably also add German versions for the strings to de\strings.xml

saileru:
The usage of the checkbox "Fit Screen" seems a bit unintuitive to me.

1. I check Fit Screen.
2. I zoom, so that not the whole range is displayed.
-> Fit Screen is still checked, but not the whole screen is shown.

Should we maybe disable zooming when "Fit Screen" is checked?

schulzalexander:
@saileru I adapted the code to fit your suggestions.

saileru:
Thank you for the improvements. The German translations seem fine!

The disabling of zooming for "Fit Screen" mode also works
However, when I try to zoom with "fit screen" enabled, there is some shaking in the view which is not so nice.
But I do not see a better way to implement it except maybe remove the "Fit Screen" option and leave the user the freedom to zoom as he prefers within the set limits.

I am sorry to bother you again on this issue...

I also discovered a crash. How to reproduce:
1. Set Time Span to e.g. 8 to 15.
2. Change "From Hour" to something greater 15
=> Crash:
E/AndroidRuntime: FATAL EXCEPTION: main
                  Process: de.tum.in.tumcampus, PID: 25562
                  java.lang.IllegalArgumentException: startHour must smaller than endHour
                      at com.alamkanak.weekview.WeekView.setMinTime(WeekView.java:2208)
                      at de.tum.in.tumcampusapp.component.tumui.calendar.CalendarActivity.applyFilterLimitHours(CalendarActivity.java:566)
                      at de.tum.in.tumcampusapp.component.tumui.calendar.CalendarActivity.onSelected(CalendarActivity.java:582)
                      at de.tum.in.tumcampusapp.component.tumui.calendar.LimitPickerDialog.notifyListeners(LimitPickerDialog.java:103)
                      at de.tum.in.tumcampusapp.component.tumui.calendar.LimitPickerDialog.access$200(LimitPickerDialog.java:22)
                      at de.tum.in.tumcampusapp.component.tumui.calendar.LimitPickerDialog$2.onClick(LimitPickerDialog.java:70)
                      at android.view.View.performClick(View.java:5293)
                      at android.view.View$PerformClick.run(View.java:21653)
                      at android.os.Handler.handleCallback(Handler.java:815)
                      at android.os.Handler.dispatchMessage(Handler.java:104)
                      at android.os.Looper.loop(Looper.java:227)
                      at android.app.ActivityThread.main(ActivityThread.java:6102)
                      at java.lang.reflect.Method.invoke(Native Method)
                      at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:961)
                      at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:822)

schulzalexander:
@saileru thanks for your review, I finally removed the filter since I couldn't find a way to prevent that inconvenient behaviour. Also I removed the bug that you pointed out.

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarActivity.java
```diff
@@ -534,40 +525,20 @@ protected void applyFilterCanceled(boolean val) {
         onResume();
     }
 
-    protected void applyFilterFitScreen(boolean val) {
```
saileru: I was thinking if it might be a good idea to leave these funcions in but rename applyFilterFitScreen(val) to just fitScreen(). This method can then be called when the user selects new time limits.

Leave minHourHeight and maxHourHeight to default values, but set hourHeight so that the whole selected range is shown.

Selecting a time limit and then zooming out manually might be tedious...

schulzalexander: good idea, changed it ;)

saileru:
I do not know how picky I should be ^^
You could change the method name from hourheightFitScreen to hourHeightFitScreen.
There is a typo in the comment in calcHourHeightToFit: devide -> divide
And you could make the indentation of the comments in the return statement of calcHourHeightToFit consistent.

Then I am happy :D

schulzalexander:
@saileru applied ;)

kordianbruck:
@saileru excellent work. Be as picky as possible ;)

Issue #762, opened 2018-04-06
Chat: add user via search


kordianbruck:
We want to be able to add users to chat rooms by searching for `first or last name` or the `full TUM ID` (Use a regex to match this format). For this we need a new endpoint on the backend to search for those things in the `member` table. 

The user can search for other users and add them to any chat room.

PR #822, opened 2018-05-17
add user to chat room via search

jacqueline8711:
## Issue
This fixes the following issue(s): #762 

## TODO
- [x] extract String resources
- [x] add a delay before querying the server for auto-complete suggestions
- [x] keep QR-Code as a second option or remove code ( @kordianbruck what do you think?)
- [x] resolve conflicts with master

## Server
relies on PR on server side (https://github.com/kordianbruck/tca-backend/pull/20)
(already deployed)

## Screenshot
screenshot_add_to_chat

scanning a QR-code is usually not practical


app/src/main/java/de/tum/in/tumcampusapp/component/ui/chat/AddChatMemberActivity.java
```diff
@@ -114,20 +122,7 @@ public void onFailure(Call<ChatMember> call, Throwable t) {
...
-                        suggestions = response.body();
-                        ((MemberSuggestionsListAdapter)searchView.getAdapter()).updateSuggestions(suggestions);
-                    }
-
-                    @Override
-                    public void onFailure(Call<List<ChatMember>> call, Throwable t) {
-                        onError();
-                    }
-                });
+                delayHandler.postDelayed(suggestionRunnable, DELAY);
```
kordianbruck: are you sure, this is not executing multiple requests, just delayed? 

Maybe we could use Rxbindings here: https://stackoverflow.com/questions/12142021/how-can-i-do-something-0-5-second-after-text-changed-in-my-edittext/46254925#46254925

kordianbruck: Ah, nevermind, you are removing the old callback above.

pfent:
@kordianbruck you looked over this one, would you also approve it?

kordianbruck:
@pfent I quickly scrolled through it. Didn't look at it in detail.

app/build.gradle
```diff
@@ -129,6 +129,8 @@ dependencies {
     implementation 'se.emilsjolander:stickylistheaders:2.7.0'
     implementation 'com.github.chrisbanes:PhotoView:2.1.3'
     implementation 'me.dm7.barcodescanner:zxing:1.9.8'
+    implementation 'com.google.zxing:core:3.3.0'
+    implementation 'com.journeyapps:zxing-android-embedded:3.2.0@aar'
```
pfent: any particular reason to not use the latest version?
I think we can use [the latest version](https://github.com/journeyapps/zxing-android-embedded/releases): 3.6.0, which should transitively pull in `zxing:core`, so we don't need to include it here

app/src/main/java/de/tum/in/tumcampusapp/api/app/Helper.java
```diff
@@ -109,6 +116,23 @@ public static String encodeUrl(String pUrl) {
...
+     * @return QR-Code or null if there was an error
+     */
+    public static Bitmap createQRCode(String message){
+        Writer multiFormatWriter = new MultiFormatWriter();
+        try {
+            BitMatrix bitMatrix = multiFormatWriter.encode(message, BarcodeFormat.QR_CODE, 400,400);
+            BarcodeEncoder barcodeEncoder = new BarcodeEncoder();
+            return barcodeEncoder.createBitmap(bitMatrix);
+        } catch (WriterException e) {
+            e.printStackTrace();
```
pfent: This should be a `Utils.log`

app/src/main/java/de/tum/in/tumcampusapp/component/ui/chat/AddChatMemberActivity.java
```diff
@@ -0,0 +1,229 @@
...
+            showConfirmDialog(member);
+        });
+
+        searchView.setOnEditorActionListener((textView, actionId, keyEvent) -> {
+            if(actionId == EditorInfo.IME_ACTION_SEARCH){
+                Utils.log("Search");
+                delayHandler.removeCallbacks(suggestionRunnable);
+                getSuggestions();
+
+                //searchView.clearFocus();
```
pfent: commented code shouldn't stay in there

pfent:
@kordianbruck merge plox

PR #837, opened 2018-05-24
Add animations to GradesActivity

jacqueline8711:
## Issue
everything in the Activity happens very abruptly, so animations will be nice

better UX


thellmund:
The fade transition between the two charts is fine

thellmund:
The fade transition between the two charts looks good to me. 

The animation for expanding/collapsing the charts needs improvement: 
1. Because the list view items don’t have a background of their own, the chart is visible underneath them during the transition. Probably best to add a dedicated list item background (1x overdraw is still reasonable).
2. Optional: The animation on the arrow looks a bit off. It appears as if the icon flips over (3D-ish). In Material Design land, I think it’s preferred if during animations, the elements stay on the same 2D plane. In our case, this would mean that the icon shape changes from down arrow to horizontal line to up arrow rather than flip over. (I can’t find any GIF to clarify this. Let me know if this comment doesn’t make sense.)

jacqueline8711:
@thellmund the fact that the chart is visible during the transition doesn't have anything to do with there not being a background. I tried adding one, but it's the same. This is the standard animation Android defines for the LinearLayout (I didn't write it myself). The only thing I can do is make the chart disappear instantly (without any fading effect whatsoever) and just let the rest of the layout slide up as before. But I'm not sure that's better, so I haven't changed it.

I did change the arrow cause that was actually a fun thing to do. So this is what it would look like in bigger and a little slower:
old version:
animation_old
new version:
animation_new

app/src/main/res/values/strings.xml
```diff
@@ -565,4 +565,10 @@ Signature: %5$s</string>
     <string name="e_mail">E-mail</string>
     <string name="additional_info">Additional Info</string>
     <string name="Refresh">Refresh</string>
+
+    <!-- pathData values for arrow animation -->
+    <string name="arrow_up_path" translatable="false">M7.41,15.41 l4.59,-4.58 l4.59,4.58 L18,14 l-6,-6 -6,6z</string>
+    <string name="arrow_down_path" translatable="false">M7.41,7.84 l4.59,4.58 l4.59,-4.58 L18,9.25 l-6,6 -6,-6z</string>
+    <string name="arrow_down_flat_path" translatable="false">M6.705,10.448 l5.295,0 l5.295,0 L17.295,12.442 l-5.295,0 -5.295,0z</string>
+    <string name="arrow_up_flat_path" translatable="false">M6.705,12.442 l5.295,0 l5.295,0 L17.295,10.448 l-5.295,0 -5.295,0z</string>
```
pfent: I don't think those values fit particularly well into `strings`
Maybe add another values resource `paths`?

pfent:
I thought those collapse / expand arrows did some kind or rotation as animation, but that doesn't seem to be the case in other apps :man_shrugging:
Your arrow animation looks fine :+1: 

And to properly animate transitions, you need to add a background and **give the replacing views an elevation.**
Adding `android:elevation="3dp"` to the `SwipeRefreshLayout` should fixes the wonky animation.

Also the pull to refresh animation is questionable :rofl: 
device-2018-05-26-134411

thellmund:
@jacqueline8711 Interesting. Then give @pfent’s solution a try. If that also does not work, maybe TransitionManager.beginDelayedTransition can be the last resort

PR #848, opened 2018-06-01
Implemented server calls getTickets() and getEvents() on client side

saileru:
I started working on the client side server calls. I implemented the methods getTickets() and getEvents(). I have verified them with a local server using the static API endpoints.

I copied the data classes Event, Ticket and TicketType from the Eventsoverview branch to be able to test my code.

Some open questions for me are:

1. Should we use a controller like in NewsController or a ViewModel as e.g. in KinoViewModel to manage the data?
2. I saw that for other components @Dao objects are used. As I understand it, the data is downloaded from the server and saved in a local database. I guess, we should also use a local database for events, right? 
Currently, I save the donwloaded data in attributes in EventsController. Can I use this approach and add the local database later?
3. I did not really understand how to pass parameters in server calls. There are multiple options like Path for including parameters in the url or Body for passing objects. The following call expects two parameters which are not part of the URL
GET api/event/ticket/validate

Parameters:

    event_id (required) -> event that the ticket is associated with
    code (required) -> ticket code

Response { "valid": true }

How should I pass these parameters?
4. Related to 3: On the backend the methods get a $post and a $params parameter. What is the difference there?

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/api/app/TUMCabeAPIService.java
+++ b/app/src/main/java/de/tum/in/tumcampusapp/api/app/TUMCabeAPIService.java
@@ -29,6 +29,8 @@ import de.tum.in.tumcampusapp.component.ui.news.model.News;
 import de.tum.in.tumcampusapp.component.ui.news.model.NewsAlert;
 import de.tum.in.tumcampusapp.component.ui.news.model.NewsSources;
 import de.tum.in.tumcampusapp.component.ui.studycard.model.StudyCard;
+import de.tum.in.tumcampusapp.component.ui.ticket.model.Event;
+import de.tum.in.tumcampusapp.component.ui.ticket.model.Ticket;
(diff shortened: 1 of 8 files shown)


kordianbruck:
1. go the MVVM way and use a ViewModel
2. Yes you are right, anything that is not locally cached in the DB, cannot be accessed when the user is offline. So by saving it to the local DB we can always display relevant information. Yes, you can add the DB later, but probably its just easiest to do it in one sweep.
3. via the body and a corresponding model class
4. `$post` is the body vars, `$parameters` are values passed over the URL  and parsed via the defined regex.

schulzalexander:
@saileru GET api/event/ticket/validate should be a POST, I updated the definition in the wiki.

schulzalexander:
@saileru 
I added functions for the other endpoints to the TUMCabeClient and -ApiService classes. In order to integrate and properly test them, we need to build those ViewModel classes first, which rely on the yet to be implemented local Database classes ;)

Issue #812, opened 2018-05-13
Ticket Sale: Event Overview


schulzalexander:
Implement a seperate activity showing a list of upcoming events, sorted by date.
The Activity should be accesible by the main menu.

Include a button to filter for events that the user bought tickets for.

For reference, see illustration at #809 .

kordianbruck:
I would suggest a card view similar to the home & news screen. Events should be fetched from our own API (TBD)

schulzalexander:
Done in #868

PR #853, opened 2018-06-02
Created EventDetailsFragment 

saileru:
I adjusted the KinoDetailsFragment to be able to display our new Event data type.
As it is not yet decided how we want to combine Ticket Sale and Kino classes, I copied some Kino classes to the new ticketsale package and renamed them. This should later be refactored!

The EventsController currently provides static Event objects to test the new fragments. I also added an entry to the navigation drawer to test the code. This should also be removed. The EventDetailsActivity should be accessed via the EventsActivity which will be implemented in #812.

I guess, you do not need to review every line since much code is just copied from the Kino classes. But I am happy to integrate your suggestions at this point if you have any ideas :)


Kailiangdong:
@saileru It seems that my recommanded code can't appear in review . Maybe you can take a reference of "android:name=".component.ui.tufilm.show_ticket" in AndroidManifest  to show title of your activity in your layout.

saileru:
@Kailiangdong Thanks for your comment! I will have a look at it.
The xml format seems to clash with the comment markup language. Just put a backslash "\" before the code. Then, it is shown as expected :)

app/src/main/java/de/tum/in/tumcampusapp/component/other/generic/drawer/DrawerMenuHelper.kt
```diff
@@ -94,7 +95,8 @@ class DrawerMenuHelper(private val mContext: Context, private val mDrawerLayout:
                 SideNavigationItem(R.string.news, R.drawable.ic_rss, NewsActivity::class.java, false, false),
                 SideNavigationItem(R.string.barrier_free, R.drawable.ic_accessible, BarrierFreeInfoActivity::class.java, false, false),
                 SideNavigationItem(R.string.opening_hours, R.drawable.ic_time, OpeningHoursListActivity::class.java, false, false),
-                SideNavigationItem(R.string.study_plans, R.drawable.ic_study_plans, CurriculaActivity::class.java, false, false)
+                SideNavigationItem(R.string.study_plans, R.drawable.ic_study_plans, CurriculaActivity::class.java, false, false),
+                SideNavigationItem(R.string.events, R.drawable.ic_time, EventDetailsActivity::class.java, false, false)
```
kordianbruck: Why do we want this in the menu? We actually only want the list of events in the menu, not the details.

saileru: Yes, correct! I only added this temporarily to test my code. This should later be removed and the Event Details will be linked from the Event list which is implemented in #812.

saileru:
@velev8 @Kailiangdong @schulzalexander I have connected the EventsOverview with the EventDetails now. The solution is not that pretty but it seems to work. If you have time, please have a look at my code and see if you have any suggestions. I need to go to sleep now ^^

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/EventDetailsAdapter.java
```diff
@@ -0,0 +1,57 @@
...
+    public int getCount() {
+        return count;
+    }
+
+    @Override
+    public CharSequence getPageTitle(int position) {
+        // returns the titles for the pagerStrip
+        String title = titles.get(position);
+        return title.substring(title.indexOf(':') + 1)
+                .trim();
```
schulzalexander: What's the purpose of taking a substring here?

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/EventDetailsActivity.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
+        // set up ViewPager and adapter
+        ViewPager mPager = findViewById(R.id.pager);
+
+        // TODO: extract key to const class
+        int clickedEventId = getIntent().getIntExtra("event_id", 0);
```
schulzalexander: The event ID you get here is wrong if in the Events List only booked events are shown. E.g. clicking on Unity3 will result in getting the ID for Unity2.
I guess that this problem has to be solved in the Event List or Event Card.

Kailiangdong: @saileru @schulzalexander I just solved this problem last night.  And I commited it . The problem results from that the event_id is not corresponding to its name.(for example id = 0 with "Unity 2" in getbookedevents()

Kailiangdong:
I think there are't much problem left and it is already pretty enough. Thank you. So I approved. May be we can make the card of events seem more separately,just like card of news. But we can make that together on Tuesday.

schulzalexander:
@Kailiangdong Changing the IDs and titles of our mocked data doesn't fix the problem, it will only make it work with that exact data.

PR #856, opened 2018-06-03
Update design of Settings

thellmund:
## Issue

1. The current design features oversized list items and text.
2. Some preferences use the wrong kind of selection controls, e.g. checkbox instead of switch.
3. Some preferences don’t offer subtitles, which could be used to display the currently selected option.

## Screenshot

Old: 
screenshot_20180603-144958

New: 
screenshot_20180603-150434


thellmund:
@pfent It was intentional, as I perceive them to be actual multi-select options.

PR #858, opened 2018-06-04
Fix text color of items in study rooms spinner

thellmund:
In a previous commit, I broke the spinner in `StudyRoomActivity` so that it would display light text on a light background. This is now fixed. 

I also used this opportunity to add a ripple effect to the spinner, giving the user visual feedback on item selection.


kordianbruck:
@thellmund could you please prefix your branch names with `th/` - thanks!

PR #861, opened 2018-06-04
Fix text in tuition fees card

thellmund:
## Issue

This fixes wrong formatting of the outstanding balance in `TuitionFeesCard`. It also unifies the formatting across `TuitionFeesCard`and `TuitionFeesActivity`.

## Screenshot

Before
screenshot_20180604-181648

After
screenshot_20180604-184851
(It uses a comma, because my device is set to English.)

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/tumui/tutionfees/model/Tuition.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/tumui/tutionfees/model/Tuition.kt
@@ -36,7 +36,7 @@ data class Tuition(@field:Element(name = "frist")
         get() {
             return try {
                 val amountText = String.format(Locale.getDefault(), "%.2f", outstandingBalance)
-                return "$amountText€"
+                return "$amountText €"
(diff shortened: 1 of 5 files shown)


app/src/main/java/de/tum/in/tumcampusapp/component/tumui/tutionfees/model/Tuition.kt
```diff
@@ -16,4 +19,27 @@ data class Tuition(@field:Element(name = "frist")
...
+            } catch (e: ParseException) {
+                0f
+            }
+        }
+
+    val outstandingBalanceText: String
+        get() {
+            return try {
+                val amountText = String.format(Locale.getDefault(), "%.2f", outstandingBalance)
+                return "$amountText€"
```
pfent: Aren't we missing a space here?

thellmund: You’re right! Added it back in.

Issue #841, opened 2018-05-26
Curricula unreadable / look horrible


jacqueline8711:
the study plans are pretty much html sites without css...
this is how it looks after you zoom out:
![screen_curricula](https://user-images.githubusercontent.com/16260112/40577240-afe8ce1e-6102-11e8-8e59-5a1cea13ae6d.png)
![screen_curricula_2](https://user-images.githubusercontent.com/16260112/40577241-b0f82192-6102-11e8-92e2-755d3966ad92.png)

pfent:
I think we can get rid of the study plans altogether.
We never supported curriculas outside of computer science, plus viewing them always was uncomfortable anyways...

I'd prefer to remove them, until we have a proper API for that information...
@xsrf Any chance for "Curricula Support" / "Studienplan" to be exposed by an API?
In theory, that should be public information (aka the same as on the website) - except for your own progress in your curriculum?

kordianbruck:
Anyone interested: I will try to setup a meeting for Friday the 8th - maybe we can organize a new API for this matter finally

kordianbruck:
So we removed them for now. Maybe in the future we will get an API, but this is still uncertain for now.

PR #866, opened 2018-06-07
remove Curricula, since since we're overhauling them

pfent:
Basically just removing the broken status quo, see #841.

I think in the future, we need to overhaul it properly. Most likely we can't reuse any of the current code, so I decided just to remove it all.


app/src/main/java/de/tum/in/tumcampusapp/component/other/generic/drawer/DrawerMenuHelper.kt
```diff
@@ -93,8 +92,7 @@ class DrawerMenuHelper(private val mContext: Context, private val mDrawerLayout:
                 SideNavigationItem(R.string.person_search, R.drawable.ic_users, PersonSearchActivity::class.java, true, false),
                 SideNavigationItem(R.string.news, R.drawable.ic_rss, NewsActivity::class.java, false, false),
                 SideNavigationItem(R.string.barrier_free, R.drawable.ic_accessible, BarrierFreeInfoActivity::class.java, false, false),
-                SideNavigationItem(R.string.opening_hours, R.drawable.ic_time, OpeningHoursListActivity::class.java, false, false),
-                SideNavigationItem(R.string.study_plans, R.drawable.ic_study_plans, CurriculaActivity::class.java, false, false)
```
jacqueline8711: maybe remove 'study_plans' string from strings.xml for now?

kordianbruck:
I'm all for removing the current status, once we have a go, that we have a different way of doing it. Right now we only scheduled a meeting for monday, that is not clear, what the result of it will be.

pfent:
@kordianbruck since we met and we have the potential for for a proper API, could we fix the utterly broken status quo?

kordianbruck:
@pfent I don't even know if we can get the API working. As you heard, the prospect of getting it is really slim.

thellmund:
@pfent @kordianbruck What’s the status here?

PR #867, opened 2018-06-07
disable crashlytics for debug and -dev builds

### LINKED ISSUE #824: Disable Crashlytics for Debug & log exceptions to cloud
https://github.com/codepath/android_guides/wiki/Crash-Reporting-with-Crashlytics

pfent:
this fixes #824 

@kordianbruck curiously, I'm still getting messages in logcat:
```
de.tum.in.tumcampus I/CrashlyticsCore: Crashlytics report upload complete: 5B199C5701A0-0001-3C87-B9ADEF9F5A41
```

I followed [the official guide](https://docs.fabric.io/android/crashlytics/build-tools.html), and verified the Crashlytics Kit `isDisabled == true`.
Maybe we need to check, if the debug messages still are shown in the dashboard...

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -23,9 +23,11 @@ android {
             minifyEnabled true
             useProguard true
         }
-        debug.debuggable true
-        debug.testCoverageEnabled true
         release.debuggable false
+        debug {
+            debuggable true
+            testCoverageEnabled true
+        }
     }
     defaultConfig {
         applicationId "de.tum.in.tumcampus"
@@ -67,7 +69,6 @@ android {
         androidTest.assets.srcDirs += files("$projectDir/src/test/java".toString())
         androidTest.assets.srcDirs += files("$projectDir/schemas".toString())
     }
-
     dataBinding {
         enabled = true
     }
@@ -87,6 +88,7 @@ configurations.all {
 }
 
 dependencies {
+
     // Support Libraries
     implementation "com.android.support:cardview-v7:$androidSupportVersion"
     implementation "com.android.support:design:$androidSupportVersion"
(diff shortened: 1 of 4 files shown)


app/build.gradle
```diff
@@ -103,7 +103,9 @@ dependencies {
     implementation "com.google.android.gms:play-services-base:$firebaseVersion"
     implementation "com.google.firebase:firebase-messaging:$firebaseVersion"
     implementation "com.google.android.gms:play-services-location:$firebaseVersion"
-    implementation 'com.crashlytics.sdk.android:crashlytics:2.9.2'
+    implementation('com.crashlytics.sdk.android:crashlytics:2.9.3@aar') {
+        transitive = true
```
kordianbruck: Do we really need this? We don't have library projects?

Instead we should use this:
```
android {
    buildTypes {
        debug {
          // Disable fabric build ID generation for debug builds
          ext.enableCrashlytics = false
```

pfent: Go ahead and try it.
`ext.enableCrashlytics = false` immediately crashes the app, before any of our code can run...

kordianbruck:
@pfent yea, its really weird. What do you think about this solution? Better? Worse? 

Seems to work for me. Previously it would still upload the reports somehow, so no clue, why it didn't work to disable it at runtime.

pfent:
@kordianbruck only disabling crash reporting for debug builds only solves the problem partially, because IIRC hitting the "play" button (:play_or_pause_button:) in Android Studio still deploys a release build.

That's why I tried to disable it for "-dev" builds at runtime, which doesn't seem to work...
Maybe need to add another "publish" build type, which only enables crashlytics for builds intended for the play store

kordianbruck:
@pfent I actually tested it with having the build variant "release" selected. It would only then be reporting crashes. In normal debug mode, it would not be compiled into the binary.

pfent:
cool, merging then

Issue #852, opened 2018-06-02
Improve user experience in Settings


thellmund:
**Situation today**
We offer in-app settings that the user can customize. 

**Problem today**
There are some areas of improvement: 
1. The current design features oversized list items and text. 
2. Some preferences use the wrong kind of selection controls, e.g. checkbox instead of switch.
3. Some preferences don’t offer subtitles, which could be used to display the currently selected option. 
4. There is an odd animation effect happening when opening the _News_ settings, as some preferences are inserted dynamically. 

**Suggestion**
I suggest the following improvements: 

- [x] Switch to a `android.support.v7.preference.PreferenceScreen` instead of `android.preference.PreferenceScreen` for proper Material Design. 
- [x] Use switch preferences instead of checkbox preferences for preferences that are simple on/off preferences rather than multi-select. 
- [x] Add proper subtitles. For instance, the default station for each campus should show the currently selected option as a subtitle. 
- [x] Fix the odd animation in _News_ settings.

thellmund:
Working on this in #856.

pfent:
@thellmund I think only the odd animation is left?

thellmund:
@pfent Yes, I did not find a way to fix that

PR #869, opened 2018-06-08
Fix odd animation in news settings

jacqueline8711:
## Issue
This fixes the following issue(s): #852


thellmund:
@jacqueline8711 Thanks for fixing this!

Issue #872, opened 2018-06-09
Drawer top cut off in MainActivity


jacqueline8711:
## Problem
I you are not logged in the image is slightly cut off at the top of the drawer when in MainActivity, it is fine in other activities

## Screenshot
![screenshot_top_cut_off](https://user-images.githubusercontent.com/16260112/41189456-85a6f262-6bce-11e8-987f-1e222dc33c11.png)

### Environment
* Phone: Pixel XL
* OS version: 8.1
* TUM Campus App version: 1.6.1 dev

PR #876, opened 2018-06-09
fix cut off image placeholder in drawer head

jacqueline8711:
## Issue

This fixes the following issue(s): #872 

## Screenshot
screenshot_not_cut_off


kordianbruck:
The whole text could be smaller - could we fix that also in this PR please?

Issue #878, opened 2018-06-09
Crash on first open


thellmund:
## Steps to reproduce
1. Clear all data of the TUM Campus app
2. Open the app

## Console output
06-09 21:52:49.814 10923-11196/de.tum.in.tumcampus E/AndroidRuntime: FATAL EXCEPTION: Thread-16
    Process: de.tum.in.tumcampus, PID: 10923
    android.database.sqlite.SQLiteException: index index_transport_favorites_symbol already exists (code 1 SQLITE_ERROR): , while compiling: CREATE UNIQUE INDEX `index_transport_favorites_symbol` ON `transport_favorites` (`symbol`)
        at android.database.sqlite.SQLiteConnection.nativePrepareStatement(Native Method)
        at android.database.sqlite.SQLiteConnection.acquirePreparedStatement(SQLiteConnection.java:901)
        at android.database.sqlite.SQLiteConnection.prepare(SQLiteConnection.java:512)
        at android.database.sqlite.SQLiteSession.prepare(SQLiteSession.java:588)
        at android.database.sqlite.SQLiteProgram.<init>(SQLiteProgram.java:58)
        at android.database.sqlite.SQLiteStatement.<init>(SQLiteStatement.java:31)
        at android.database.sqlite.SQLiteDatabase.executeSql(SQLiteDatabase.java:1769)
        at android.database.sqlite.SQLiteDatabase.execSQL(SQLiteDatabase.java:1698)
        at android.arch.persistence.db.framework.FrameworkSQLiteDatabase.execSQL(FrameworkSQLiteDatabase.java:240)
        at de.tum.in.tumcampusapp.database.TcaDb_Impl$1.createAllTables(TcaDb_Impl.java:135)
        at android.arch.persistence.room.RoomOpenHelper.onCreate(RoomOpenHelper.java:61)
        at android.arch.persistence.db.framework.FrameworkSQLiteOpenHelper$OpenHelper.onCreate(FrameworkSQLiteOpenHelper.java:113)
        at android.database.sqlite.SQLiteOpenHelper.getDatabaseLocked(SQLiteOpenHelper.java:393)
        at android.database.sqlite.SQLiteOpenHelper.getWritableDatabase(SQLiteOpenHelper.java:298)
        at android.arch.persistence.db.framework.FrameworkSQLiteOpenHelper$OpenHelper.getWritableSupportDatabase(FrameworkSQLiteOpenHelper.java:93)
[...]

PR #880, opened 2018-06-10
Fix DB index error on first launch

thellmund:
On the first launch of the app, we accidentally build the database twice. This results in a SQLite exception, as we attempt to create a unique index a second time. See #878 for console output.

This change fixes this problem by ensuring that we only build the database once.

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/database/TcaDb.java
+++ b/app/src/main/java/de/tum/in/tumcampusapp/database/TcaDb.java
@@ -106,7 +106,7 @@ public abstract class TcaDb extends RoomDatabase {
     private static TcaDb instance;
 
     public static synchronized TcaDb getInstance(Context context) {
-        if (instance == null || !instance.isOpen()) {
+        if (instance == null) {
             instance = Room.databaseBuilder(context.getApplicationContext(), TcaDb.class, Const.DATABASE_NAME)
                            .allowMainThreadQueries()
                            .addMigrations(migrations)
--- a/app/src/test/java/de/tum/in/tumcampusapp/activities/KinoActivityTest.kt
+++ b/app/src/test/java/de/tum/in/tumcampusapp/activities/KinoActivityTest.kt
@@ -20,6 +20,7 @@ import io.reactivex.schedulers.Schedulers
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.After
 import org.junit.Before
+import org.junit.Ignore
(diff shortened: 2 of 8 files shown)


app/src/main/java/de/tum/in/tumcampusapp/database/TcaDb.java
```diff
@@ -106,7 +106,7 @@
     private static TcaDb instance;
 
     public static synchronized TcaDb getInstance(Context context) {
-        if (instance == null || !instance.isOpen()) {
+        if (instance == null) {
```
pfent: The fix is not that easy...
The `|| !instance.isOpen()` was there for a reason, however I don't exactly recall why.
@psukys introduced it with #584, do you maybe recall why this was important?

pfent:
[Travis reports many tests as broken](https://travis-ci.org/TCA-Team/TumCampusApp/builds/390367445#L3335)

pfent:
Do two separate methods `getInstance()` and `create()` make sense here?
Then the tests could create new instances as they like, but on real devices we only have one instance...

thellmund:
There’s a larger problem with the test cases – they fail for all pull requests (despite being displayed as passed on GitHub).

kordianbruck:
@pfent @thellmund feel free to fix the tests. I could for the love of god not figure out why room would be throwing "Illegal connection pointer" during the tests. I've spent like 3 hours on this, so if you have ideas go right ahead. For the mean time I've ignored failing tests.

app/src/main/java/de/tum/in/tumcampusapp/database/TcaDb.java
kordianbruck: For the test cases. After each case, the DB is closed to force a flush of all cursors. This somehow does no longer work.

pfent:
@kordianbruck that Roboelectric testing framework is a mess, implementing their own SQLite engine, with *special* behaviour...
I'm not toughing that :see_no_evil:

kordianbruck:
@pfent so back to instrumentation tests, now that they are supposed to be faster?

pfent:
Last I checked, it was impossible to run instrumentation tests on travis or on our build server, due to the emulator needing massive resources... :man_shrugging: 
IMHO, we could spend time much more effectively actually working on features, that fighting testing frameworks

kordianbruck:
Y is testing android apps such a mess?

PR #881, opened 2018-06-10
Fix crash in TransportDetails

thellmund:
This fixes a crash that occurred when building the `MVVSymbolView`. Because the MVV API occasionally only returns "U" for a subway departure instead of the full symbol (e.g. "U3"), the app crashed while attempting to get the right subway line color for the departure. 

The fix simply provides a default subway line color for these cases.

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/ui/transportation/MVVSymbolView.java
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/ui/transportation/MVVSymbolView.java
@@ -1,6 +1,7 @@
 package de.tum.in.tumcampusapp.component.ui.transportation;
 
 import android.graphics.Canvas;
+import android.graphics.Color;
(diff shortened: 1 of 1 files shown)


app/src/main/java/de/tum/in/tumcampusapp/component/ui/transportation/MVVSymbolView.java
```diff
@@ -67,8 +73,10 @@ public MVVSymbolView(String line) {
                 } else if (num == 8) {
                     triangle = 2;
                     backgroundColor = 0xffeb6a27;
-                } else {
+                } else if (num != 0) {
                     backgroundColor = U_LINE_COLOR[num - 1];
```
pfent: :man_facepalming: 
we already had the `Optional.fromNullable(...).or(0)` in place, but then accessed out of bounds

PR #883, opened 2018-06-10
Prevent crash caused by WifiManager

thellmund:
When starting a scan with `WifiManager`, a `NullPointerException` occasionally occurred within `WifiManager`. See [Crashlytics](https://console.firebase.google.com/project/tca-backend-0001/crashlytics/app/android:de.tum.in.tumcampus/issues/5ac7317936c7b23527adce98?time=last-seven-days&sessionId=5B1CD063022300023A092389244C668C_DNE_0_v2) for the console output.

We temporarily fix this problem by wrapping these calls in a try-catch block. In the long term, we should investigate the reason for the exception.


thellmund:
Thanks for the thorough research, @pfent! I’ll take a look at the WiFiLock approach 👍

kordianbruck:
I don't think we are using the wifiscans plus its really inefficient to launch background tasks on connectivity change. We should probably just remove this.

kordianbruck:
@pfent please never, never, NEVERRRR approve and merge your own changes. Thats not how reviews work nor how the four eyes principle work. Yes, its annoying, to wait for somebody else, but that's a principle we all should try to adhere to. (Yes, I'm also guilty of doing it in the past and I'll try to also adhere to it) - cheers!

Apart from that: good riddance :clinking_glasses:

kordianbruck:
For example: how does the Eduroam Setup Card get displayed now?

pfent:
@kordianbruck just followed your suggestions :man_shrugging:, but you're right I probably shouldn't have removed the `WifiScanHandler`, but only the `NetworkChangeReceiver`

PR #885, opened 2018-06-10
Fix crashes when device is offline

thellmund:
Performing requests with Retrofit while the device is offline used to result in exceptions. This occurred when the `DownloadService`was invoked in the background. 

This change adds a `ConnectivityInterceptor`, which checks whether the device is connected to the Internet (via `NetUtils`). If not, it throws a `NoNetworkConnectivityException`, which can be reacted to in Retrofit’s `onFailure()` callback.

This prevented any crashes in my tests. Please give it a thorough review.

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/api/app/ConnectivityInterceptor.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/api/app/ConnectivityInterceptor.kt
@@ -1,0 +1,23 @@
+package de.tum.`in`.tumcampusapp.api.app
+
+import android.content.Context
+import de.tum.`in`.tumcampusapp.api.app.exception.NoNetworkConnectionException
+import de.tum.`in`.tumcampusapp.utils.NetUtils
+import okhttp3.Interceptor
+import okhttp3.Response
+import java.io.IOException
+
+class ConnectivityInterceptor(private val context: Context) : Interceptor {
+
+    @Throws(IOException::class)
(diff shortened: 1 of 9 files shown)


app/src/main/java/de/tum/in/tumcampusapp/api/app/ConnectivityInterceptor.kt
```diff
@@ -0,0 +1,23 @@
...
+import java.io.IOException
+
+class ConnectivityInterceptor(private val context: Context) : Interceptor {
+
+    @Throws(IOException::class)
+    override fun intercept(chain: Interceptor.Chain): Response {
+        val request = chain.request()
+
+        if (!NetUtils.isConnected(context)) {
+            throw NoNetworkConnectionException()
```
pfent: do we actually handle that exception somewhere?

thellmund: Not by us. When `NoNetworkConnectionException` is thrown, it’s caught and passed to Retrofit’s `onFailure()` callback and can be handled there, whereas previously it would result in an uncaught exception. 

Every caller of Retrofit can handle network exceptions themselves. Right now, we mostly only log those.

pfent: So why does this then fix the crashes, if we just throw other exceptions?
Are IOExceotions already handled by OkHttp?

PR #886, opened 2018-06-11
Update design of CreateEventActivity

thellmund:
## Issue

- Improves the design of the event creation screen
- Adds cancel dialog to prevent accidental loss of data
- Removes empty navigation drawer

## Screenshot
screenshot_20180611-092832

- More in line with the design of the rest of the app
- No data loss when accidentally tapping the back button


kordianbruck:
...about also adding a background in this activity or some what to offset the form from the background?

thellmund:
@kordianbruck Android Jetpack will introduce a new design for the EditText component which will help with this.

screen shot 2018-06-12 at 11 04 58

kordianbruck:
Yet another library? But it looks smooth.

thellmund:
Not “another” library, but rather a replacement library. Jetpack will basically unify support library, architecture components and a whole bunch of other things.

PR #887, opened 2018-06-11
Payment Selection and Checkout

schulzalexander:
This includes a payment "selection" (by now only credit card via stripe) which leads to the final checkout page, where the user can enter their credit card information.

This last page can eventually be replaced by an overview of creditcards that the user saved, but due to the short timeframe until the presentation on Wednesday, I just took the form to enter new card data here for now.


schulzalexander:
bildschirmfoto 2018-06-11 um 13 58 10
bildschirmfoto 2018-06-11 um 13 58 01

Kailiangdong:
NICE JOB

PR #897, opened 2018-06-12
Merge master into ticketsale

Kailiangdong:


saileru:
@kordianbruck Merging directly into the feature branch does not seem to work for us. We are expected to have two approving review. Are we supposed to create a pull request every time we want to merge master into our feature branch?

kordianbruck:
@saileru you are right. As I've setup branch protection for the feature branch, you cannot directly pull into the branch. If you want, I can reduce the number of reviews to one or completely remove the branch protection, if the team agrees to continue to do reviews for each other. Deal?

saileru:
I think the branch protection should probably stay as it gives some safety. It makes it more difficult to break something on the feature branch accidentially.
Another idea would be to merge master into our feature branches before merging them back to feature/ticketsale from time to time. That way we could combine feature ticket reviews and master merging.

kordianbruck:
@saileru that makes changes from master be part of the usual review. Not too nice to look at. For the time being, lets just do a weekly update from master to feature, after our scrum meeting okay?

saileru:
@kordianbruck Ok, sounds good!

PR #901, opened 2018-06-13
Rework WiFi scanning

pfent:
Follow up from #883

The current state wouldn't show the Eduroam card anymore, since we wouldn't do periodic scans anymore.
However this time, we start the `WifiScanHandler` exactly once from the `BackgroundService` instead of on each connectivity changed.
Additionally, aquire a `WifiLock` to hopefully properly fix #883


kordianbruck:
@pfent cheers :beers:

Issue #790, opened 2018-04-29
Upload obsfucated IDs


kordianbruck:
The TUMonline `id` endpoint now gives us the obsfucated IDs for the current logged in user. We can use those to display the profile pictures in Chat and other upcoming interaction features (Study Cards :+1: ) to make the whole experience more personalized.

We need to upload these IDs to our API backend (tumcabe), so that we can deliver them with each ChatMessage and generally give the possibility to get a profile pic for any user.

There are several ways to do this:
* Augment the current existing API for uploading the chat registration to also include those IDs
* Add a new API that will allow us to upload IDs once.
* ???

Please look at the existing implementation and let me know what makes sense. (Backward compatibility and ease of use)

jacqueline8711:
https://wiki.tum.de/display/tumonlineappdevndoc/id

jacqueline8711:
- [ ] background service that lets us check what IDs have been uploaded and upload missing ones if possible
- [ ] upload the obfuscated ids directly after logging into tum online after checking if they already exist
- [ ] rename remaining gcm names to fcm / firebase
- [ ] check if merging DeviceVerification and ChatVerification would be a good idea e.g. ChatVerification extends DeviceVerification

backend:
- [ ] alter table member (add three columns for the obfuscated ids)
- [ ] create api endpoint that allows user to upload the ids (only once, a second upload should cause error)
- [ ] api endpoint to check for all existing ids (student/employee/external obfuscated id, RSA key, device id, firebase id) by sending device id which is added to the server db automatically
- [ ] accept all rsa keys and verify them with TUMonline later on (and then set the flag pkactive), flag has to be set for all requests that use ChatVerification

thellmund:
Finished in #908.

kordianbruck:
Follow up issue, due to renaming fields in models by @thellmund #1059

PR #902, opened 2018-06-14
Check which ids/keys/tokens have been uploaded and upload missing ones

jacqueline8711:
## Issue
This fixes the following issue(s): #790


kordianbruck:
@jacqueline8711 if the renaming part is done, lets merge, so I don't have to look at oversized PRs :tada:

jacqueline8711:
@kordianbruck yes, we can do that

PR #908, opened 2018-06-15
Check which ids have been uploaded and upload missing ones, part 2

jacqueline8711:
## Issue
This fixes the following issue(s): #790


jacqueline8711:
@kordianbruck all things mentioned in the issue should work now, the only thing I wasn't sure about is what should happen if the server can't verify the pk with TUMonline

kordianbruck:
@jacqueline8711 sounds good. In case it can't verify, it should show an error. Can you please also merge latest master?

pfent:
@kordianbruck Unfortunately, I don't really know what you and Jacqueline were up to with this PR, but I merged the master, so we *could* merge it.

I don't see obvious problems with the code, but it would be great if you could test it and see if the logic makes sense.

kordianbruck:
@pfent @thellmund we will need to talk about this someday in the future. I'm not sure, the current implementation is backward compatible and I want to avoid any errors when we change the API in this manner. I would give it some more time, until Till and I had a chance to talk about it.

kordianbruck:
#1059 is also part of this PR

Issue #915, opened 2018-06-19
Ticket Sale: Send events to calendar app


saileru:
TS-Mi-2: As a student, I can add the event to my preferred calendar app, so that I can see it in context of my schedule.
Each event should have an “add to calendar“-button in its detail activity (regardless of whether a user has already bought tickets for an event or not). Additionally, we can add an icon for this action to the section of each event in the event overview (possibly a tableView), if there is still space for this. Once an event is added to the local calendar, the icons/buttons should change their appearance to signal that the calendar entry has been created. Clicking on the very same icons/buttons again should simply remove the event from the calendar.
Adding an event to the default calendar can be done by using the Calendar Provider, which offers a bunch of Calendar Intents. These allow to specify the action you want to perform on the calendar, in our case, add an event. When executed, this will take the user right to the calendar app itself, where they can still modify and finally confirm the event. Afterwards, the user will be taken back to where they were in our app. Using this approach, we don’t need the WRITE_CALENDAR user-permission in our manifest, as compared to the other possibility of just adding the event to the calendar without user interaction, which would require that permission.

Maybe something like this could also work: https://stackoverflow.com/questions/6617231/how-to-use-calendar-intent

pfent:
We already have some code to add calendars, maybe you want to have a look at `CalendarActivity`

velev8:
After the scrum meeting our team decided that we will create the mockup for this issue this weekend since we are still waiting for a little more progress!

Issue #896, opened 2018-06-12
Ticket Sale: Feedback Event Details


saileru:
Please give us feedback how to improve the ticket sale UI

Points from the Scrum Meeting are:

- "Buy" button as floating action button
- obvious/motivating color like green/red
- include price in the event details, maybe on the buy button
- link as separate button below the picture, so that the picture has more room 

![screenshot_20180612-142624](https://user-images.githubusercontent.com/12066533/41290695-eff7f52c-6e4d-11e8-9924-348140c0bc6a.png)
![screenshot_20180612-142637](https://user-images.githubusercontent.com/12066533/41290696-f095da94-6e4d-11e8-92e9-5a6cee7997f3.png)

kordianbruck:
* properly format the date according to locale
* link address to google maps / room finder
* Not sure we need the pager / spinner on top, to switch between events. The user flow should be that one selects one event from the list and looks at it. Otherwise, why would we have the list in the first place?

thellmund:
Aren’t event posters in landscape format? In that case, I think it makes sense to adapt the layout accordingly.

velev8:
I would also make the different tab buttons (tunix, winterball, garnix) stand out more as buttons

velev8:
To Do: 

- [ ] Format date according to locale
- [ ] Link Address to google maps (click listener)

pfent:
I really dislike the buttons for the link / buy ticket.
Please refer to [the material design guidelines](https://material.io/design/components/buttons.html#hierarchy-placement), especially for placement of buttons.

IMO, floating action buttons at the bottom of the screen are the way to go, but feel free to come up with ideas.

velev8:
New mockup for this issue, please give feedback 
<img width="271" alt="screen shot 2018-06-26 at 15 34 46" src="https://user-images.githubusercontent.com/38428686/41915622-70b14728-7956-11e8-8c44-cb6c12c6ccdf.png">

kordianbruck:
@velev8 Looks really good, love it. Some things:

1. I would put the website link as a button on the very bottom
2. Why should we split date & time? Do we really support date ranges?
3. Can we add the event to the calendar, once somebody clicks on the date? (make it a button?)

saileru:
@velev8 @kordianbruck We planned to have pictures in something like 16:9 landscape, right? This format won't work for the new UI layout.

kordianbruck:
[...]

PR #922, opened 2018-06-23
add Event to calendar

Kailiangdong:
## Issue
According to issue #915. As a minor feature in Ticket sales, user can add their interested event to local calendar.
As description in issue 915, I use the Calendar Provider, which offers a bunch of Calendar Intents. When user click on the button"Add in Calendar" it will take the user right to the calendar app itself, where they can still modify and finally confirm the event. In this case, WRITE_CALENDAR user-permission is not needed.
What's more, because we have address in our events detail, users can open  google map after they create events in calendar by just click on the address in Calendar. 
The UI of EventsDetailOverview will be improved in other issue #896. This issue is just for function.

## Screenshot
image
image
image


app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/EventDetailsFragment.java
```diff
@@ -111,18 +113,20 @@ private void createEventHeader(LinearLayout rootView) {
         // initialize all buttons
         Button link = headerView.findViewById(R.id.button_link);
         Button ticket = headerView.findViewById(R.id.button_ticket);
+        Button export_calendar = headerView.findViewById(R.id.button_export_eventcalendar);
```
saileru: Use the java naming convention for variables: exportCalendar

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/EventDetailsFragment.java
```diff
@@ -159,6 +163,18 @@ private void buyTicket() {
         startActivity(intent);
     }
 
+    private void addToCalendar() {
+        Intent intent = new Intent(Intent.ACTION_INSERT)
+                .setData(CalendarContract.Events.CONTENT_URI)
+                .putExtra(CalendarContract.EXTRA_EVENT_BEGIN_TIME, event.getDate().getTime())
+                .putExtra(CalendarContract.EXTRA_EVENT_END_TIME, event.getDate().getTime() + 7200000)
```
saileru: We currently do not have a duration for events.
Assuming two hours for all events should only be a temporary workaround.
Maybe add a TODO to adjust this when/if a model adjustment of the event class is possible.

@kordianbruck @schulzalexander Is it possible and does it make sense to provide a duration for events?

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/EventDetailsFragment.java
```diff
@@ -159,6 +163,18 @@ private void buyTicket() {
...
+                .putExtra(CalendarContract.Events.TITLE, event.getTitle())
+                .putExtra(CalendarContract.Events.DESCRIPTION, event.getDescription())
+                .putExtra(CalendarContract.Events.EVENT_LOCATION, event.getLocality())
+                .putExtra(CalendarContract.Events.AVAILABILITY, CalendarContract.Events.AVAILABILITY_BUSY);
```
saileru: What does this line do? 
This might be a good place to add a comment to explain what it does and why we need it as I  do not understand it from just reading the code.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/EventDetailsFragment.java
saileru: We talked about it shortly in the Hangouts meeting today. We agreed that it is not a critical issue at the moment. But if we think this might be a reasonable addition in the future, we could add a ticket now, so that we do not forget it.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/EventDetailsFragment.java
Kailiangdong: fixed it . Thx

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/EventDetailsFragment.java
Kailiangdong: Nice. I keep it now

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/EventDetailsFragment.java
Kailiangdong: I change it to CalendarContract.Events.AVAILABILITY_FREE. It indicates that this event is free time and will not conflict with other events. I make a comment here.

Issue #898, opened 2018-06-12
Ticket Sale: Feedback Show Ticket


saileru:
Please give us feedback how we can improve the ticket sale UI

![screenshot_20180612-142629](https://user-images.githubusercontent.com/12066533/41292264-5ca79af2-6e52-11e8-8978-e0d5fcdf6138.png)

kordianbruck:
* Titel of event could be bigger
* link address to google maps
* Properly format date according to currently set locale
* "Ticket QR Code"
* This activity should set the screen brightness to 100%, so scanning is easier
* Ticket Redepmtion status should be shown?
* Ticket Price should be shown

thellmund:
I don’t think we need the headers. If we layout the content appropriately (e.g. make the title bigger), it should be easy enough to grasp. And particularly for the QR code, the header is unnecessary. 
- [ ] Remove the headers

The background of the activity should be the same as the background of the QR code image. 
- [ ] Use the same background for background and QR code

I support @kordianbruck’s requests. The link to Google Maps though should however not be a text link. Instead, set a click listener on the text view (or is parent, if appropriate).

velev8:
- [ ] Bigger title of event
- [ ] Remove the headers
- [ ] Use the same background for background and QR Code 
- [ ] Link address to google maps (click listener on the text view or its parent)
- [ ] Format Date to current locale
- [ ] "Ticket QR Code" (add space, capitalze C)
- [ ] Set screen brightness to 100% 
- [ ] Show ticket redemption status
- [ ] Show ticket price

PR #923, opened 2018-06-24
Improve show ticket

Kailiangdong:
## Issue

Improve nearly all problem from issue #898 . Except showing ticket redemption status, not quite understand this. Have discussed with team-member.  Adding this means we should change our data model?

Except this, users can click the address then jump to google map, the text will then become red after clicking. 
Other change is showing on the screenshot.
## Screenshot

image


saileru:
screenshot_20180625-153645

saileru:
I made a few more adjustments: 
format price to e.g. "4,50 €"
added attribute for redemption state to the Ticket class

@Kailiangdong @schulzalexander @velev8 @kordianbruck What are your thoughts about the redemption state attribute in the Ticket class? I think it might be a good idea.

I am also open to suggestions about how to better display the redemption status in the UI.

Kailiangdong:
@saileru  I will create a issue about that

kordianbruck:
@saileru @Kailiangdong in the backend it is a datetime field - no? Are we not exporting this over the API?

PR #924, opened 2018-06-25
Implemented local database for events

saileru:
I implemented the local database using the Model-View-Controller pattern. I took the News implementation as a reference. I removed most of the static data which we previously used for testing purposes and replaced it by actual data which is fetched from the server. I still had to add some dummy data for testing purposes as there is not much actual data available on the server right now. I marked the places in the code with TODOs.

I understand that MVVM would be the preferred way.
I tried this approach using the Kino classes, (KinoViewModel, KinoLocalRepository, KinoRemoteRepository etc.) as a reference. However, I was not able to completely understand what is going on there as most of it is Kotlin and my Kotlin skills are ... still to improve ^^


app/schemas/de.tum.in.tumcampusapp.database.TcaDb/10.json
```diff
@@ -2,7 +2,7 @@
   "formatVersion": 1,
   "database": {
     "version": 10,
```
kordianbruck: When you add tables, you have to increase the database version!

kordianbruck: Yes, otherwise it will crash for anyone trying it out. Yes, you should add migrations and increase the version number for each change in the DB.

saileru: Version increment done in #944

Issue #929, opened 2018-06-26
Ticket Sale: Create event data on server for demonstration


saileru:
At our presentation for the Android practical course next Tuesday (July 3rd), we want to show a live demonstration of the ticket sale feature. As the server calls are now implemented on client side, we want to use event data from the server.

For that purpose event data should be added on the server.
E.g.
6 Events
3 Tickets for dummy user with id 1

PR #931, opened 2018-06-26
Fixed the crash in ShowTicketActivity after merge

saileru:
I simply adjusted the dummy value to an existing one for now to prevent the crash. After the event data on the server is adjusted in #929 we can remove some of the dummy values on the client.


saileru:
AS this is a very small pull request, I used the opportunity to also merge master. The only conflicts were in the strings.xml files.

kordianbruck:
@Kailiangdong please the delete the branch after merging, if it is no longer needed :+1:

PR #932, opened 2018-06-28
Bump database version

thellmund:
The crashes in 1.6.0-beta1 reveal that we forgot to bump the database version. This commit fixes that.

### DIFF
--- a/app/schemas/de.tum.in.tumcampusapp.database.TcaDb/11.json
+++ b/app/schemas/de.tum.in.tumcampusapp.database.TcaDb/11.json
@@ -1,0 +1,1102 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 11,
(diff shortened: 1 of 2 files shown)


thellmund:
@kordianbruck @jacqueline8711 @pfent How do we prevent this in the future?

kordianbruck:
@thellmund we did not change any of the models or did we?

kordianbruck:
@thellmund for the error I've emailed you before, we've last changed that file in February and still it would try to create some index on it - I really don't know why.

See https://github.com/TCA-Team/TumCampusApp/blob/ad28cb21894c58d899cd8186171fe09b78c7462a/app/src/main/java/de/tum/in/tumcampusapp/component/ui/transportation/model/TransportFavorites.kt

thellmund:
@kordianbruck Regarding the error you emailed me: I’ll have a look tonight, but it’s not related to the schema-related crashes that this commit addresses.

PR #943, opened 2018-06-30
As/integrate stripe creditcard overview

schulzalexander:
The credit card checkout is not finished yet, but in my opinion ready for the demo.

Things that are still left to do:
- find out how to retrieve user's ID (the one that is used as primary key in the database), which is needed to reserve a ticket (by now, one will always "purchase" the same ticket with a static ID)
- when the ticket is bought, we jump to the ShowTicketActivity, which expects the new ticket to be available from the TicketController. After adding the localDatabases, we still have to add a method there that adds the new ticket to the TicketController/database (by now, the activity is opened with one of the static tickets from the TicketController
- when the user presses the purchase button, the TUMCabeClient returns a static ticket, because the recently merged purchasing-backend-logic is not compatible with the float-value price of the current database content, we have to wait here until the new sample data has been added to the database to use the actual purchasing server endpoint.


schulzalexander:
duplicate PR

PR #955, opened 2018-07-04
fix crash by making WifiLock nullable

### LINKED ISSUE #953: Crash in WifiScanHandler
Firebase reports the following crash for 1.6.1-beta3: In the `WifiScanHandler`, an `UninitializedPropertyAccessException` occurs (lateinit property wifiLock has not been initialized). 

Fix this before the final release of 1.6.1.

pfent:
this fixes #953

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/service/WifiScanHandler.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/service/WifiScanHandler.kt
@@ -17,9 +17,10 @@ import java.util.concurrent.atomic.AtomicBoolean
 class WifiScanHandler : Handler() {
     private var isRunning = AtomicBoolean(false)
     private lateinit var wifiManager: WifiManager
-    private lateinit var wifiLock: WifiManager.WifiLock
-    private val periodicalScan = {
-        wifiLock.acquire()
+    private var wifiLock: WifiManager.WifiLock? = null
+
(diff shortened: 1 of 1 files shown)


app/src/main/java/de/tum/in/tumcampusapp/service/WifiScanHandler.kt
```diff
@@ -35,8 +36,9 @@ class WifiScanHandler : Handler() {
     }
 
     fun onScanFinished() {
-        if (wifiLock.isHeld) {
-            wifiLock.release()
+        wifiLock?.apply {
+            if (isHeld)
```
thellmund: Can you use curly brackets here?

app/src/main/java/de/tum/in/tumcampusapp/service/WifiScanHandler.kt
```diff
@@ -35,8 +36,9 @@ class WifiScanHandler : Handler() {
     }
 
     fun onScanFinished() {
-        if (wifiLock.isHeld) {
-            wifiLock.release()
+        wifiLock?.apply {
```
thellmund: Please use `let` instead of `apply`, as this is the general practice for null checks. 

`apply` is typically used for scenarios where you will return the variable, e.g. `val x = SomeObject().apply { doSomethingOnSomeObject() }`

PR #956, opened 2018-07-04
Add network resiliency testing interceptor and improve error handling

pfent:
We still have some `OnErrorNotImplementedExceptions`, which only manifest in on-device tests with shaky network conditions.

In this PR I tried to implement a way to test this in debug builds and reliably reproduce shaky network conditions.
For this I was inspired by Netflix's [ChaosMonkey project](http://principlesofchaos.org/), thus the naming of the network Interceptor.

Additionally, I already added some error handling for TUMonline requests, but I still get random crashes, when deleting the cache...

Unfortunately, the logs are relatively cluttered with the network errors...
We can maybe tune down the logging of network error messages a bit, to get a better look on the actual crashes.

If you want to reliably crash the app, set `ChaosMonkeyInterceptor.failProbability = 1` :tada:


thellmund:
@pfent Great effort! Let’s discuss it (and our general approach) some more on Friday – I’ll have a more detailed look at this pull request before then.

PR #958, opened 2018-07-05
Fix NoSuchMethodError in Exam

thellmund:
This commit fixes a `NoSuchMethodError` that occurs on devices with API levels lower than 24 due to the usage of `Map.getOrDefault()`.

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/tumui/grades/model/Exam.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/tumui/grades/model/Exam.kt
@@ -44,12 +44,17 @@ data class Exam(@field:Element(name = "lv_titel")
                 headerId.compareTo(other.headerId) * (-1)
             }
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/de/tum/in/tumcampusapp/component/tumui/grades/model/Exam.kt
```diff
@@ -44,12 +44,15 @@ data class Exam(@field:Element(name = "lv_titel")
                 headerId.compareTo(other.headerId) * (-1)
             }
 
-    fun getGradeColor(c: Context) = GRADE_COLOR.getOrDefault(grade, R.color.grade_default).let {
-        ContextCompat.getColor(c, it)
+    fun getGradeColor(context: Context): Int {
+        // While using getOrDefault() compiles, it results in a NoSuchMethodError on devices with
+        // API levels lower than 24.
+        val resId = GRADE_COLORS[grade] ?: R.color.grade_default
```
pfent: Props for the explaining comment!

PR #969, opened 2018-07-07
fix event details layout and date display

saileru:
@kordianbruck If I remember correctly, in the last meeting we discussed that the date and tickets text views in the event details should be added to a horizontal scroll view to avoid a line break.

Futhermore, I fixed the date display. Before, 14:00 or 2 pm was displayed as 02:00.
I found that the easiest way to specify a date format is using getDateTimeInstance(int dateStyle, int timeStyle);
It is Locale sensible and offers some predefined styles like DateFormat.DEFAULT or DateFormat.Short

Here are two screenshots for German and US Locale

screenshot_20180707-132815
screenshot_1530962689


Kailiangdong:
image
It seems the German version on my emulator is not as show as your screenshot

saileru:
@Kailiangdong This post describes the issue: https://stackoverflow.com/questions/48636884/java-dateformat-short-in-android-not-working-as-expected

PR #976, opened 2018-07-09
As/adapt events add endtime

schulzalexander:
We added a endtime to the event class, and changed naming of "date" to "start".

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/EventDetailsFragment.java
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/EventDetailsFragment.java
@@ -22,12 +22,20 @@ import android.widget.TextView;
 import com.squareup.picasso.Callback;
 import com.squareup.picasso.Picasso;
 
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
 import de.tum.in.tumcampusapp.R;
+import de.tum.in.tumcampusapp.api.app.TUMCabeClient;
 import de.tum.in.tumcampusapp.component.tumui.calendar.CreateEventActivity;
 import de.tum.in.tumcampusapp.component.ui.ticket.model.Event;
+import de.tum.in.tumcampusapp.component.ui.ticket.payload.TicketStatus;
 import de.tum.in.tumcampusapp.utils.Const;
 import de.tum.in.tumcampusapp.utils.DateTimeUtils;
 import io.reactivex.disposables.CompositeDisposable;
+import retrofit2.Call;
+import retrofit2.Response;
 
 /**
  * Fragment for EventDetails. Manages content that gets shown on the pagerView
@@ -135,10 +143,7 @@ public class EventDetailsFragment extends Fragment {
         eventLocationTextView.setOnClickListener(view -> showMap());
 
         //set remaining tickets,following code is just for testing purpose.
-        //TODO:The remaining tickets should get from backend.Like event.getremainningtickets()
-        int remainingTickets = 30;
-        String remainingTicketsString = remainingTickets + " Tickets left!";
-        eventRemainingTicketTextView.setText(remainingTicketsString);
+        setAvailableTicketCount(eventRemainingTicketTextView);
 
         String eventDescriptionString = event.getDescription();
         eventDescriptionTextView.setText(eventDescriptionString);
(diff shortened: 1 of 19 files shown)


app/src/main/java/de/tum/in/tumcampusapp/database/Converters.kt
```diff
@@ -9,10 +9,14 @@ import org.joda.time.DateTime
...
-    fun isoToDateTime(str: String): DateTime = DateTimeUtils.getDateTime(str)
+    fun isoToDateTime(str: String): DateTime? {
+        return if (str.isNotEmpty()) DateTimeUtils.getDateTime(str) else null
+    }
 
     @TypeConverter
-    fun fromDateTime(date: DateTime): String = DateTimeUtils.getDateTimeString(date)
+    fun fromDateTime(date: DateTime?): String {
+        return if (date != null) DateTimeUtils.getDateTimeString(date) else ""
+    }
```
schulzalexander: @kordianbruck is it ok if we change these converter methods to also accept null? or should we solve this within our class.

kordianbruck: This should be more correct:

```
@TypeConverter
    public static Long toLong(Date value) {
        return value == null ? null : value.getTime();
    }
```

That way you will also have `null`s in the db.

app/src/main/java/de/tum/in/tumcampusapp/utils/Const.kt
```diff
@@ -131,6 +131,7 @@ object Const {
     const val EVENT_START = "pVon"
     const val EVENT_END = "pBis"
     const val EVENT_NR = "pTerminNr"
+    const val EVENT_DEFAULT_DURATION = 7200000 // Milliseconds
```
kordianbruck: Please don't do this.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/EventDetailsFragment.java
```diff
@@ -178,18 +173,20 @@ private void addToTUMCalendar() {
...
     private void addToExternalCalendar() {
         Intent intent = new Intent(Intent.ACTION_INSERT)
                 .setData(CalendarContract.Events.CONTENT_URI)
-                .putExtra(CalendarContract.EXTRA_EVENT_BEGIN_TIME, event.getDate().getMillis())
-                .putExtra(CalendarContract.EXTRA_EVENT_END_TIME, event.getDate().
-                        plus(eventDuration).getMillis())
+                .putExtra(CalendarContract.EXTRA_EVENT_BEGIN_TIME, event.getStart().getMillis())
+                .putExtra(CalendarContract.EXTRA_EVENT_END_TIME, event.getEnd() != null
+                        ? event.getEnd()
+                        : event.getStart().plus(Const.EVENT_DEFAULT_DURATION).getMillis())
```
kordianbruck: Just don't show the end date, in case it is not set.

Kailiangdong:
Bug of adding to external event:
GARNIX and TUM SOM work well with adding to external event. But the calendar will crash on WINTERBALL  and TUNIX

Kailiangdong:
when I click the "CONTINUE TO PAYMENT" the app will crash
**07-09 22:35:06.007 7479-7479/de.tum.in.tumcampus E/AndroidRuntime: FATAL EXCEPTION: main
    Process: de.tum.in.tumcampus, PID: 7479
    java.lang.NullPointerException: Attempt to invoke virtual method 'int de.tum.in.tumcampusapp.component.ui.ticket.payload.TicketReservationResponse.getTicketHistory()' on a null object reference
        at de.tum.in.tumcampusapp.component.ui.ticket.BuyTicketActivity$3.onResponse(BuyTicketActivity.java:182)
        at retrofit2.ExecutorCallAdapterFactory$ExecutorCallbackCall$1$1.run(ExecutorCallAdapterFactory.java:70)
        at android.os.Handler.handleCallback(Handler.java:751)
        at android.os.Handler.dispatchMessage(Handler.java:95)
        at android.os.Looper.loop(Looper.java:154)
        at android.app.ActivityThread.main(ActivityThread.java:6077)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:866)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:756)**

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/EventDetailsFragment.java
schulzalexander: Well, at least for the TUM Calendar, when I look at the "initStartEndDates" method it seems like both "start" and "end" have to be set to something. For the google calendar, you could leave the extra away and it would use a default duration of one our, but I would just provide the same default duration there and keep it consistens with the tum calendar.

app/src/main/java/de/tum/in/tumcampusapp/utils/Const.kt
schulzalexander: I moved that to the companion object of the Event class.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/EventDetailsFragment.java
kordianbruck: I see, that makes sense. A comment for that would've been nice, but it should be alright.

app/src/main/java/de/tum/in/tumcampusapp/database/Converters.kt
kordianbruck: @schulzalexander what about this?

schulzalexander: @kordianbruck sorry forgot that.
You mean like
```
@TypeConverter
    fun isoToDateTime(str: String?): DateTime? {
        return if (str == null) null else DateTimeUtils.getDateTime(str)
    }

    @TypeConverter
    fun fromDateTime(date: DateTime?): String? {
        return if (date == null) null else DateTimeUtils.getDateTimeString(date)
    }
```
?

kordianbruck: Looks good yeah. Make sure the Model fields are nullable as well.

schulzalexander: ok, yes the "end" property is nullable already.

schulzalexander: @kordianbruck pushed that change

PR #981, opened 2018-07-10
As/refactor ticket sale

schulzalexander:
This includes refactorings of all kinds for the whole ticketsale feature, including

- rearranging of the package structure
- removing the ticket scan (will be included in the admin app)
- resolving todos
- removing unnecessary imports
- changing the date format to use DateTimeFormat
- etc


saileru:
@schulzalexander The mail formatting is still a bit weird. What is the status there

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/model/Event.kt
```diff
@@ -34,6 +35,18 @@ data class Event(@PrimaryKey
         const val defaultDuration = 7200000 // Milliseconds
     }
 
+    fun getFormattedDateTime(): String {
+        return "${getFormattedDate()}, ${getFormattedTime()}"
+    }
+
+    fun getFormattedDate(): String {
+        return DateTimeFormat.forPattern("dd MMMM yyyy").print(start)
```
kordianbruck: You are not respecting locale here. Instead something like `DateTimeFormat.mediumDate` should be used instead of statically setting the format

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/model/Event.kt
```diff
@@ -34,6 +35,18 @@ data class Event(@PrimaryKey
...
+    }
+
+    fun getFormattedTime(): String {
+        return DateTimeFormat.forPattern("HH:mm").print(start)
```
kordianbruck: Same here - how about am/pm here?

app/src/main/res/values/strings.xml
```diff
@@ -621,7 +621,7 @@ Signature: %5$s</string>
     <string name="external_calendar">External Calendar</string>
     <string name="add_to_calendar_info">To which calendar do you want to add the event?</string>
     <string name="event_export" translatable="false">Event Export</string>
-    <string name="scanner">Scanner</string>
+    <string name="update_failure_message">Update failed.</string>
```
kordianbruck: This error message is not very descriptive.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/model/Event.kt
saileru: return DateTimeFormat.shortTime().print(start)

This will respect the current locale, e.g. "2:00 pm" vs "14:00"

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/model/Event.kt
```diff
@@ -34,6 +35,18 @@ data class Event(@PrimaryKey
         const val defaultDuration = 7200000 // Milliseconds
     }
 
+    fun getFormattedDateTime(): String {
+        return "${getFormattedDate()}, ${getFormattedTime()}"
```
saileru: If we use the new date formatting as suggested in my comment, I think we should remove the comma here

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/model/Event.kt
saileru: return DateTimeFormat.mediumDate().print(start)

"05.06.2018" vs "Jun 5, 2018"

saileru:
Here are are few more ideas for improvements 
- [] "Ok" button in AlertDialog in BuyTicketActivity (In my opinion it is nicer from a usability standpoint to be able to click on "ok". Just clicking outside the dialog to dismiss it feels weird)
- [] Redemption string in ShowTicketActivity should be a Date
- [] Get redemptionDate from Server (always set to false currently...)
- [] Maybe we can hide the image view for a null image in the EventDetailsFragment

These points do not necessarily need to be implemented in this PR. However, we could add them as tickets for future improvement.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/model/Event.kt
pfent: @kordianbruck I think this actually respects the default locale by default.

saileru: Yes, it kind of does. It prints
"05 Juni 2018" vs "05 June 2018".
However, to me it seems nicer to me to have
"05.06.2018" vs. "June 05, 2018" what my suggestion does if I remember correctly

schulzalexander: Right, will change that.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/model/Event.kt
schulzalexander: Yeah right, but the problem of using mediumTime() here is that it will return a time like "2:00 nachm" instead of "14:00", apparently that's some kind of bug. We can instead ask the DateFormat class whether 24 hour format should be used, and then set the pattern accordingly.

You can have a look when I have commited that.

app/src/main/res/values/strings.xml
schulzalexander: I'll remove it, since it is not really used anymore.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/model/Event.kt
schulzalexander: see answer above this post. shortTime will return "2:00 nachm." instead of "14:00"

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/model/Event.kt
schulzalexander: Agree

schulzalexander:
@saileru regarding the mail, I find it hard to recreate this layout error.
The mail I received from the server also looked weird like that, but when I send the very same HTML template from my local mail client to myself, it looks just fine.

So it would be nice if someone with access to the mailserver could test this.

schulzalexander:
@saileru thanks for the checklist, I added this stuff, we only need to wait until the PR on the backend is approved to return the actual Date of the redemption ;)

schulzalexander:
@kordianbruck I also added the 3 files that were missing from the .idea directory

PR #984, opened 2018-07-11
Ticket Sale: fixed some leftovers

saileru:
* replaced the thread to getTicketTypes by retrofit callback
* fixed leftover date formatting in BuyTicketActivity
* added Ok button to alert dialog for "Not logged in"
* fixed NullPointerException (when I tried to buy a ticket again which I have previously purchased but not fetched from the server response.body() returned null...)
* removed ticket button from kino layouts

@kordianbruck We should always use Utils.log(t) instead of t.printstackstrace()?

I seemed to have a problem with retrieving purchased tickets form the server. I got an "Unatuthorized" response. I thought this problem was fixed in https://github.com/kordianbruck/tca-backend/pull/44?
Thus, currently I cannot retrieve Tickets from Server at startup...
UPDATE (2018-07-13 14:22): I reinstalled the app and created a new token. Now it works again


app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/activity/BuyTicketActivity.java
```diff
@@ -56,50 +55,59 @@ protected void onCreate(Bundle savedInstanceState) {
...
-        }
+                    @Override
+                    public void onFailure(Call<List<TicketType>> call, Throwable t) {
+                        // if ticketTypes could not be retrieved from server, e.g. due to network problems
+                        Utils.log(t);
+                        Utils.showToast(getApplicationContext(), R.string.no_network_connection);
+                        // go back to event details
+                        Intent intent = new Intent(getApplicationContext(), EventDetailsActivity.class);
+                        intent.putExtra("event_id", eventId);
+                        startActivity(intent);
```
schulzalexander: you can also just call finish() here instead of starting the intent.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/activity/BuyTicketActivity.java
```diff
@@ -174,16 +182,25 @@ private void reserveTicket() {
...
+                        AlertDialog.Builder builder = new AlertDialog.Builder(BuyTicketActivity.this);
+                        builder.setTitle(getString(R.string.sorry))
+                                .setMessage(getString(R.string.ticket_not_fetched))
+                                .setPositiveButton(R.string.ok, (dialog, which) -> {
+                                    reservationProgressBar.setVisibility(View.INVISIBLE);
+                                    paymentButton.setEnabled(true);
+                                });
+                        AlertDialog alertDialog = builder.create();
+                        alertDialog.show();
+                    }else if(response.body().getError() != null){
```
schulzalexander: code formatting^^

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/activity/BuyTicketActivity.java
```diff
@@ -56,50 +55,59 @@ protected void onCreate(Bundle savedInstanceState) {
...
-                    alertDialog.show();
-                }
-            });
-        }
+                    @Override
+                    public void onFailure(Call<List<TicketType>> call, Throwable t) {
+                        // if ticketTypes could not be retrieved from server, e.g. due to network problems
+                        Utils.log(t);
+                        Utils.showToast(getApplicationContext(), R.string.no_network_connection);
+                        // go back to event details
```
schulzalexander: Please also use R.string.no_internet_connection here as suggested in #973, as our no_network_connection is kind of a duplicate to that.
I've already removed no_network_connection from the strings.xml files in my PR #989

saileru: Done

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/activity/BuyTicketActivity.java
saileru: Ah nice!

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/activity/BuyTicketActivity.java
saileru: Yeah, good point ^^

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/EventsController.java
```diff
@@ -60,27 +60,43 @@ public void downloadFromService(boolean force) {
...
+    private void loadTicketTypesForTickets(List<Ticket> tickets){
+        // get ticket type information for all tickets
+        for (Ticket ticket : tickets){
+            TUMCabeClient.getInstance(context).getTicketTypes(ticket.getEventId(),
+                    new Callback<List<TicketType>>(){
+
+                        @Override
+                        public void onResponse(Call<List<TicketType>> call, Response<List<TicketType>> response) {
+                            List<TicketType> ticketTypes = response.body();
+
```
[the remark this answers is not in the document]
saileru: Done

saileru:
I also thought that we should probably add a server call for tickets in onRefresh in EventFragment. Reloading them from the database as it is currently implemented does not seem to make much sense...
But we can do that in a separate PR. Otherwise it might get too confusing

PR #986, opened 2018-07-12
Make chaosmonkey improvements

thellmund:
This pull request adds more information to the `ChaosMonkeyException` output and properly implements error handling for some API calls.

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/api/app/ChaosMonkeyInterceptor.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/api/app/ChaosMonkeyInterceptor.kt
@@ -21,8 +21,10 @@ class ChaosMonkeyInterceptor : Interceptor {
     override fun intercept(chain: Interceptor.Chain): Response {
         if (ThreadLocalRandom.current().nextDouble() < failProbability) {
             Utils.log("Chaos Monkey is resilience testing your network code")
-            throw ChaosMonkeyException()
+            val url = chain.call().request().url().toString()
+            throw ChaosMonkeyException(url)
         }
         return chain.proceed(chain.request())
     }
+
 }
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/ui/news/KinoViewModel.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/ui/news/KinoViewModel.kt
@@ -56,7 +56,7 @@ class KinoViewModel(private val localRepository: KinoLocalRepository,
                         .doOnError { Utils.log(it) }
                         .flatMapIterable { it }
                         .filter { it.isFutureMovie() }
-                        .subscribe { localRepository.addKino(it) }
+                        .subscribe({ localRepository.addKino(it) }, { t -> Utils.log(t) })
         )
     }
 
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/ui/news/TopNewsViewModel.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/ui/news/TopNewsViewModel.kt
@@ -6,7 +6,6 @@ import de.tum.`in`.tumcampusapp.component.ui.news.repository.TopNewsRemoteReposi
 import de.tum.`in`.tumcampusapp.component.ui.overview.CardManager
 import de.tum.`in`.tumcampusapp.utils.Const
 import de.tum.`in`.tumcampusapp.utils.Utils
-import io.reactivex.Observable
 import io.reactivex.disposables.CompositeDisposable
 import io.reactivex.schedulers.Schedulers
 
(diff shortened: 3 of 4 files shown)


app/src/main/java/de/tum/in/tumcampusapp/api/app/ChaosMonkeyInterceptor.kt
```diff
@@ -21,8 +21,10 @@ class ChaosMonkeyInterceptor : Interceptor {
     override fun intercept(chain: Interceptor.Chain): Response {
         if (ThreadLocalRandom.current().nextDouble() < failProbability) {
             Utils.log("Chaos Monkey is resilience testing your network code")
-            throw ChaosMonkeyException()
+            val url = chain.call().request().url().toString()
+            throw ChaosMonkeyException(url)
```
pfent: That's actually a really good idea for debuggability :+1:

app/src/main/java/de/tum/in/tumcampusapp/component/ui/news/TopNewsViewModel.kt
```diff
@@ -16,27 +15,33 @@ import io.reactivex.schedulers.Schedulers
...
-                })
-        )}
+                            // There is a NewsAlert update if the image link or the date changed
+                            // --> Card should be displayed again
+                            val update = oldShowUntil != newsAlert.displayUntil || oldImage != newsAlert.url
+                            if (update) {
+                                Utils.setSetting(context, CardManager.SHOW_TOP_NEWS, true)
+                            }
+                            Utils.setSetting(context, Const.NEWS_ALERT_SHOW_UNTIL, newsAlert.displayUntil)
+                        }, { t -> Utils.log(t) })
```
pfent: We already have a `.doOnError { Utils.log(it) }` call in here, we can probably remove that one then...

app/src/main/java/de/tum/in/tumcampusapp/component/ui/news/KinoViewModel.kt
```diff
@@ -56,7 +56,7 @@ class KinoViewModel(private val localRepository: KinoLocalRepository,
                         .doOnError { Utils.log(it) }
                         .flatMapIterable { it }
                         .filter { it.isFutureMovie() }
-                        .subscribe { localRepository.addKino(it) }
+                        .subscribe({ localRepository.addKino(it) }, { t -> Utils.log(t) })
```
pfent: we also have a `.doOnError { Utils.log(it) }` here...

app/src/main/java/de/tum/in/tumcampusapp/component/ui/news/TopNewsViewModel.kt
thellmund: You’re right – I forgot to remove that 🙈 

We want to go with the new one (`onError`), because it actually consumes the error. `doOnError` does not.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/news/KinoViewModel.kt
thellmund: See reasoning above ☝️

app/src/main/java/de/tum/in/tumcampusapp/component/ui/news/TopNewsViewModel.kt
pfent: So i guess search + replace `doOnError` -> `onError`

thellmund: Sorry, I should be clearer: There is no single `onError` method. It’s a `Consumer` that you pass to `subscribe`.

However, there are other ways to achieve the same thing: We could use `onErrorReturnItem` or `onErrorResumeNext` instead and return an empty `Observable` or `Flowable`. I’ll have to investigate if that makes any difference between those and my current implementation.

thellmund: @pfent In the meantime, please remove the approval for this branch again.

pfent: So generally, we want to use an `onError` consumer, to properly consume everything...

I'm not a fan of `onErrorReturn` or `onErrorResumeNext`, since we probably should do proper error handling with dedicated UI elements that would be shown in the error subscriber...

thellmund: There are a few things here that make proper error handling difficult. Let’s talk in person tomorrow about how we can refactor this.

thellmund:
@pfent Please review this branch. I’d like to merge it soon, as there are still Retrofit-related issues in 1.6.1.

PR #988, opened 2018-07-13
Update libraries

kordianbruck:
By specifying dependencies individually we can update them using our bot

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -79,9 +79,6 @@ android {
 
 def androidSupportVersion = '27.1.1'
 def retrofitVersion = '2.3.0'
-def firebaseVersion = '15.0.0'
-def roomVersion = '1.1.0'
-def ankoVersion='0.10.5'
 
 configurations.all {
     resolutionStrategy {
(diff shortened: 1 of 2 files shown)


app/build.gradle
```diff
@@ -79,9 +79,6 @@ android {
 
 def androidSupportVersion = '27.1.1'
 def retrofitVersion = '2.3.0'
```
pfent: what about those?

app/build.gradle
```diff
@@ -79,9 +79,6 @@ android {
...
-def firebaseVersion = '15.0.0'
```
thellmund: Why did you remove those?

app/build.gradle
thellmund: GSON is also available in a new version and can be updated.

kordianbruck: Dependabot should handle those.

pfent: so why do we need to specify the other dependencies individually, but those are OK being specified as variables?

kordianbruck: Because only firebase and android architecture components announced, that they will no longer simultaneously update their libs.

PR #992, opened 2018-07-13
Remove maxHeight on image view to make it adapt to image aspect ratio

thellmund:
This commit makes sure that the TopNews image is displayed with the correct aspect ratio.


kordianbruck:
Sure, why not. Dunno why it was in there to begin with.

PR #879, opened 2018-06-09
Add intelligent notification scheduling

thellmund:
## Issue

This fixes the following issue(s): #879 
- Move notification scheduling responsibility out of Card classes.
- More intelligently schedule notifications about the cafeteria menu, tuition fees and calendar events. For instance, cafeteria menu notifications appear at 11am on weekdays. Notifications about the deadline for tuition fees are increased towards the deadline, reminding the student to pay. 

## Why this is useful for all students
- Smarter notification scheduling provides the user with the right information at the right time.

## Notes
- `TransportNotificationsProvider` currently returns no notifications, as it will need to be further refactored to provide intelligent scheduling. In the future, we might show transport notifications 15 minutes before the end of the day’s last lecture if the user is at the location of the lecture.

PR #995, opened 2018-07-14
Fix NotificationAwareCard crash

thellmund:
We currently receive the 40 next departures from the MVV API. Until now, all these departures were included in the MVV notification as pages (for Android Wear). This caused a TransactionTooLargeException. With this commit, we only display the first five departures in the notification to prevent the crash.


thellmund:
@pfent Yes, we can do a proper refactoring of this at some point (also with regards to #879). For now, I just want to fix the bug that resulted in the `MainActivity` staying empty.

PR #996, opened 2018-07-14
Properly finish setup wizard

### LINKED ISSUE #935: Pressing back after completing wizard goes back to wizard
should not go back to wizard, e.g. close app

thellmund:
This commit fixes #935. Currently, the user can return to the setup wizard by pressing the back button in MainActivity. To avoid this, we start `StartupActivity` in a new `Task` and remove the `Task` to which the setup wizard activities belong.


pfent:
Something like a Wizard [state machine transition table](https://en.wikipedia.org/wiki/State_transition_table) probably would come in handy for this kind of stuff

Issue #980, opened 2018-07-10
Ticket Sale: Integrate events to CardView


saileru:
CardView is the main screen, once you open the app. We can supply an event card for that view so the next event is always visible there.

saileru:
Closed in #997

PR #997, opened 2018-07-14
Ticket Sale: integrated EventCard to CardView

saileru:
## Issue

This fixes the following issue(s): #980

## Comments

I integrated the events to the card view to the best of my knowledge. The earliest event is now shown in the MainActivity.
However, I do not have a good overview of all the interdependencies which I have to consider. It might very well be that I missed important cases.

The EventCard is not NotificationAware yet. Can we add that later if needed?

@kordianbruck Do you want this issue ready before the deadline tomorrow? If we are not sure if this PR could break something, we should maybe add it later. Maybe we should put the priority on finishing the remaining issues for the Admin app?

## Screenshot

screenshot_20180714-121127

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/EventCard.java
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/EventCard.java
@@ -59,8 +59,8 @@ public class EventCard extends Card {
 
     @Override
     protected void discard(@NotNull SharedPreferences.Editor editor) {
-        // NOTE: We would need to add a dismissed attribute to the event table.
-        //       Is this necessary? (see NewsController for reference)
+        // TODO: add a dismissed attribute to the event table later
+        //       see NewsCard for reference
         //       Do nothing for now...
     }
 }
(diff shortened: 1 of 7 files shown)


thellmund:
No need to make it a `NotificationAwareCard`, as those will be removed with #879. Please update the design as specified in my feedback of #973.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/EventCard.java
```diff
@@ -17,13 +27,40 @@ public void setEvent(Event event) {
...
         Intent intent = new Intent(getContext(), EventDetailsActivity.class);
         intent.putExtra("event_id", event.getId());
         return intent;
     }
+
+    @Override
+    protected void discard(@NotNull SharedPreferences.Editor editor) {
+        // NOTE: We would need to add a dismissed attribute to the event table.
+        //       Is this necessary? (see NewsController for reference)
+        //       Do nothing for now...
```
kordianbruck: Hmm, maybe long term we want to change this behaviour. For now its goo.d

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/EventDao.java
```diff
@@ -24,8 +24,9 @@
     @Query("SELECT * FROM event ORDER BY start")
     List<Event> getAll();
 
-    @Query("SELECT id FROM event ORDER BY id DESC LIMIT 1")
-    int getLatestId();
+    // Get earliest event
+    @Query("SELECT * FROM event ORDER BY start LIMIT 1")
```
kordianbruck: You want all events *after* today. So you will need to add a `where` clause. Then the function name also fits better ;)

PR #1003, opened 2018-07-14
Ticket Sale: feedback from PR

saileru:
implement Till's feedback from #973


kordianbruck:
@saileru @schulzalexander @Kailiangdong any chance you will implement the remaining feedback?

saileru:
@kordianbruck Generally, we would definitely like to finish the implementation of the feedback. However, I am currently quite busy. I hope we will have some time this weekend...

Kailiangdong:
@kordianbruck I would definitely like to finish all and make them to perfect situation. But I'd like to do it after 2nd August after exam. But I will see, maybe still spend some time on it before exam

PR #1012, opened 2018-07-21
Optimize resources to reduce APK size

thellmund:
The release APK size of 1.6.2 is 7.8 MB. By removing unused images and converting most of the remaining ones to JPG, I reduced the APK size to 5.6 MB.

**Not optimized**
screenshot 2018-07-21 at 09 44 45

**Optimized**
screenshot 2018-07-21 at 09 52 27

---

We can further reduce the APK size. Android Wear 2.0 notifications no longer use the background image of the `Notification` ([source](https://developer.android.com/reference/android/support/v4/app/NotificationCompat.WearableExtender#setBackground(android.graphics.Bitmap))), so we can remove a couple of images used as notification backgrounds and get the APK size to around 5 MB. I will do this in #879.

We might also try [Android App Bundles](https://developer.android.com/platform/technology/app-bundle/) in the next release, @kordianbruck.


pfent:
I really thought a enabled [`shrinkResources`](https://developer.android.com/studio/build/shrink-code#shrink-resources), but it currently seems to be missing...
With it enabled, even the unused strings and drawables shouldn't be included in the APK.
Could you please enable it?

Pictures of text, i.e. the widget previews, and more generally everything with hard borders, have some really noticeable artifacts as JPGs, that really isn't great for high DPI screens...

thellmund:
You’re right, I forgot about resource shrinking 🙈 

I can’t make out any artifcats on a Pixel 2 XL.

thellmund:
We might also consider using [WebP](https://developer.android.com/topic/performance/reduce-apk-size#use-webp). What do you think, @pfent?

thellmund:
Android Studio offers a conversion tool from PNG/JPG to WebP. That should be good enough for us.

I found that using WebP did not reduce the APK size quite as much as converting the files to JPG, but 🤷‍♂️

kordianbruck:
@thellmund I had a little heart attack, when you said you converted to JPG in your first message. Usually digital graphics are smaller using the PNG format. WEBP is a good compromise, as it has both: lossy and lossless compression with that solid VP8 algo. 

So WebP all the way :+1:

kordianbruck:
Also: app bundles are not possible with an app signing key, but only with an upload key. So right now this is not an option for us.

PR #1018, opened 2018-07-25
Do MVV calls with Retrofit

### LINKED ISSUE #977: Refactor MVV API calls
We’re currently using our `NetUtils` to call the MVV API. We should switch to Retrofit and use GSON to parse the result into `Departure` objects. This will help us streamline error handling and provide higher stability.

pfent:
fixes #977 

The `MvvStationListSerializer` is a bit ugly. 
@thellmund do you know, if there's a better way to parse schema-free JSON, then this?

I also added some unit tests, so the parsing should be quite stable...

When testing I also noticed, that `TransportationDetailsActivity` is really not great, since we're doing a lot of work in `onLoadFinished()`.
We probably need to switch that to a RecyclerView sometimes, so we don't drop frames like crazy


app/src/main/java/de/tum/in/tumcampusapp/component/ui/transportation/api/MvvApiService.kt
```diff
@@ -0,0 +1,72 @@
+package de.tum.`in`.tumcampusapp.component.ui.transportation.api
```
kordianbruck: Would this maybe not be better in the API package, where we have the other APIs?

app/src/main/java/de/tum/in/tumcampusapp/component/ui/transportation/api/MvvApiService.kt
```diff
@@ -0,0 +1,72 @@
...
+     */
+    @GET("XSLT_DM_REQUEST?outputFormat=JSON" +
+            "&stateless=1" +
+            "&coordOutputFormat=WGS84" +
+            "&type_dm=stop" +
+            "&itOptionsActive=1" +
+            "&ptOptionsActive=1" +
+            "&mergeDep=1" +
+            "&useAllStops=1" +
+            "&mode=direct")
```
kordianbruck: Not the prettiest, but I guess there is no alternative - using default parameters for `@Query` seems overkill for things that should never change. :+1:

app/src/main/java/de/tum/in/tumcampusapp/component/ui/transportation/api/MvvClient.kt
```diff
@@ -0,0 +1,37 @@
...
+import android.content.Context
+import com.google.gson.GsonBuilder
+import de.tum.`in`.tumcampusapp.api.app.Helper
+import org.joda.time.DateTime
+import retrofit2.Retrofit
+import retrofit2.converter.gson.GsonConverterFactory
+
+class MvvClient {
+    companion object {
+        private const val BASE_URL = "http://efa.mvv-muenchen.de/mobile/" // No HTTPS support :(
```
kordianbruck: Are we sure, that there is no https? It just seemed to work a minute ago, but now I'm getting an error message.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/transportation/TransportController.kt
```diff
@@ -0,0 +1,176 @@
...
+
+        /**
+         * Find stations by station name prefix
+         *
+         * @param prefix Name prefix
+         * @return List of StationResult
+         */
+        fun getStationsFromExternal(context: Context, prefix: String): List<StationResult> {
+            try {
+                val (results) = MvvClient.getInstance(context).getStations(prefix).execute().body()
```
thellmund: Why the parentheses?

app/src/main/java/de/tum/in/tumcampusapp/component/ui/transportation/TransportationActivity.java
```diff
@@ -47,7 +47,7 @@ public void onCreate(Bundle savedInstanceState) {
 
         // Initialize stations adapter
         List<Recent> recentStations = recentsDao.getAll(RecentsDao.STATIONS);
-        adapterStations = new ArrayAdapter<>(this, android.R.layout.simple_list_item_1, TransportController.getRecentStations(recentStations));
+        adapterStations = new ArrayAdapter<>(this, android.R.layout.simple_list_item_1, TransportController.Companion.getRecentStations(recentStations));
```
thellmund: Consider adding `@JvmStatic` before the method.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/transportation/api/MvvDateSerializer.kt
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Parses the MVV specific dateTime, with json elements of:
+ * year, month, day, hour, minute
+ */
+class MvvDateSerializer : JsonDeserializer<DateTime> {
+
+    override fun deserialize(json: JsonElement?, typeOfT: Type?, context: JsonDeserializationContext?): DateTime {
+        if (json !is JsonObject) {
+            throw JsonParseException("Invalid MvvDateTime: $json")
```
thellmund: Does it make sense ton throw an `IOException` instead, @pfent? We can’t react to runtime exceptions with Retrofit, so this would result in a crash of the app. If we used an `IOException`, we could call `showErrorLayout()` in case of an exception.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/transportation/api/MvvDateSerializer.kt
```diff
@@ -0,0 +1,31 @@
...
+        val hour = json.get("hour")?.asInt.orThrow(json)
+        val minute = json.get("minute")?.asInt.orThrow(json)
+
+        return DateTime()
+                .withDate(year, month, day)
+                .withTime(hour, minute, 0, 0)
+    }
+
+    private fun <T> T?.orThrow(json: JsonElement): T {
+        return this ?: throw JsonParseException("Invalid MvvDateTime: $json")
```
thellmund: Same as above.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/transportation/api/MvvStationListSerializer.kt
```diff
@@ -0,0 +1,47 @@
...
+import de.tum.`in`.tumcampusapp.component.ui.transportation.model.efa.StationResult
+import java.lang.reflect.Type
+
+/**
+ * Parses the weird MVV XML_STOPFINDER_REQUEST response
+ */
+class MvvStationListSerializer : JsonDeserializer<MvvStationList> {
+    override fun deserialize(json: JsonElement?, typeOfT: Type?, context: JsonDeserializationContext?): MvvStationList {
+        if (json !is JsonObject) {
+            throw JsonParseException("Invalid MvvStationList: $json")
```
thellmund: Again, maybe an `IOException` instead?

app/src/main/java/de/tum/in/tumcampusapp/component/ui/transportation/api/MvvStationListSerializer.kt
```diff
@@ -0,0 +1,47 @@
...
+                parseStationResult(it as JsonObject)
+            }
+            return MvvStationList(resultList)
+        }
+
+        throw JsonParseException("Unknown MvvStationList: $json")
+    }
+
+
+    private fun parseStationResult(json: JsonObject): StationResult {
```
thellmund: Could this be a class method on `StationResult`, e.g. `StationResult.fromJson(json)`?

app/src/main/java/de/tum/in/tumcampusapp/component/ui/transportation/api/MvvDateSerializer.kt
kordianbruck: You could even make it extend `IOException` and use a custom exception message.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/transportation/api/MvvClient.kt
pfent: Ok, seems like they have HTTPS now, changeing that

app/src/main/java/de/tum/in/tumcampusapp/component/ui/transportation/TransportController.kt
pfent:
```kotlin
// This is a MvvStationList
val results = (...)

// results now is that MvvStationList unpacked into a List<StationResult>
val (results) = (...)
```

We actually receive a `MvvStationList` to trigger parsing by the `MvvStationListSerializer`, but the station list is just a wrapper around `List<StationResult>`, so we need to unpack that list

app/src/main/java/de/tum/in/tumcampusapp/component/ui/transportation/api/MvvDateSerializer.kt
pfent: You're both wrong. :dagger: 
And you should feel bad about it.

The `JsonParseException` is *literally* declared in Gson's `JsonDeserializer<T>` interface and the recommended way of error handling when deserializing JSON with Gson.

Gson's Javadoc states:
`@throws JsonParseException if json is not in the expected format of {@code typeofT}`

app/src/main/java/de/tum/in/tumcampusapp/component/ui/transportation/api/MvvDateSerializer.kt
pfent: Same as above. :wink:

app/src/main/java/de/tum/in/tumcampusapp/component/ui/transportation/api/MvvStationListSerializer.kt
pfent: Same same

app/src/main/java/de/tum/in/tumcampusapp/component/ui/transportation/api/MvvDateSerializer.kt
thellmund: @pfent How do those exceptions propagate through the system? As they’re not subclasses of IOException, Retrofit does not invoke the onFailure() callback when they occur.

pfent: @thellmund I actually don't know.
Since Retrofit supports multiple deserialization mechanisms, I guess they handle erroneous parsing and handle that as a failure.
Alas our `DateSerializer` also throws a `JsonParseException` and failure handling there works fine.

Nevertheless, parsing data is still not in any way an I/O operation.

PR #1023, opened 2018-07-26
Fix NullPointerException when clicking an event

thellmund:
When clicking an event in the week view, the corresponding `CalendarItem` is retrieved from the database. If the query returns null for whatever reason, the app crashes (happened very rarely). I’ve added a null check to prevent this.

Additionally, instead of using the event's start and end time to query for the `CalendarItem`, the proper event ID is now used.


app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarDao.java
```diff
@@ -70,10 +70,6 @@
             "ORDER BY dtend, dtstart LIMIT 4")
     List<CalendarItem> getNextCalendarItems();
 
-    // TODO: migrate dtstart and dtend to be proper dates
-    @Query("SELECT * FROM calendar WHERE dtstart LIKE '%' || :start AND dtend LIKE '%' || :end ")
-    CalendarItem getCalendarItemByStartAndEndTime(String start, String end);
-
```
pfent: good riddance

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarController.java
```diff
@@ -186,10 +186,9 @@ public void deleteLectureFromBlacklist(int widgetId, String lecture) {
         return lectures;
     }
 
-    public CalendarItem getCalendarItemByStartAndEndTime(DateTime start, DateTime end) {
-        String startString = DateTimeUtils.INSTANCE.getDateTimeString(start);
-        String endString = DateTimeUtils.INSTANCE.getDateTimeString(end);
-        return calendarDao.getCalendarItemByStartAndEndTime(startString, endString);
+    @Nullable
+    public CalendarItem getCalendarItemById(String id) {
+        return calendarDao.getCalendarItemById(id);
```
pfent: IMHO `getCalendarItemById()` shouldn't be nullable, but I guess you can't guarantee that that ID exists.
Also I think the id should probably be an int :man_shrugging:

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarActivity.java
```diff
@@ -438,9 +438,16 @@ public void onEditEvent(@NotNull CalendarItem calendarItem) {
...
-                new DateTime(weekViewEvent.getStartTime()),
-                new DateTime(weekViewEvent.getEndTime()));
+        String eventId = weekViewEvent.getIdentifier();
+        if (eventId == null) {
+            return;
+        }
+
+        CalendarItem item = calendarController.getCalendarItemById(eventId);
+        if (item == null) {
+            return;
```
pfent: This shoudn't happen, if everything was set up correctly.
Could we add some logging here, so we maybe get a chance to debug this eventually?

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarActivity.java
```diff
@@ -269,7 +269,7 @@ private void refreshWeekView() {
             // Lets change some dimensions to best fit the view.
             mWeekView.setTextSize((int) TypedValue.applyDimension(TypedValue.COMPLEX_UNIT_SP, 16, getResources().getDisplayMetrics()));
             mWeekView.setEventTextSize((int) TypedValue.applyDimension(TypedValue.COMPLEX_UNIT_SP, 12, getResources().getDisplayMetrics()));
-            mWeekView.setXScrollingSpeed(0.4f);
+            mWeekView.setXScrollingSpeed(0.5f);
```
kordianbruck: any more info on this change?

thellmund: Oops, somehow slipped in there. I was just testing out something for myself and forgot to remove it.

app/src/main/res/layout/activity_calendar.xml
```diff
@@ -43,7 +43,7 @@
                     app:noOfVisibleDays="1"
                     app:textSize="14sp"
                     app:todayBackgroundColor="#1848adff"
-                    app:xScrollingSpeed="0.4" />
+                    app:xScrollingSpeed="0.5" />
```
kordianbruck: Also here: this slipped in

PR #1025, opened 2018-07-28
Refactor grades activity

thellmund:
We had crashes in GradesActivity due to nullability. I converted the class to Kotlin and improved the code.


app/src/main/java/de/tum/in/tumcampusapp/component/tumui/grades/GradesActivity.kt
```diff
@@ -0,0 +1,394 @@
...
+    private var exams: List<Exam>? = null
+    private var programIds: List<String>? = null
+
+    private var spinnerPosition = 0
+
+    private var isFetched = false
+
+    // everything for the charts
+    private var barMenuItem: MenuItem? = null
+    private var pieMenuItem: MenuItem? = null
```
pfent: lateinit?

I don't think we have a use-case where those being null makes sense?

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/grades/GradesActivity.kt
```diff
@@ -0,0 +1,394 @@
...
+import java.text.NumberFormat
+import java.util.*
+
+/**
+ * Activity to show the user's grades/exams passed.
+ */
+class GradesActivity : ActivityForAccessingTumOnline<ExamList>(R.layout.activity_grades) {
+
+    // exams data and list
+    private var exams: List<Exam>? = null
```
pfent: `exams` is passed around as argument several times and assigned but never read as member variable.

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/grades/GradesActivity.kt
```diff
@@ -0,0 +1,394 @@
...
+    private var programIds: List<String>? = null
```
pfent: same here

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/grades/GradesActivity.kt
```diff
@@ -0,0 +1,394 @@
...
+        }
+        return gradeDistribution
+    }
+
+    /**
+     * Initialize the spinner for choosing between the study programs.
+     */
+    private fun initSpinner(exams: List<Exam>) {
+        // Set Spinner data
+        val filters = arrayListOf<String>(getString(R.string.all_programs))
```
pfent: I'd use `mutableListOf()`

thellmund: Yep, that better describes the intentions 👍

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/grades/GradesActivity.kt
thellmund: We’ve had NullPointerExceptions due to them being null

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/grades/GradesActivity.kt
```diff
@@ -0,0 +1,391 @@
...
+        instanceState.putInt(KEY_SPINNER_POSITION, spinnerPosition)
+    }
+
+    /**
+     * Toggles between list view and chart view in landscape mode.
+     */
+    fun toggleInLandscape(view: View) {
+        val showChart = chartsContainer.visibility == View.GONE
+
+        //swipeRefreshLayout.visibility = if (showChart) View.GONE else View.VISIBLE
```
pfent: :bell::bell::bell: That's not a real comment

PR #1036, opened 2018-07-30
Update shortcut icons

thellmund:
This commit updates the shortcut icons to the proper styling. They now: 
1. use the primary color
2. are the proper size
3. have a light-gray background circle

**Old**

**New**


pfent:
how about ldpi and xxxhdpi? :trollface: 

Can we please stick to vector graphics for icons, so we have less hassle supporting future screen resolutions?

thellmund:
@pfent Both `ldpi` and `xxxhdpi` are generally not considered necessary, as _very_ few devices have these screen densities. Android Studio’s image asset tool also omits them (except for the launcher icon).

However, I understand that a vector drawable might be the better option. I’ll do the conversion.

kordianbruck:
@thellmund also with vector graphics we don't need to separately keep track of the source files for the binaries, but can just directly edit the svgs, which is nice.

PR #1077, opened 2018-09-13
Improve MVV card and details activity

thellmund:
This pull request makes minor UI improvements and fixes an incorrect behavior in `TransportationDetailsActivity`. It also removes the `TransportationActivity` from the navigation drawer.

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/other/generic/drawer/DrawerMenuHelper.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/other/generic/drawer/DrawerMenuHelper.kt
@@ -25,7 +25,6 @@ import de.tum.`in`.tumcampusapp.component.ui.openinghour.OpeningHoursListActivit
 import de.tum.`in`.tumcampusapp.component.ui.overview.InformationActivity
 import de.tum.`in`.tumcampusapp.component.ui.studyroom.StudyRoomsActivity
 import de.tum.`in`.tumcampusapp.component.ui.ticket.activity.EventsActivity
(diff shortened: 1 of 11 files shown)


app/src/main/java/de/tum/in/tumcampusapp/component/other/generic/drawer/DrawerMenuHelper.kt
```diff
@@ -86,7 +85,6 @@ class DrawerMenuHelper(private val mContext: Context, private val mDrawerLayout:
                 SideNavigationItem(R.string.menues, R.drawable.ic_cutlery, CafeteriaActivity::class.java, false, false),
                 SideNavigationItem(R.string.study_rooms, R.drawable.ic_group_work, StudyRoomsActivity::class.java, false, false),
                 SideNavigationItem(R.string.roomfinder, R.drawable.ic_place, RoomFinderActivity::class.java, false, false),
-                SideNavigationItem(R.string.mvv, R.drawable.ic_mvv, TransportationActivity::class.java, false, false),
```
pfent: I think @kordianbruck vetoed the removal, but IMO we want to improve notifications and cards for that.

pfent: I'd additionally remove that "opening hours" activity, since we really don't update those too often

thellmund: It was @kordianbruck who also suggested the removal while keeping the card and the details activity.

PR #1080, opened 2018-09-17
Add logout option and update settings design

thellmund:
This commit adds a logout button (which replaces the option to clear the cache) and again applies the Material design to the settings (which got broken when updating to Support Library 28).


app/src/main/java/de/tum/in/tumcampusapp/component/other/settings/SettingsFragment.java
```diff
@@ -43,7 +45,8 @@
         implements SharedPreferences.OnSharedPreferenceChangeListener, Preference.OnPreferenceClickListener {
 
     public static final String FRAGMENT_TAG = "my_preference_fragment";
-    private static final String BUTTON_CLEAR_CACHE = "button_clear_cache";
+    //private static final String BUTTON_CLEAR_CACHE = "button_clear_cache";
```
kordianbruck: Just remove this stuff please.

app/src/main/java/de/tum/in/tumcampusapp/component/other/settings/SettingsFragment.java
```diff
@@ -241,24 +239,52 @@ public boolean onPreferenceClick(Preference preference) {
...
+
+    private void logout() {
+        clearData();
         Utils.setSetting(mContext, Const.EVERYTHING_SETUP, false);
 
         mContext.finish();
         startActivity(new Intent(mContext, StartupActivity.class));
     }
 
+    private void clearData() {
```
kordianbruck: This could go to separate class and should not be part of some random activity.

app/src/main/java/de/tum/in/tumcampusapp/component/other/settings/SettingsFragment.java
```diff
@@ -220,19 +224,13 @@ public boolean onPreferenceClick(Preference preference) {
...
-                        .setNegativeButton(R.string.cancel, null)
-                        .create();
-
-                if (dialog.getWindow() != null) {
-                    dialog.getWindow().setBackgroundDrawableResource(R.drawable.rounded_corners_background);
-                }
-
-                dialog.show();
+                showDialogWithAction(R.string.delete_cache_sure, R.string.delete, ((dialog, i) -> clearCache()));
+                break;
```
kordianbruck: just delete please

app/src/main/java/de/tum/in/tumcampusapp/component/other/settings/SettingsFragment.java
kordianbruck: Nevermind, I couldn't think of a better class for it.

PR #1084, opened 2018-09-17
Minor UI adjustment in StartupActivity

thellmund:
This pull request replaces the loading labels with a progress bar that is shown on the first app launch.

screen shot 2018-09-17 at 23 47 23


app/src/main/java/de/tum/in/tumcampusapp/service/DownloadService.kt
```diff
@@ -227,6 +227,11 @@ class DownloadService : JobIntentService() {
                     Const.TOP_NEWS -> success = service.downloadTopNews()
                     else -> {
                         success = service.downloadAll(force)
+
+                        if (success) {
+                            Utils.setSetting(service, Const.EVERYTHING_SETUP, true)
+                        }
+
                         val isSetup = Utils.getSettingBool(service, Const.EVERYTHING_SETUP, false)
                         if (isSetup) {
```
kordianbruck: Can this following code then be removed?

app/src/main/java/de/tum/in/tumcampusapp/service/DownloadService.kt
```diff
@@ -227,6 +227,11 @@ class DownloadService : JobIntentService() {
                     Const.TOP_NEWS -> success = service.downloadTopNews()
                     else -> {
                         success = service.downloadAll(force)
+
+                        if (success) {
+                            Utils.setSetting(service, Const.EVERYTHING_SETUP, true)
```
kordianbruck: What exactly is hidden behind this flag? It would be great to document it somewhere, as I don't think to recall what exactly the idea was and I'm not sure it is used here correctly.

kordianbruck:
This also fixes #1085 by removing the slogan (Always great to reference issues from PRs)

app/src/main/java/de/tum/in/tumcampusapp/service/DownloadService.kt
thellmund: It describes whether a full first launch of the app has occurred, including downloading all data. If this is not true, then `StartupActivity` displays the loading indicator, indicating a longer-than-normal loading.

app/src/main/java/de/tum/in/tumcampusapp/service/DownloadService.kt
thellmund: See comment below

app/src/main/java/de/tum/in/tumcampusapp/service/DownloadService.kt
kordianbruck: Why do we separately handle syncing the calendar below then?

kordianbruck: Especially why do we set this variable to true a second time in a if-branch that almost never will be executed?

PR #1086, opened 2018-09-17
Fix problem that resulted in a Bad Request to tickets API

thellmund:
While moving from `DeviceVerification` to `TUMCabeVerification`, I forgot to make an adjustment in `TUMCabeClient`, which resulted in `400 Bad Request` responses for requests to some ticket API endpoints. The reason was that I was not sending along some required data in the body.


kordianbruck:
I can't see any difference in logic. I see two unrelated changes plus a cleanup of a function that is there twice. Where exactly are you now sending more information than before?

PR #1087, opened 2018-09-18
Fix several problems in chat

thellmund:
We currently experience several problems in the chat: 
- Messages are not sent successfully, due to a past model change in `ChatMessage`.
- Undelivered message were not marked as such, but continued to show a progress indicator. 
- Undelivered messages could not be re-sent.

This pull request addresses those problems and attempts to also refactor some parts of the architecture to provide easier-to-understand code.


app/src/main/java/de/tum/in/tumcampusapp/api/app/TUMCabeClient.java
```diff
@@ -184,12 +184,16 @@ public void addUserToChat(ChatRoom chatRoom, ChatMember member, TUMCabeVerificat
     }
 
     public Observable<ChatMessage> sendMessage(int roomId, TUMCabeVerification verification) {
-        //If the id is zero then its an new entry otherwise try to update it
-        int id = ((ChatMessage) verification.getData()).getId();
-        if (id == 0) {
+        ChatMessage message = (ChatMessage) verification.getData();
+        if (message == null) {
```
kordianbruck: Ever hear of: https://docs.oracle.com/javase/7/docs/api/java/lang/ClassCastException.html

app/src/main/java/de/tum/in/tumcampusapp/api/app/TUMCabeClient.java
```diff
@@ -184,12 +184,16 @@ public void addUserToChat(ChatRoom chatRoom, ChatMember member, TUMCabeVerificat
...
+            throw new IllegalStateException("TUMCabeVerification data is not a ChatMessage");
+        }
+
+        if (message.isNewMessage()) {
             return service.sendMessage(roomId, verification);
         }
-        return service.updateMessage(roomId, id, verification);
+
+        return service.updateMessage(roomId, message.getId(), verification);
```
kordianbruck: I think this can also be reduced into one API. I can simply check the sendMessage API if it has a message ID.

app/src/main/java/de/tum/in/tumcampusapp/api/app/TUMCabeClient.java
thellmund: Couldn’t it still be null if no `data` has been set?

kordianbruck: Practically yes, but then the exception message makes no sense

thellmund: I think it does, as the method could be invoked with a `TUMCabeVerification` that has no `data` set. In this case, `IllegalStateException` seems like exactly the right exception to me. 

How would you handle it?

kordianbruck: C++ templates would be much better here. To properly handle it you would probably use https://stackoverflow.com/questions/7526817/use-of-instance-of-in-java

You can still check for a null value even before the cast and then throw the illegalstate.

PR #1093, opened 2018-09-19
Make tweaks to calendar design and behavior

thellmund:
- Update design of `CalendarDetailsFragment` (see below; let me know if something looks off on non-Pixel devices)
- Update to 2.0.1 of [my fork of WeekView](https://github.com/thellmund/Android-Week-View), which adds horizontal start and end margins
- Update behavior when creating a new calendar event (now uses the date last visible in the `WeekView` component)

---

Old: 
screenshot_20180919-170943

New: 
screenshot_20180919-171418

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CreateEventActivity.java
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CreateEventActivity.java
@@ -17,6 +17,7 @@ import android.widget.TextView;
 import android.widget.Toast;
 
 import org.joda.time.DateTime;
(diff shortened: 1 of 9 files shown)


app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CreateEventActivity.java
```diff
@@ -144,6 +160,7 @@ private void initStartEndDates(Bundle extras) {
             end = new DateTime(start)
                     .plusHours(1);
         }
+        */
```
kordianbruck: Just remove this code please

PR #1094, opened 2018-09-20
Add more calendar improvements

thellmund:
- A line now indicates the current time.
- A newly created event is now displayed immediately.
- A _today_ button now allows the user to quickly jump back to the current day.
- The overflow menu is simplified by reducing a single-item group to just the item itself.

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarActivity.java
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarActivity.java
@@ -101,9 +101,9 @@ public class CalendarActivity extends ActivityForAccessingTumOnline<EventsRespon
         mWeekView.setMonthChangeListener(this);
         mWeekView.setOnEventClickListener(this);
         mWeekView.setScrollListener((newFirstVisibleDay, oldFirstVisibleDay) -> {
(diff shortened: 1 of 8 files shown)


app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarActivity.java
```diff
@@ -89,13 +94,22 @@ public CalendarActivity() {
...
         mWeekView = findViewById(R.id.weekView);
 
         // The week view has infinite scrolling horizontally. We have to provide the events of a
         // month every time the month changes on the week view.
         mWeekView.setMonthChangeListener(this);
         mWeekView.setOnEventClickListener(this);
+        mWeekView.setScrollListener((newFirstVisibleDay, oldFirstVisibleDay) -> {
+            DateTime visibleDay = new LocalDate(newFirstVisibleDay).toDateTimeAtStartOfDay();
+            DateTime today = DateTime.now().withTimeAtStartOfDay();
+            boolean isToday = visibleDay.getMillis() == today.getMillis();
```
pfent: Comparing millis is really tricky. Since we're using Joda, please just use the `.equals()`
Also `LocalDate.now()` gives today as a date, so we don't need to call `withTimeAtStartOfDay()`

I think I implemented this properly in `CafeteriaWithMenus`, but since this class is not Kotlin, we can't really reuse the extension function...

PR #1100, opened 2018-09-26
Make improvements to cards

thellmund:
This pull request moves the card options from a long press, which interfered with swipe and drag gestures, to a dedicated button. 

---

screenshot_20180926-093252

screenshot_20180926-093257

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/ui/cafeteria/CafeteriaMenuViewHolder.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/ui/cafeteria/CafeteriaMenuViewHolder.kt
@@ -14,7 +14,7 @@ class CafeteriaMenuViewHolder(itemView: View) : CardViewHolder(itemView) {
             cafeteriaNameTextView.text = cafeteria.name
             menuDateTextView.text = DateTimeFormat.mediumDate().print(cafeteria.nextMenuDate)
 
-            if (contentContainerLayout.childCount == 1) {
+            if (contentContainerLayout.childCount == 0) {
                 // We have not yet added the cafeteria menu items to the card
                 CafeteriaDetailsSectionFragment.showMenu(contentContainerLayout,
                         cafeteria.id, cafeteria.nextMenuDate, false, cafeteria.menus)
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/ui/overview/card/CardViewHolder.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/ui/overview/card/CardViewHolder.kt
@@ -12,7 +12,9 @@ import android.widget.ImageView
 import de.tum.`in`.tumcampusapp.R
 import de.tum.`in`.tumcampusapp.component.other.navigation.NavigationManager
 import de.tum.`in`.tumcampusapp.component.other.settings.UserPreferencesActivity
+import de.tum.`in`.tumcampusapp.component.ui.overview.CardInteractionListener
 import de.tum.`in`.tumcampusapp.utils.Const
+import de.tum.`in`.tumcampusapp.utils.Utils
 
 open class CardViewHolder(itemView: View) : RecyclerView.ViewHolder(itemView) {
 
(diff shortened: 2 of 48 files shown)


thellmund:
Current build error is due to #1098. It’s fixed in #1099.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/overview/card/CardViewHolder.kt
```diff
@@ -1,67 +1,77 @@
...
+        moreIcon?.setOnClickListener { openOptionsPopup(it) }
     }
 
-    override fun onClick(v: View) {
-        val destination = currentCard?.getNavigationDestination()
-        destination?.let {
-            NavigationManager.open(activity, it)
+    private fun openOptionsPopup(anchorView: View) {
+        val card = currentCard ?: return
+        val menuId = if (card.hasSettings) {
```
pfent: I think a boolean property isn't the right abstraction for this...
If we want to go down this route, we probably should let the cards itself specify it's options popup layout. A boolean property just obscures what options the card actually needs to support.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/cafeteria/CafeteriaMenuViewHolder.kt
```diff
@@ -14,7 +14,7 @@ class CafeteriaMenuViewHolder(itemView: View) : CardViewHolder(itemView) {
             cafeteriaNameTextView.text = cafeteria.name
             menuDateTextView.text = DateTimeFormat.mediumDate().print(cafeteria.nextMenuDate)
 
-            if (contentContainerLayout.childCount == 1) {
+            if (contentContainerLayout.childCount == 0) {
```
pfent: :sparkles: magic numbers :sparkles:

app/src/main/res/layout/card_cafeteria_menu.xml
```diff
@@ -1,7 +1,67 @@
...
+            tools:ignore="UselessLeaf">
+
+            <!-- Cafeteria menu items will be inserted here -->
+
+        </LinearLayout>
+
+    </android.support.constraint.ConstraintLayout>
+
+    <!--
     <LinearLayout
```
pfent: commented code

app/src/main/java/de/tum/in/tumcampusapp/component/ui/overview/card/CardViewHolder.kt
thellmund: You’re right – I can change it so that every card provides its own menu: 
`menuInflater.inflate(currentCard.optionsMenu, menu)`

Is that what you had in mind, @pfent?

app/src/main/res/layout/card_cafeteria_menu.xml
thellmund: By now, when someone asks about my biggest weaknesses, I can add “always forget commented code” to that list ;)

Will remove it.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/overview/card/CardViewHolder.kt
pfent: Something like that, yes

app/src/main/res/layout/card_cafeteria_menu.xml
pfent: :joy: That's what reviews are for.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/cafeteria/CafeteriaMenuViewHolder.kt
thellmund: Added a clarifying comment

kordianbruck: Would love to see that Clarifying comment - where did you add it? :grinning:

thellmund: [here](https://github.com/TCA-Team/TumCampusApp/pull/1100/commits/8b3504a4f4dc46e8d4e52b3615119f665321d052)

app/src/main/java/de/tum/in/tumcampusapp/component/ui/overview/card/CardViewHolder.kt
kordianbruck: We could then even make those menus a bit more useful eventually - good thinking :+1:

PR #1103, opened 2018-09-28
Fixes in MVV widget

thellmund:
This pull request was intended to only simplify `MVVSymbolView` (or rather: move from `MVVSymbolView` to `MVVSymbol`, as the former only encapsulated information and was never actually drawn to the screen). 

In the process, I noticed problems with the MVV widget (error “App isn’t installed”), which I also fixed in this pull request.


app/src/main/java/de/tum/in/tumcampusapp/component/ui/transportation/DepartureView.java
```diff
@@ -91,26 +91,25 @@ public DepartureView(Context context, boolean big) {
      *
      * @param symbol Symbol e.g. U6, S1, T14
      */
-    @SuppressWarnings("deprecation")
     public void setSymbol(String symbol, boolean highlight) {
-        MVVSymbolView d = new MVVSymbolView(symbol);
-        mSymbolView.setTextColor(d.getTextColor());
+        MVVSymbol mvvSymbol = new MVVSymbol(symbol);
```
pfent: And proper variable names :tada:

app/src/main/java/de/tum/in/tumcampusapp/component/ui/transportation/MVVSymbol.java
```diff
@@ -0,0 +1,90 @@
...
+import android.graphics.Color;
+
+import com.google.common.base.Optional;
+import com.google.common.primitives.Ints;
+
+/**
+ * Encapsulates information about the symbol shown next to departure information. Contains the
+ * background color and text color, as computed based on the line type and line number.
+ */
+public class MVVSymbol {
```
pfent: Might as well directly use Kotlin?

PR #1104, opened 2018-09-29
Improve offline experience

thellmund:
This pull request improves the offline experience of the app. 
1. Fixes a bug that prevented `MainActivity` from displaying cached content when offline. 
2. Removes `ConnectivityInterceptor` and `NoNetworkConnectionException`, as they prevented a proper use of the Retrofit cache.
3. If cached content is already being displayed, a failed refresh (e.g. due to being offline) now displays a `Snackbar` instead of a full-screen error view. This way, we don’t hide the cached content.


app/src/main/java/de/tum/in/tumcampusapp/component/other/generic/activity/ProgressActivity.kt
```diff
@@ -94,6 +103,41 @@ abstract class ProgressActivity(
         }
     }
 
+    protected fun showErrorSnackbar(t: Throwable) {
+        val messageResId = when (t) {
```
pfent: central error handliong makes a lot of sense! nice

app/src/main/java/de/tum/in/tumcampusapp/component/other/generic/activity/ProgressActivity.kt
```diff
@@ -94,6 +103,41 @@ abstract class ProgressActivity(
...
+            is RequestLimitReachedException -> R.string.error_request_limit_reached
+            else -> R.string.error_unknown
+        }
+
+        showErrorSnackbar(messageResId)
+    }
+
+    protected fun showErrorSnackbar(messageResId: Int) {
+        runOnUiThread {
+            Snackbar.make(contentView, messageResId, Snackbar.LENGTH_LONG)
```
pfent: When there's no internet connection, I would opt for a persistent snackbar ([LENGTH_INDEFINITE](https://developer.android.com/reference/android/support/design/widget/Snackbar#LENGTH_INDEFINITE))

app/src/main/java/de/tum/in/tumcampusapp/component/other/generic/activity/ProgressActivity.kt
```diff
@@ -94,6 +103,41 @@ abstract class ProgressActivity(
...
+                    .setAction(R.string.retry) { retryRequest() }
+                    .setActionTextColor(Color.WHITE)
+                    .show()
+        }
+    }
+
+    protected fun showErrorLayout(throwable: Throwable) {
```
pfent: Snackbars are the superior error display element anyways
Might as well deprecate the error layouts?

pfent: Nevermind, just read your last commit message :tada:

app/src/main/java/de/tum/in/tumcampusapp/component/other/generic/activity/ProgressActivity.kt
thellmund: That would potentially obscure content. Additionally, the `Snackbars` currently can’t be swiped away – we need to introduce `CoordinatorLayouts` for that.

PR #1106, opened 2018-09-30
Register NetworkCallbacks instead of deprecated CONNECTIVITY_ACTION broadcasts

pfent:
[API 28 deprecated](https://developer.android.com/reference/android/net/ConnectivityManager#CONNECTIVITY_ACTION) the Intent we were using.

Instead we should use the NetworkRequest / NetworkCallback API added in API 21.


app/build.gradle
```diff
@@ -124,7 +124,7 @@ dependencies {
 
     // Kotlin
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
-    implementation "org.jetbrains.anko:anko:0.10.5"
+    implementation "org.jetbrains.anko:anko:0.10.6"
```
kordianbruck: Please don't update this here. @thellmund mentioned, that there is a problem with the version in question. We have dependabot, that will create PRs for any changes in that matter.

PR #1107, opened 2018-09-30
Redo StudyRoomsActivity

pfent:
The whole studyrooms component was a bit tangled, so I decided to clean it up in a bit nicer Kotlin code.

Also the layout displaying the occupied / free rooms wasn't too nice:
* Room titles did not include any room number. So there's no way to quickly see, what room this status is for.
* The card layout used a LinearLayout, wasting the right half of the display with empty space.

We now display the room number in addition to the room's name.
Also we now display the room details side by side with the link to the RoomFinder, resulting in a bit more condensed layout

Old:
screenshot_1538338191

New:
screenshot_1538342356


thellmund:
I’m not a fan of the new layout. The button does not look right in this place. Having it below the text labels brings the advantage that the text label can fill the whole width, which is helpful on smaller screens.  

However, I do like the addition of the room number. Maybe we can extract it from the title and show it in a separate line, @pfent?

pfent:
> the text label can fill the whole width

My point is, there are no study rooms, where this is actually the case - even on very small screens.
Currently, we fill those cards to about half which just looks weird to me.
And even when we have massive amounts of details text, I think a line break imposed by the layout constraints wouldn't be too bad.

But let me play around with those cards a little bit more, maybe I can come up with a more satisfying solution…

pfent:
@thellmund how about this?
Now organized in a grid to not waste too much space.
Plus, if we ever have the need for rooms with text that spans the whole horizontal space, we can customize the Recyclerview with a [SpanSizeLookup](https://developer.android.com/reference/android/support/v7/widget/GridLayoutManager.SpanSizeLookup) for multi-column cards.

screenshot_1538395859

PR #1122, opened 2018-10-03
More beta fixes

thellmund:
This pull request fixes some problems encountered in v2.0 beta 3 and 4: 
1. Introduces `EduroamFixCardViewHolder` and `NextLectureCardViewHolder` for greater stability and easier-to-understand code.
2. Updates the design of `NextLectureCard`, as the buttons on the right-hand side were both hard to hit (very small) and interfered with the options button. Now, we display a “show x more ...” button so that the user can expand the card and see up to four upcoming lectures.

---

Collapsed: 
screenshot_20181005-212644

Expanded: 
screenshot_20181005-212650


thellmund:
@kordianbruck @pfent @jacqueline8711 I’m not sure about the design yet. Maybe divider lines would help. What’s your take?

jacqueline8711:
@thellmund Yes, the expanded card looks kind of weird to me with the 'Hide 1 more...' in the middle.
Maybe somewhat like this?

screen shot 2018-10-07 at 09 27 22
screen shot 2018-10-07 at 09 27 32

or with the expand / collapse action at the bottom

screen shot 2018-10-07 at 09 27 43
screen shot 2018-10-07 at 09 27 51

thellmund:
@jacqueline8711 I like the design ideas. Let me know what you think of the current state that I’ve just pushed.

thellmund:
The accidental renaming of "card" to "cardView" in many comments is also reverted.

jacqueline8711:
I think we should keep the link to the Roomfinder in the NextLectureCard, it's quite useful if you need to find your room quickly

thellmund:
I addressed the bug in `NextLectureCard`. 

Regarding the room finder: I’m now directly opening the event fragment within the `CalendarActivity` when tapping on the event in the card, removing one unnecessary step from the flow. I don’t think a direct link to the room finder is necessary: I strongly assume that the assumption that users need the room finder frequently is wrong.

jacqueline8711:
Yes, the roomfinder is probably mostly needed for the first week of the semester. At least I need it a lot then.
Opening the fragment directly is definitely better. It's only working for the first event though, can we make it work for the other ones as well?

.idea/codeStyleSettings.xml
```diff
@@ -1,264 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
jacqueline8711: We probably want to keep the .idea files?

thellmund:
> Yes, the roomfinder is probably mostly needed for the first week of the semester. At least I need it a lot then.
> Opening the fragment directly is definitely better. It's only working for the first event though, can we make it work for the other ones as well?

Fixed.

.idea/codeStyleSettings.xml
thellmund: Yes, Android Studio keeps doing that automatically somehow. Will revert it.

app/src/main/res/layout/layout_lecture_selection.xml
```diff
@@ -6,58 +6,55 @@
...
-            android:layout_width="@dimen/next_lecture_select_width"
-            android:layout_height="@dimen/next_lecture_select_width"
-            android:layout_margin="1dp"
-            android:background="@drawable/next_lecture_select"
-            android:text="4"
-            android:textColor="#ffffffff"
-            tools:ignore="HardcodedText" />
-
-    </LinearLayout>
+    <TextView
```
jacqueline8711: I don't see any usages for this layout, is it still needed?

app/src/main/java/de/tum/in/tumcampusapp/component/ui/studycard/StudyCardListAdapter.java
```diff
@@ -58,7 +58,7 @@ public View getView(int position, View view, ViewGroup parent) {
         }
 
         if (card != null) {
-            // viewHolder.title.setText(card.getTitle());
+            // viewHolder.title.setText(cardView.getTitle());
```
jacqueline8711: This still says cardView

app/src/main/java/de/tum/in/tumcampusapp/component/ui/studycard/CardsDetailActivity.java
```diff
@@ -58,14 +58,14 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
-        /* if (!card.is_valid()) {
+        /* if (!cardView.is_valid()) {
             return null;
         }
         try {
             ChatMember chatMember = Utils.getSetting(this, Const.CHAT_MEMBER, ChatMember.class);
             final TUMCabeVerification v = TUMCabeVerification.create(this, chatMember);
             final Context c = this;
-            return TUMCabeClient.getInstance(c).addStudyCard(card, v);
+            return TUMCabeClient.getInstance(c).addStudyCard(cardView, v);
```
jacqueline8711: This is still supposed to be card

app/src/main/res/layout/layout_lecture_selection.xml
thellmund: Is removed now.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/studycard/CardsDetailActivity.java
thellmund: Anything related to study cards is currently not used in the app, so I did not bother with it. 
@kordianbruck can the study cards stuff be removed?

PR #1135, opened 2018-10-15
Custom event: DatePicker one month off

jacqueline8711:
## Issue
When creating/editing a custom event the datePicker opened to one month in the future (although the date was right) and when selecting the current date the event was created one month the past

Now the date picker is showing the right month and the event is created on the day that the user selected.


thellmund:
The change looks good to me. I think an explanatory commentary would be good to avoid future confusion.

PR #1138, opened 2018-10-16
Fix ActivityNotFoundException in InfoActivity

pfent:
The app crashes on the most recent beta when opening the Facebook link, when the Facebook App is installed with a ActivityNotFoundException.

Some Stackoverflow answers suggest (https://stackoverflow.com/a/24547437, https://stackoverflow.com/a/34564284), the FB App does not accept the URI we're trying to feed it since 2014 :hankey: 
I guess either nobody has the FB App installed anymore, or people simply do not like to click that link :man_shrugging: 

Do we even care? In the SupportCard we just open the browser in any case...

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/ui/overview/InformationActivity.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/ui/overview/InformationActivity.kt
@@ -3,11 +3,12 @@ package de.tum.`in`.tumcampusapp.component.ui.overview
 import android.content.ClipData
 import android.content.ClipboardManager
 import android.content.Intent
-import android.content.pm.PackageManager
+import android.content.Intent.ACTION_VIEW
 import android.content.pm.PackageManager.NameNotFoundException
 import android.net.Uri
 import android.os.Bundle
 import android.preference.PreferenceManager
+import android.support.v4.content.pm.PackageInfoCompat
 import android.text.format.DateUtils
 import android.view.View
 import android.view.ViewGroup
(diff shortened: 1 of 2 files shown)


app/src/main/java/de/tum/in/tumcampusapp/component/ui/overview/InformationActivity.kt
```diff
@@ -55,16 +56,24 @@ class InformationActivity : BaseActivity(R.layout.activity_information) {
...
-        // Open the Facebook app or view page in a browser if Facebook is not installed
+        val default = Intent(Intent.ACTION_VIEW, Uri.parse(getString(R.string.facebook_link)))
         val facebook = try {
             packageManager.getPackageInfo("com.facebook.katana", 0)
             Intent(Intent.ACTION_VIEW, Uri.parse(getString(R.string.facebook_link_app)))
-        } catch (e: PackageManager.NameNotFoundException) {
-            Intent(Intent.ACTION_VIEW, Uri.parse(getString(R.string.facebook_link)))
+        } catch (e: Exception) {
+            default
         }
```
kordianbruck: Assignment seems overly complicated. You can just have `startActivity` inside the try block here and save on that extra logic.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/overview/InformationActivity.kt
```diff
@@ -55,16 +56,24 @@ class InformationActivity : BaseActivity(R.layout.activity_information) {
...
 
-        startActivity(facebook)
+        try {
+            startActivity(facebook)
+        } catch (e: Exception) {
+            // Don't make any assumptions about another app, just start the browser instead
+            startActivity(default)
```
kordianbruck: Just run it in the catch about. No need to separate it into its own catch, right?

Issue #1134, opened 2018-10-13
Crashes if AppMeasurementService is missing


schoersch:
## Suggested Enhancement
first of all thanks for this app.
I'm currently having a android phone with microg as replacement for gms services. Sadly, the campus app crashes on startup, because google's analytics are not available.
If possible I want to help getting this app running without gms by deactivating features or just skipping things like analytics as they are not needed for functionality.
I attached the startup log.
[log10-13-2018_20-41-10.txt](https://github.com/TCA-Team/TumCampusApp/files/2475760/log10-13-2018_20-41-10.txt)

### Environment
* OS version: Android 9.0 (Lineageos 16)
* TUM Campus App version: 1.6.2

kordianbruck:
We do not only use GMS for analytics but also for Messaging. We tried to rip those parts out and publish the app on the f-droid store, but it has proven to be too much effort for very little return.

If you want, feel free to create a branch/fork and see if you can work out something, that does not include maintaining two separate code bases. Analytics removal should be pretty simple, if you just remove the dependencies from the gradle file.

schoersch:
Oke, then I'll try it, although it doesn't look so successful .

'For Messaging' you means the chat, or also other updates for cards or other parts of the app?

kordianbruck:
> Messaging 
Some Notifications + Chat

pfent:
@schoersch I think you can simply comment the crashlytics line in `app/build.gradle`, like so:
```gradle
//releaseImplementation 'com.crashlytics.sdk.android:crashlytics:2.9.5'
```

A rebuild, then install on your device *should* work, it would be awesome if you could test that.
I think MicroG includes the messaging parts, but unfortunately does not work with crashreporting...

If you have any troubles building / testing the App, feel free to ask any questions right here :)

schoersch:
@pfent  I'm just failing to build the App with intellij idea, getting the error message below. Did I make an error during importing?

```
Circular dependency between the following tasks:
:app:checkManifestChangesDebug
\--- :app:instantRunMainApkResourcesDebug
     \--- :app:transformClassesAndDexWithShrinkResForDebug
          \--- :app:transformDexArchiveWithDexMergerForDebug
               +--- :app:preColdswapDebug
               |    \--- :app:incrementalDebugTasks
               |         +--- :app:transformClassesAndClassesEnhancedWithInstantReloadDexForDebug
               |         |    \--- :app:transformClassesWithInstantRunForDebug
[...]

PR #1143, opened 2018-10-17
InstantRun doesn't play nicely with shrinkResources.

pfent:
Disable it for now
This fixes the Circular dependency issue discussed in #1134


kordianbruck:
I got to agree with till here. Its easy to disable instant run in the settings and its not much of an inconvenience for now.

PR #1164, opened 2018-10-24
Improve display of upcoming events via CalendarDetailsFragment

thellmund:
Instead of heading to CalendarActivity and displaying the CalendarDetailsFragment there, we display it directly in MainActivity. This way, the user can go back to browsing their feed more easily. 

The fragment now also contains a "Show in calendar" button, so that the user can open the CalendarActivity and perform actions on the event (such as edit and delete).


app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarDetailsFragment.kt
```diff
@@ -93,22 +98,39 @@ class CalendarDetailsFragment : RoundedBottomSheetDialogFragment() {
             descriptionTextView.text = calendarItem.description
         }
 
-        if (calendarItem.url.isEmpty()) {
+        if (!isShownInCalendarActivity) {
+            showInCalendarButtonContainer.visibility = View.VISIBLE
+            showInCalendarButton.setOnClickListener { openEventInCalendarActivity(calendarItem) }
+        }
+
+        if (calendarItem.isEditable && !isShownInCalendarActivity) {
```
jacqueline8711: This is the wrong way around, you're showing the edit buttons in the MainActivity (not working there) and not in the CalendarActivity. Please change.

thellmund: Oopsie ... fixed.

thellmund:
Addressed the code review feedback. Will merge this now, so that we have as few pull requests as possible open when we merge #1173.

PR #1165, opened 2018-10-24
Fix Exception caused by too many alarms

### LINKED ISSUE #1162: SecurityException: Too many alarms registered
@thellmund already tried to fix this issue with f5037b67eae1477da37c5656c523f060798a6a19, but that apparently didn't fix the issue.

I guess we need to keep track of the "active" notifications by manually counting how many are active. I.e. increment in `scheduleAlarm()` and decrement in `NotificationAlarmReceiver.onReceive()`

pfent:
this fixes #1162

We now persistently track the active alarms in the DB.
@thellmund voiced concerns about important Notifications, so I decided to not limit the creation of *single* notification alarms, but only when they are added in batch.

Also the Calendar and Transport notifications now only use up up to half of the remaining "space" for notifications, so there should always be space for the most important notifications.

### DIFF
--- a/app/schemas/de.tum.in.tumcampusapp.database.TcaDb/2.json
+++ b/app/schemas/de.tum.in.tumcampusapp.database.TcaDb/2.json
@@ -1,0 +1,1234 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 2,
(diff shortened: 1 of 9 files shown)


thellmund:
I can’t see any problems in this PR. Maybe wait until @jacqueline8711 can take a look, as she always finds the problems that I overlook ;)

PR #1173, opened 2018-10-28
Use AndroidX instead of old SupportLibrary

pfent:
Google recently refactored the android support libraries to the AndroidX
namespace with different, per artifact versioning.
see: https://android-developers.googleblog.com/2018/05/hello-world-androidx.html
and https://developer.android.com/jetpack/androidx/

This commit gets rid of `android.support.*` and uses `androidx.*` instead.
Also since PhotoView recently moved to AndroidX, it now builds unlike in #1108


kordianbruck:
@thellmund thanks for checking this PR out, but why exactly did we merge `alpha01` libraries into the stable master? I'm hesitant to release a beta with those libraries in place. Any suggestions? You think its good to release?

pfent:
@kordianbruck just release a `alpha01` instead of a beta 🙈

PR #1183, opened 2018-10-31
Refactor FeedbackActivity

### LINKED ISSUE #1160: Stop sending Feedback Images if Feedback could not be created
`sendFeedback` in the `FeedbackActivity` currently does not abort sending the Images, if the initial request to send the feedback itself fails. Please refactor this, to first let the feedback to be created and just later on, it will create the images if the first request was successful. 

Also you are using the same callback for different requests. This makes it harder to understand, what the intention of the programmer was and requires me to take a deeper look into the code to understand what is going on. 

So, heres a checklist, I would love to see a similar commit structure:
- [x] Create a new helper class maybe, that only deals with business logic = controller
- [x] Pull code from the FeedbackActivity (=View) regarding sending the feedback and put it into the controller.
- [x] Improve the code to resolve the above comments and simply pass a model between controller and view

[I know we are not beeing super correct here, but Activities should rather not contain any logic]

jacqueline8711:
## Issue
fixes #1160

- divides code that was completely in FeedbackActivity into three classes (FeedbackActivity, FeedbackController and ImageUtils)
- separates sending the images from sending the main content of the feedback
- also removes divider from the feedback layout between feedback message and the images 

screenshot_20181128-101035


app/src/main/java/de/tum/in/tumcampusapp/component/tumui/feedback/FeedbackController.java
```diff
@@ -0,0 +1,332 @@
...
+    }
+
+    String getCurrentPhotoPath() {
+        return mCurrentPhotoPath;
+    }
+
+    void startTakingPicture(Activity activity) {
+        Intent takePictureIntent = new Intent(MediaStore.ACTION_IMAGE_CAPTURE);
+        // Ensure that there's a camera activity to handle the intent
+        if (takePictureIntent.resolveActivity(mContext.getPackageManager()) != null) {
```
pfent: Please do an early return instead of a deeper nesting level

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/feedback/FeedbackController.java
```diff
@@ -0,0 +1,332 @@
...
+            File photoFile = null;
+            try {
+                photoFile = ImageUtils.createImageFile(mContext);
+                // Save a file: path for use with ACTION_VIEW intents
+                mCurrentPhotoPath = photoFile.getAbsolutePath();
+            } catch (IOException e) {
+                Utils.log(e);
+            }
+            // Continue only if the File was successfully created
+            if (photoFile != null) {
```
pfent: same here: early return, keep the nesting level low

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/feedback/FeedbackController.java
```diff
@@ -0,0 +1,332 @@
...
+            return;
+        }
+
+        showProgressBarDialog();
+        TUMCabeClient client = TUMCabeClient.getInstance(mContext);
+        client.sendFeedback(feedback, new Callback<Success>() {
+            @Override
+            public void onResponse(@NonNull Call<Success> call, @NonNull Response<Success> response) {
+                Success success = response.body();
+                if (success != null && success.wasSuccessfullySent()) {
```
pfent: Please keep the error handling consistent, I like how you did it in `sendFeedbackImages()`, but two different ways to handle it is confusing

PR #1188, opened 2018-10-31
Use WorkManager instead of JobIntentServices

### LINKED ISSUE #914: Use architecture component's WorkManager
Instead of using AlarmManager, we should use [WorkManager](https://developer.android.com/topic/libraries/architecture/workmanager) to schedule tasks.

The current system relies on `BOOT_COMPLETED` and `WIFI_STATE_CHANGED` broadcasts to schedule services, which aren't really reliable on newer versions of Android.
Instead of working around those changes, we should rely on [the robust system Workmanager implements](https://developer.android.com/reference/androidx/work/WorkManager).  This probably also means to change the Services to `OneTimeWorkRequest` or `PeriodicWorkRequest`.

@thellmund maybe this also allows for easier Notification scheduling in #879?

pfent:
this fixes #914

WorkManager has much more robust scheduling strategies, with automatic retry strategies built-in. 
I also noticed, that the Wifi measurements were collected, but never uploaded, so I removed that logic altogether.

What do you think?
Did I miss any "background work", which we do with threads / background loopers and should be done with WorkManager?


app/src/main/java/de/tum/in/tumcampusapp/component/other/wifimeasurement/WifiMeasurementManager.java
```diff
@@ -1,94 +0,0 @@
...
-import de.tum.in.tumcampusapp.database.TcaDb;
-import de.tum.in.tumcampusapp.utils.Utils;
-import retrofit2.Call;
-import retrofit2.Callback;
-import retrofit2.Response;
-
-public class WifiMeasurementManager {
-    public static final String WIFI_SCAN_MINIMUM_BATTERY_LEVEL = "WIFI_SCAN_MINIMUM_BATTERY_LEVEL";
-
-    private final WifiMeasurementDao dao;
```
kordianbruck: What about the `WifiMeasurementDao` and `WifiMeasurement` those should probably also be removed?

thellmund:
@kordianbruck Notification scheduling is done via `AlarmManager` and should not impacted by this change. 

However, I’m a little surprised to see `WorkManager` used in `StartupActivity`. Isn’t it supposed to be used for deferrable tasks? This seems very non-deferrable to me.

kordianbruck:
@thellmund the scheduling for when the notification should plop yes, but these alarms are setup in the background service/worker whenever the data is updated. So question is: are we fine with those simply stopping to work, whenever the workmanager does not feel like doing stuff.

thellmund:
@kordianbruck After a quick glance at `BackgroundService`: it seems to use a `PeriodicWorkRequest`, which guarantees that it runs at least once in the specified timeframe. Thus, I can’t imagine that notifications will be impacted in any way.

app/src/main/java/de/tum/in/tumcampusapp/component/other/wifimeasurement/WifiMeasurementManager.java
pfent: Good catch!
We probably also need a mirgation to drop that table

pfent:
@thellmund The BackgroundWorker now specifies constraints for the battery to not be low, so it is not necessarely run.
The BackgroundWorker schedules the DownloadWorker, which in turn uses all the different Managers to fetch new data. New data would mean new Notifications to be scheduled.

My reasoning was, that on low battery, we should probably not download data and schedule new alarms, but only use existing data.

>are we fine with those simply stopping to work

@kordianbruck *new* notifications won't be scheduled. We have the very same problem already, when there is no network connectivity and no data to be downloaded. IMHO no biggie, since we schedule some alarms ahead of time (probably still too much, see #1162)

thellmund:
@pfent I don’t understand why `DownloadWorker` is even needed. As far as I can see, it’s only used in two places: `StartupActivity` and `BackgroundWorker`.

In `StartupActivity`, we should not be using a `Worker`, since we’re performing non-deferrable work. `Workers` are for deferrable work.

In `BackgroundWorker`, we can call the download functionality in a different way (maybe in an `Executor`?) instead of launching yet another `Worker`.

Especially the latter would remove any uncertainty regarding notification scheduling: The `BackgroundWorker` is a periodic work request and therefore guaranteed to run within the specific time frame.

thellmund:
The documentation states “The PeriodicWorkRequest is guaranteed to run exactly one time during this interval”

pfent:
And the documentation also states:

> WorkManager is a library used to enqueue work that is guaranteed to execute *after* its constraints are met

thellmund:
Ambiguous documentation 🤷‍♂️

pfent:
@kordianbruck I handled the DB changes, take a look

@thellmund I unbundled the DownloadWorker some more, so it's not used in `StatupActivity` and `ActivityForDownloadingExternal` anymore.
However, I still like the separation of one Worker to download stuff and another Worker, that's regularly scheduled. IMO all other ways of running the download actions only have drawbacks...

kordianbruck:
anyone volunteering to update this PR? :checkered_flag:

pfent:
@kordianbruck I'd actually wait a little bit more with this PR, workmanager is currently on version `1.0.0-beta02`

should be usable in production is a couple of weeks

thellmund:
I’ve spent some time getting rid of the merge conflicts and will push a commit tomorrow.

Issue #1180, opened 2018-10-29
tu-film instead of tu-kino


kordianbruck:
![screenshot_20181029-175445](https://user-images.githubusercontent.com/298860/47666524-09c7ab00-dba4-11e8-97ed-eb5b22654047.png)

pfent:
fixed in #1201

PR #1203, opened 2018-11-05
Fixing case for 'TU Film' string.

rbaid:
Making sure German and English has same translation.

## Issue
[Case 1180](https://github.com/TUM-Dev/Campus-Android/issues/1180)

Correct casing and makes translation same for TU Film.


pfent:
This fixes  #1180

Issue #1196, opened 2018-11-01
NPE first time user tries to pay


kordianbruck:
```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'void com.stripe.android.PaymentSession.presentPaymentMethodSelection()' on a null object reference
       at de.tum.in.tumcampusapp.component.ui.ticket.activity.StripePaymentActivity.lambda$initViews$0(StripePaymentActivity.java:114)
       at de.tum.in.tumcampusapp.component.ui.ticket.activity.-$$Lambda$StripePaymentActivity$hvnVaRNbsCaa8e4wY_nB-RKjBjU.onClick(Unknown Source:2)
       at android.view.View.performClick(View.java:6891)
       at android.view.View$PerformClick.run(View.java:26083)
       at android.os.Handler.handleCallback(Handler.java:789)
       at android.os.Handler.dispatchMessage(Handler.java:98)
       at android.os.Looper.loop(Looper.java:164)
       at android.app.ActivityThread.main(ActivityThread.java:6938)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.Zygote$MethodAndArgsCaller.run(Zygote.java:327)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1374)
```

https://console.firebase.google.com/u/0/project/tca-backend-0001/crashlytics/app/android:de.tum.in.tumcampus/issues/5bdb63bdf8b88c29632b80c9

kordianbruck:
Alright, the logic for locking the event 4 hours before start were wrong. In addition for the missing error checks on the client side, this resulted in a broken state. This should no longer happen now with the latest merged PR #1209

PR #1204, opened 2018-11-07
Prevent NPE in StripePaymentActivity

jacqueline8711:
## Issue
This fixes the following issue(s): #1196

NPE occurred when the user tried to select the payment method before loading was finished.


kordianbruck:
Are you sure this is the root of the problem? In my case I really had to go back again and only once entering the ticket type for the second time it works.

kordianbruck:
Nevermind, I'll test it locally again :+1:

jacqueline8711:
@kordianbruck This worked for me, I reproduced the exact same error the way that I described above and this fixed it

Issue #1195, opened 2018-11-01
TicketSale: Terms of Service


kordianbruck:
We need to add a link to the terms of service, that the user has to check a checkbox before clicking on buy in the PaymentActivity.

For the TU-Film this is: https://www.tu-film.de/pages/view/nutzungsbedingungen

You can get this URL and the stripe public key from this new API endpoint: https://app.tum.de/Api/event/ticket/payment/1/

Goals: 
- [ ] Add checkbox with "I agree to the following conditions" to payment activity
- [ ] Use the stripe public key from the API instead of hard coding it

PR #1205, opened 2018-11-07
Adds terms of service checkbox and link

jacqueline8711:
## Issue

This fixes the following issue(s): #1195 

## Screenshot
coming later

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/activity/StripePaymentActivity.java
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/activity/StripePaymentActivity.java
@@ -3,11 +3,6 @@ package de.tum.in.tumcampusapp.component.ui.ticket.activity;
 import android.content.Intent;
 import android.os.Build;
 import android.os.Bundle;
-import androidx.annotation.NonNull;
-import androidx.annotation.Nullable;
-import androidx.annotation.RequiresApi;
-import com.google.android.material.button.MaterialButton;
-import androidx.appcompat.app.AlertDialog;
 import android.text.Editable;
 import android.text.TextWatcher;
 import android.transition.TransitionManager;
@@ -18,6 +13,7 @@ import android.widget.FrameLayout;
 import android.widget.TextView;
 import android.widget.ViewSwitcher;
 
+import com.google.android.material.button.MaterialButton;
 import com.stripe.android.CustomerSession;
 import com.stripe.android.PaymentConfiguration;
 import com.stripe.android.PaymentSession;
@@ -30,6 +26,10 @@ import com.stripe.android.view.PaymentMethodsActivity;
 import java.util.ArrayList;
 import java.util.List;
 
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.annotation.RequiresApi;
+import androidx.appcompat.app.AlertDialog;
 import de.tum.in.tumcampusapp.R;
 import de.tum.in.tumcampusapp.api.app.TUMCabeClient;
 import de.tum.in.tumcampusapp.api.app.exception.NoPrivateKey;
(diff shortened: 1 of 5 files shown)


jacqueline8711:
@kordianbruck I didn't delete the TicketType table since it is used in the ShowTicketActivity which as far as I know is supposed to work offline.

kordianbruck:
@jacqueline8711 makes sense :+1:

jacqueline8711:
@kordianbruck Yes, they should always be up to date the way it is now. They are replaced whenever we download them.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/model/TicketType.kt
```diff
@@ -20,7 +21,10 @@ data class TicketType(
         @SerializedName("ticket_type")
         var id: Int = 0,
         var price: Int = 0,
-        var description: String = ""
+        var description: String = "",
+        @Ignore
+        @SerializedName("payment")
+        var paymentInfo: Payment = Payment()
```
kordianbruck: This leads to a database change no? Missing the migrations + schema.json then...

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/activity/BuyTicketActivity.java
```diff
@@ -204,6 +204,8 @@ private void handleTicketReservationSuccess(TicketType ticketType,
         Intent intent = new Intent(this, StripePaymentActivity.class);
         intent.putExtra(Const.KEY_TICKET_PRICE, ticketType.getFormattedPrice());
         intent.putExtra(Const.KEY_TICKET_HISTORY, response.getTicketHistory());
+        intent.putExtra(Const.KEY_TERMS_LINK, ticketType.getPaymentInfo().getTermsLink());
+        intent.putExtra(Const.KEY_STRIPE_API_PUBLISHABLE_KEY, ticketType.getPaymentInfo().getStripePublicKey());
```
kordianbruck: I mean it should not happen, but what happens if paymentInfo is `null`

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/activity/StripePaymentActivity.java
```diff
@@ -68,8 +73,13 @@ public void onCreate(Bundle savedInstanceState) {
 
         ticketPrice = getIntent().getStringExtra(Const.KEY_TICKET_PRICE);
         ticketHistory = getIntent().getIntExtra(Const.KEY_TICKET_HISTORY, -1);
+        termsOfServiceLink = getIntent().getStringExtra(Const.KEY_TERMS_LINK);
+        stripePublishableKey = getIntent().getStringExtra(Const.KEY_STRIPE_API_PUBLISHABLE_KEY);
```
kordianbruck: There needs to be some handling if the key is null. Either here or above. The payment activity should not proceed, if its empty.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/activity/BuyTicketActivity.java
pfent: @kordianbruck paymentInfo is a not-null type in Kotlin.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/model/TicketType.kt
jacqueline8711: No, paymentInfo is not saved in the database. I don't think it's necessary since we download it whenever the user wants to buy a ticket and we don't need it later.
```@Ignore``` makes Room ignore it (doesn't change the schema and doesn't try to insert it)

Issue #1207, opened 2018-11-07
Ticket Sale: hide past events


kordianbruck:
On the overview, we no longer want to see events, that have passed already.

PR #1210, opened 2018-11-09
TicketSale: hide past events

jacqueline8711:
## Issue

1) This fixes the following issue(s): #1207 
Only showing events that start in the future (tickets should not be affected)

2) Also prevents events from being deleted if they have not passed yet (might still be needed for the ticket), might have been a problem for offline users


kordianbruck:
Follow up: what happens, if the event is deleted from the local table but the ticket still references it? I mean currently `removePastEvents` does nothing, because all the events are always returned from the server and it instantly gets replaced, but we have not yet thought about, if we somehow need to special case this and return bought events in addition to just returning future events on the main end point?

@jacqueline8711 thoughts? can you maybe come up with a suggestion? Or lets discuss this the next time we meet.

Issue #1199, opened 2018-11-02
TicketSale: View improvements


kordianbruck:
So now that we have a few ticktes in the view, there are a few improvements, that seem kind of obvious and that would be great to implement:
- [ ] Images can also be vertical, in a 9:16 format. The overview could somehow limit the height and shrink those, but keep the 16:9 overall size in terms of height.
- [ ] On the details page, the image is kind of big right now, maybe show 70% but make it scrollable if the user wants to see the full size image
- [x] On the start page we now have the events and tu film right next to each other. Best option would be, to display the ticket if there is one, otherwise just display the news item
- [x] It would be great, if we could link from the event to the tu film - let me know, if you need a reference from the backend for this
- [x] Is the address necessary on the overview?
- [x] Make the banner part of the scrollable view in the overview

Thanks Till for taking up this task - cause I really suck at UI :hankey:

jacqueline8711:
- If the image is vertical I would say it would look best if we show it like the tu film cards with the image on the side (those events are mostly movies fow now anyway)
- What do you think of merging the tu film view with the ticket view if the event is a movie (we would probably need a field for the event that contains the movie id) and show the view both when going to the view from the news and the events?
Mock:
![moviecard](https://user-images.githubusercontent.com/16260112/48134135-55abdb80-e299-11e8-9a07-ae9507e9fb12.png)
![screenshot_20181107-135041](https://user-images.githubusercontent.com/16260112/48134172-7116e680-e299-11e8-84dc-c8fb6abe125f.png)

pfent:
Considering the details image:
IMHO the image is actually kind of small right now, looking at the screenshot the text on the image is hardly readable

kordianbruck:
@jacqueline8711 I'm with you on all accounts! It looks great! We could of course just integrate the events card for tufilm into their news card? Whatever is easy - I like the hint, that one can get ticktest :+1:

jacqueline8711:
@kordianbruck could you add a reference in the event to the id of the tu film?

kordianbruck:
@jacqueline8711 sure, the event list now has the following additional fields:
```
"news": "513304",
"kino": "113",
```

https://app.tum.de/Api/event/list/

kordianbruck:
@jacqueline8711 feel free to reopen, in case you want to finish the other points above.

jacqueline8711:
@kordianbruck yes, I'm working on some of the other points now, I just didn't want to put everything in one PR

[...]

PR #1222, opened 2018-11-21
Display show ticket button in kinoActivity

jacqueline8711:
## Issue
This fixes the following issue(s): #1199 

The KinoDetailsFragment was only showing 'Buy Ticket' but not 'Show Ticket' if the user already has a ticket


kordianbruck:
LGTM :clinking_glasses:

PR #1226, opened 2018-11-26
Tweak design in kino details view

thellmund:
This PR tweaks the design of the kino details view, as shown below.

Main changes: 
- Use shorter date format
- Show “x / 10” for rating
- Improve formatting of the description to prevent full stops without a subsequent space

Old
screenshot_20181126-193902

New
screenshot_20181126-194026


thellmund:
@jacqueline8711 It’s based on the device locale, so it should display the correct format for each user.

kordianbruck:
I'd suggest using just 28. Nov, as its pretty clear which year is meant. Thoughts?

thellmund:
@kordianbruck Sure, let’s go with that. Will push changes in a bit.

thellmund:
Think it looks fine now: 
screenshot_20181129-222201

PR #1229, opened 2018-11-29
Fix crashes in v3.0

thellmund:
This pull request fixes the crashes that affect the most users in version 3.0:
1. `NetworkOnMainThreadException` in `BarrierFreeFacilitiesActivity`(fixed by moving the network call to a background thread)
2. Situations where view was null in `StudyRoomsActivity` (fixed by making it nullable; it’s not used anyway)
3. Crash when `CalendarController.QueryLocationsService` could not be started (convert `IntentService` into `JobIntentService`, which don’t experience the same crashes)

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/other/locations/LocationManager.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/other/locations/LocationManager.kt
@@ -20,6 +20,7 @@ import de.tum.`in`.tumcampusapp.database.TcaDb
 import de.tum.`in`.tumcampusapp.utils.Const
 import de.tum.`in`.tumcampusapp.utils.Utils
 import de.tum.`in`.tumcampusapp.utils.tryOrNull
+import org.jetbrains.anko.doAsync
 import java.io.IOException
 import java.lang.Double.parseDouble
 import java.util.*
@@ -236,7 +237,9 @@ class LocationManager(c: Context) {
      * @return the id of current building
      */
     fun fetchBuildingIDFromCurrentLocation(callback: (String?) -> Unit) {
-        fetchBuildingIDFromLocation(getCurrentOrNextLocation(), callback)
+        doAsync {
+            fetchBuildingIDFromLocation(getCurrentOrNextLocation(), callback)
+        }
     }
 
     /**
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/other/locations/LocationManager.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/other/locations/LocationManager.kt
@@ -19,6 +19,7 @@ import de.tum.`in`.tumcampusapp.component.ui.transportation.model.efa.StationRes
 import de.tum.`in`.tumcampusapp.database.TcaDb
 import de.tum.`in`.tumcampusapp.utils.Const
 import de.tum.`in`.tumcampusapp.utils.Utils
+import de.tum.`in`.tumcampusapp.utils.tryOrNull
 import java.io.IOException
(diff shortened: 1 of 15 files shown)


app/src/main/java/de/tum/in/tumcampusapp/service/QueryLocationsService.kt
```diff
@@ -0,0 +1,76 @@
...
+    private fun loadGeo() {
+        val locationManager = LocationManager(this)
+        val calendarDao = TcaDb.getInstance(this).calendarDao()
+        val roomLocationsDao = TcaDb.getInstance(this).roomLocationsDao()
+
+        val calendarItems = calendarDao.lecturesWithoutCoordinates
+        for (calendarItem in calendarItems) {
+            val location = calendarItem.location
+            if (location.isEmpty()) {
+                continue
```
pfent: wanna do a `filter` for this?

app/src/main/java/de/tum/in/tumcampusapp/service/QueryLocationsService.kt
```diff
@@ -0,0 +1,76 @@
...
+            }
+
+            val geo = locationManager.roomLocationStringToGeo(location)
+            geo?.let {
```
pfent: `mapNotNull`?

app/src/main/java/de/tum/in/tumcampusapp/component/other/locations/LocationManager.kt
```diff
@@ -215,33 +217,30 @@ class LocationManager(c: Context) {
...
-                    for (map in result) {
-                        buildingToGpsDao.insert(map)
-                    }
-                } catch (e: IOException) {
-                    Utils.log(e)
-                    return ArrayList()
-                }
+    private fun fetchBuildingsToGps(): List<BuildingToGps> {
+        val results = buildingToGpsDao.all.orEmpty()
+        if (results.isEmpty()) {
```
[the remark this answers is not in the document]
thellmund: Take a look at the new version and let me know what you think

app/src/main/java/de/tum/in/tumcampusapp/service/QueryLocationsService.kt
thellmund: Addressed.

app/src/main/java/de/tum/in/tumcampusapp/service/QueryLocationsService.kt
thellmund: Addressed.

app/src/main/java/de/tum/in/tumcampusapp/service/QueryLocationsService.kt
```diff
@@ -0,0 +1,81 @@
...
+import de.tum.`in`.tumcampusapp.component.tumui.calendar.CalendarController
+import de.tum.`in`.tumcampusapp.component.tumui.calendar.model.CalendarItem
+import de.tum.`in`.tumcampusapp.component.tumui.lectures.model.RoomLocations
+import de.tum.`in`.tumcampusapp.database.TcaDb
+import de.tum.`in`.tumcampusapp.utils.Const
+import de.tum.`in`.tumcampusapp.utils.Utils
+import de.tum.`in`.tumcampusapp.utils.sync.SyncManager
+import org.jetbrains.anko.doAsync
+
+class QueryLocationsService : JobIntentService() {
```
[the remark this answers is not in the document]
thellmund: We could split it into two distinct services and invoke the calendar-syncing service once the room locations are loaded. But I’d keep that for another PR.

Issue #1115, opened 2018-10-02
"Whats New" Card


kordianbruck:
After updates of the app, we could display a "Whats New" card that explains new features to the user.

Suggestion from: https://www.support.tum.de/otrs/index.pl?Action=AgentTicketZoom;TicketID=2910099

PR #1230, opened 2018-11-30
Adds card that displays an update note

jacqueline8711:
## Issue
This fixes the following issue(s): #1115 

Update message is downloaded from the server (could be hardcoded, but this way we can change it later if we forgot something and it's not a lot of data)
Card reappears at every version change if there is a note. 

## Screenshot
Layout very simple for now (message a simple TextView)
updatenotecard

Students can easily discover new features


pfent:
Looking good :+1: 
Backend should also be ready -> let's merge

PR #1231, opened 2018-12-02
Refactor CafeteriaActivity and view model

thellmund:
This PR refactors `CafeteriaActivity` and `CafeteriaViewModel` to be more reactive and extensible. It contains the following changes: 
1. Use `LiveData` to react to database queries and user interaction.
2. Move download of cafeteria information to `CafeteriaLocalRepository`, so that we don’t have to use a ViewModel in `DownloadService.`


kordianbruck:
Cool, very good refactoring!

PR #1232, opened 2018-12-02
Refactor view models

thellmund:
This pull request follows #1231 and refactors some of our view models. Mainly, it’s about distributing responsibilities to local and remote repositories. The benefits: 
- We no longer call view models, which are intended to be used with lifecycle-bound views, in `DownloadService`.
- `Disposable` objects in our view models are now properly disposed.
- We can reuse the repositories more easily across view models, e.g. use EventsRemoteRepository in both the events and the TU Film screen when loading the amount of available tickets.


kordianbruck:
@thellmund unfortunately this pr was overlapping "a bit" with your other changes =/
You want me to take a look already or do you first want to rebase those changes?

PR #1234, opened 2018-12-04
Introduce Dagger in some components

thellmund:
With this pull request, I’m introducing dependency injection via Dagger in the `Cafeteria` component, `Kino` component and in the `DownloadService`. 

For injecting `DownloadService`, I’ve made some changes to how we store `NewsAlerts`. Instead of using `Utils.setSetting(Context, NewsAlert)`, I have created a `TopNewsStore` that can be injected into `DownloadService` This architecture will allow for easier testing in the future without relying on the Android framework, e.g. by using a `FakeTopNewsStore`.


pfent:
Cool :+1:  slow and stable progress.

IMO documentation on `TopNewsStore` is even a little bit over the top, but I like the spirit :wink:.
I feel like `AppComponent` would deserve some explanation as it ties together everything else.
I think it would be handy for new contributors to have a small cheat sheet what each subcomponent needs, so they don't copy just blindly from other components.
But that's not the focus of this PR, maybe just something for the future...

thellmund:
@pfent @kordianbruck We could add some information about the Dagger components to the wiki. I think that’s a better place for long-form explanations than the code. What do you think?

Issue #1227, opened 2018-11-27
Allow Sale of multiple tickets


kordianbruck:
We want to enable users to not only buy one ticket, but let them buy a variable amount up to X. (X is served with the ticket type, inside the ticket payment, being the max tickets of this type that the user is allowed to buy)

jacqueline8711:
How will the ticket check work if one user has multiple tickets? Do we already have any plans for that?

kordianbruck:
@jacqueline8711 the checkin dialog should show the number of tickets for that ticket, that is being redeemed. For now a ticket should still only be redeemed once.

PR #1236, opened 2018-12-12
Let users buy multiple tickets

jacqueline8711:
## Issue
This fixes the following issue(s): #1227 

- Let the user choose any amount of tickets between minTickets and min(maxTickets,availableTickets) of multiple ticket types
- user can buy tickets once and not buy two tickets today and then another one tomorrow
- tickets can only be redeemed together

## Screenshot
screenshot_20190206-112154

One person can buy the tickets for a whole group (less annoying)


kordianbruck:
> @kordianbruck Do you think this would be too restrictive?

I think this should be fine for now, we might want to expand it further in the future.

kordianbruck:
Still need to merge the backend changes first - please wait.

Issue #1172, opened 2018-10-26
Calendar: too much spacing


kordianbruck:
We are getting complaints, that the calendar is not the best, as with many events, nothing gets displayed. Could we maybe reduce the spacing inside? Reduce font size?
The Dayview seems okaish. The Week view is really bad. Other ideas?

![screenshot_2018-10-26-10-41-48](https://user-images.githubusercontent.com/298860/47555760-ec2ce400-d90c-11e8-8dce-7eb4ced4f663.jpg)
![screenshot_2018-10-26-10-40-50](https://user-images.githubusercontent.com/298860/47555761-ecc57a80-d90c-11e8-8226-1c1d93191d47.jpg)

pfent:
I just think you're not supposed to have 4 lectures at the same time :man_shrugging: 

I think smaller font sizes won't really improve on this, since the text is already kinda hard to read. The only thing a can come up with, would be some kind of abbreviation system, but that's really hard for the general case, especially for german titles...
Maybe run something to split up compound words (like [this](https://github.com/danielnaber/jwordsplitter)) then build an abbreviation of the first letters of those components?

Otherwise we can also provide a horizontal version of the week view, so people could rotate their phone...

kordianbruck:
It works in horizontal viewmode, no problemo. On the very left there are no letters displayed at all. I know that 4 events is not normal, but it was working better with our previous calendar version. (At least the user says so)

So lets sum up, what things we can try: 
(@jacqueline8711 is an expert in calendars & finding those extra pixels :rofl: )

- [ ] No bold text
- [ ] Less spacing between events
- [ ] Smaller text size
- [ ] third display mode: next 3 days
- [ ] Text in vertical mode, so maybe its more readable? (Rotate 90°)
- [ ] ???

Please post some things that you try from that list

thellmund:
I think that’s a problem in the framework. Since we're using my fork since a few weeks, I can take a look at horizontal spacing. 

Also, if you have a such a packed schedule, week view is just not for you.

kordianbruck:
> Kommentar zu neuem Design: Die Lücke zwischen Aktivitäten im Kalender ist unpraktisch erhöht. 

Its not something users are imaging. Previously the spacing between the items was less, as I've read on multiple accounts now. Saying that the users schedule is too packed for the view is easy to say. I have a similarly packed schedule in Google Calendar and it is still usable.

jacqueline8711:
[...]

PR #1239, opened 2018-12-22
Update calendar view

thellmund:
This pull request updates the calendar view library to the latest version and makes some changes in our own code due to the library’s updated API. 

The pull request also contains some UI updates, mainly adding a vertical and horizontal header line in the calendar view, resulting in a cleaner design.


app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/WidgetCalendarItem.kt
```diff
@@ -0,0 +1,53 @@
...
+import com.alamkanak.weekview.WeekViewEvent
+import de.tum.`in`.tumcampusapp.component.tumui.calendar.model.CalendarItem
+import de.tum.`in`.tumcampusapp.component.tumui.roomfinder.model.RoomFinderSchedule
+import de.tum.`in`.tumcampusapp.utils.DateTimeUtils
+import org.joda.time.DateTime
+
+/**
+ * A class to represent events for the integrated WeekView calendar
+ */
+class WidgetCalendarItem private constructor(
```
pfent: Would it make sense for this to be a data class?
It only implemets WeekViewDisplayable...

pfent:
> The pull request also contains some UI updates, mainly adding a vertical and horizontal header line in the calendar view, resulting in a cleaner design.

I'd like to check this out, but I'm currently not at my development machine :christmas_tree: 
Could you provide a screenshot, then we can maybe also convince @kordianbruck, that #1172 is resolved?

jacqueline8711:
@pfent 
screenshot_calendar

Issue #688, opened 2018-02-21
Notification for new Grades


kordianbruck:
Previously we had this in the background service:
```
/**
     * This method should fetch the grade in order to get updates grades. It is not implemented,
     *  since the grade web service is under change and redevelops the grade interface.
     */
    private void fetchGrades() {
        // fetching xml from tum online
		TUMOnlineRequest requestHandler = new TUMOnlineRequest(TUMOnlineConst.EXAMS,
                                                               getApplicationContext());
		String rawResponse = requestHandler.fetch();
		Serializer serializer = new Persister();
		ExamList examList = null;
		// Deserializes XML response
		try {
			examList = serializer.read(ExamList.class, rawResponse);
		} catch (Exception e) {
			Utils.log(e);
		}
		// generating notification
		generateNotification(examList);
	}

	private void generateNotification(ExamList examList) {

		SharedPreferences settings = PreferenceManager
				.getDefaultSharedPreferences(this);
		int gradeCount = settings.getInt(Const.Grade_Count, 0);
		// Log.d("Grade Count", "" + gradeCount);
		int newSize = examList.getExams().size();
		if (gradeCount != 0) {
			if (newSize > gradeCount) {
				Editor editor = settings.edit();
				editor.putInt(Const.Grade_Count, newSize);
				editor.commit();
				int icon = R.drawable.ic_notification;
				// Generating GCMNotification
				NotificationCompat.Builder mBuilder = new NotificationCompat.Builder(
						this)
						.setSmallIcon(icon)
						.setContentTitle(getString(R.string.notification_title))
						.setContentText(
								getString(R.string.notification_content));
				// Creates an explicit intent for an Activity in your app
				Intent resultIntent = new Intent(this, GradesActivity.class);

				// The stack builder object will contain an artificial back
				// stack for the
				// started Activity.
				// This ensures that navigating backward from the Activity leads
				// out of
				// your application to the Home screen.
				TaskStackBuilder stackBuilder = TaskStackBuilder.create(this);
				// Adds the back stack for the Intent (but not the Intent
				// itself)
				stackBuilder.addParentStack(GradesActivity.class);
				// Adds the Intent that starts the Activity to the top of the
				// stack
				stackBuilder.addNextIntent(resultIntent);
				PendingIntent resultPendingIntent = stackBuilder
						.getPendingIntent(0, PendingIntent.FLAG_UPDATE_CURRENT);
				mBuilder.setContentIntent(resultPendingIntent);
[...]

PR #1244, opened 2019-01-11
Show notifications for new grades

thellmund:
This pull request implements notifications for new grades, as specified in #688. Whenever `DownloadService` is run, `GradesBackgroundUpdater` fetches the user’s grades and compares them with existing grades in `GradesStore`. If any new grades have been added, we display a notification via `GradesNotificationProvider`.

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/tumui/grades/GradesStore.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/tumui/grades/GradesStore.kt
@@ -18,9 +18,8 @@ class GradesStore @Inject constructor(
     private val gradedCourseNames: List<String>
         get() = sharedPrefs.getStringSet(KEY_GRADED_COURSES, emptySet()).toList()
 
-    fun getGradedCourses(): GradesStoreState {
-        return GradesStoreState(isFirstRefresh, gradedCourseNames)
-    }
+    val state: GradesStoreState
+        get() = GradesStoreState(isFirstRefresh, gradedCourseNames)
 
     fun storeGradedCourses(courses: List<String>) {
         sharedPrefs.edit().putStringSet(KEY_GRADED_COURSES, courses.toSet()).apply()
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/tumui/grades/GradesStore.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/tumui/grades/GradesStore.kt
@@ -3,12 +3,23 @@ package de.tum.`in`.tumcampusapp.component.tumui.grades
 import android.content.SharedPreferences
 import javax.inject.Inject
 
+data class GradesStoreState(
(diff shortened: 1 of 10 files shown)


app/src/main/java/de/tum/in/tumcampusapp/component/tumui/grades/GradesStore.kt
```diff
@@ -0,0 +1,32 @@
...
+        val isFirstRefresh: Boolean,
+        val existingGrades: List<String>
+)
+
+class GradesStore @Inject constructor(
+        private val sharedPrefs: SharedPreferences
+) {
+
+    private val isFirstRefresh: Boolean
+        get() = sharedPrefs.contains(KEY_GRADED_COURSES).not()
```
kordianbruck: Yea... no. This is not the best idea. Besides polluting the shared prefs with stuff that should not be in there, it is a rather cumbersome solution for the initialization problem. How about just skipping the notification, if there are more than 5 new grades? Seems a lot simpler and in practice, nobody will receive 5 grades at exactly the same time or within a refresh period. 

Thoughts?

kordianbruck: (Yes, magic constants and so on, that @pfent will complain about, but its a simpler solution for that problem)

app/src/main/java/de/tum/in/tumcampusapp/service/DownloadService.kt
```diff
@@ -111,7 +115,9 @@ class DownloadService : JobIntentService() {
         val newsSuccess = downloadNews(force)
         val eventsSuccess = downloadEvents()
         val topNewsSuccess = downloadTopNews()
-        return updateNoteSuccess && cafeSuccess && kinoSuccess && newsSuccess && topNewsSuccess && eventsSuccess
+        val gradesSuccess = downloadGrades()
+        return updateNoteSuccess && cafeSuccess && kinoSuccess
+                && newsSuccess && topNewsSuccess && eventsSuccess && gradesSuccess
```
kordianbruck: uggh, this class is getting uglier by the number of times we are editing it. We should really write up a todo, to finally fix it.

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/grades/GradesStore.kt
thellmund: Yes, I was not too happy with the initialization problem either. Will update as per your suggestion.

app/src/main/java/de/tum/in/tumcampusapp/service/DownloadService.kt
pfent: I tried to break it up a little bit in  #1188.
Really doesn't make a lot of sense having it all in one big service :man_shrugging:

PR #1246, opened 2019-01-14
Refactor display of cafeteria menus with RecyclerView and adapter

thellmund:
WIP


app/src/main/java/de/tum/in/tumcampusapp/utils/Extensions.kt
```diff
@@ -135,3 +135,22 @@ fun TextView.addCompoundDrawablesWithIntrinsicBounds(
 operator fun CompositeDisposable.plusAssign(disposable: Disposable) {
     this.add(disposable)
 }
+
+fun <T1, T2> List<T1>.splitOnChanged(transform: (T1) -> T2): List<List<T1>> {
```
pfent: wait. I don't really get this method.
Isn't this essentially the same as:
```kotlin
this.groupBy(transform).values
```
or am I missing something?

pfent: This is what I came up with.
```kotlin
fun <T1, T2> List<T1>.groupConsecutive(transform: (T1) -> T2): List<List<T1>> {
    return zip(map(transform)).foldRight(
        Pair<MutableList<MutableList<T1>>, T2?>(mutableListOf(), null)
    ) { (item, value), (results, latestValue) ->
        if (value == latestValue) {
            results.last().add(item)
        } else {
            results.add(mutableListOf(item))
        }
        results to value
    }.first
}
```
I'm not exactly sure this "a good idea for how to simplify it" :see_no_evil:

PR #1249, opened 2019-01-22
TopNews Crashes Main Page

kordianbruck:
Fix bug in topnews, where empty response would crash the front page

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/ui/news/TopNewsStore.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/ui/news/TopNewsStore.kt
@@ -50,7 +50,7 @@ class RealTopNewsStore @Inject constructor(
 ) : TopNewsStore {
 
     override fun isEnabled(): Boolean {
-        return sharedPrefs.getBoolean(CardManager.SHOW_TOP_NEWS, true)
+        return sharedPrefs.getBoolean(CardManager.SHOW_TOP_NEWS, false)
     }
 
     override fun setEnabled(isEnabled: Boolean) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/de/tum/in/tumcampusapp/component/ui/news/TopNewsStore.kt
```diff
@@ -60,7 +61,10 @@ class RealTopNewsStore @Inject constructor(
         val showTopNews = sharedPrefs.getBoolean(CardManager.SHOW_TOP_NEWS, true)
 
         val displayUntil = sharedPrefs.getString(Const.NEWS_ALERT_SHOW_UNTIL, "")
```
thellmund: We should rather add an extension method to `SharedPreferences` that can return `null`. Then, we can singly perform a null check instead of using a try-catch.

kordianbruck: That still does not make this code part any cleaner or nicer to look at. The best option would rather be to actually not trigger the display of the card, if there are no top news returned from the API. That seemed also a bit complex as in regard how the data is parsed and when the display of the card is checked. 

Feel free to push to this branch.

kordianbruck: Actually a 404 causes an exception, that I did not know how to deal with, as it would not allow to be passed on to the Store, but would fail in the repository.

I've added some better logic to handle an empty response from the backend.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/news/TopNewsStore.kt
```diff
@@ -84,7 +91,7 @@ class RealTopNewsStore @Inject constructor(
         // --> Card should be displayed again
         val update = oldShowUntil != newsAlert.displayUntil || oldImage != newsAlert.url
         if (update) {
-            sharedPrefs.edit().putBoolean(CardManager.SHOW_TOP_NEWS, true).apply()
+            setEnabled(true)
         }
 
         sharedPrefs.edit().putString(Const.NEWS_ALERT_SHOW_UNTIL, newsAlert.displayUntil).apply()
```
jacqueline8711: Why aren't we updating this value in the SharedPrefs when we are updating the image url and link a few lines before? Shouldn't make a difference and would look nicer

PR #1253, opened 2019-02-04
New dep versions

kordianbruck:
Updated all those alpha libs


kordianbruck:
@jacqueline8711 good enough

PR #1269, opened 2019-02-28
Improve placeholders

jacqueline8711:
## Screenshots
screenshot_20190228-163324
screenshot_20190228-163320
screenshot_20190228-165326

Showing the movie title when the cover is not/slowly loading makes it possible for users to know what movie they are looking at offline/with poor connection (or simply when swiping quickly).
The other changes just make things look nicer.


kordianbruck:
Very cool!

PR #1272, opened 2019-03-13
Refactor feedback screen

thellmund:
This pull request refactors the feedback screen into an MVP architecture. Mainly, this results in a cleaner interplay between View and Presenter (formerly FeedbackController). I’ve also removed functionality that was not needed, such as location tracking (we really only need to retrieve the location when the user sends the feedback).


thellmund:
I switched to using `Parcelize` ([link](https://android.jlelse.eu/yet-another-awesome-kotlin-feature-parcelize-5439718ba220)) instead of implementing `Parcelable` ourselves. This resolves the crash and also reduces the amount of boilerplate code we have to write.

jacqueline8711:
@thellmund I fixed it myself, I hope that was okay. I think everything works now.

thellmund:
@jaqueline8711 Thanks for finishing the PR :+1

Issue #491, opened 2017-06-28
Get rid of hard coded opening hours


kordianbruck:
Currently all opening hours are hardcoded inside the `locations.csv` - this is really bad.

Need to implement an API endpoint and redo the manager.

arisun:
Hi I would like to contribute on this issue. Is the issue  still open ?

kordianbruck:
@arisun yes! We still have that ugly CSV in the assets folder. 

We would love to have it replaced by a API method on our server. Should I set this up for you?

zepatrik:
Someone working on this should also add the possibility to have separate opening hours during the semester and during the semester break.

jacqueline8711:
@i-walker Are you still planning on implementing this? The first step being the usage of the endpoint https://app.tum.de/Api/openingtimes

kordianbruck:
@jacqueline8711 I think he is currently traveling, but will be back by the end of the week. Let us give him some more time.

i-walker:
Can someone help me with the Apiclient I can't find it ?

PR #1280, opened 2019-03-31
opening setup

i-walker:
## Issue
[#491](https://github.com/TUM-Dev/Campus-Android/issues/491)

Still in WIP but I need review along the way. 

Can I swap the old Javafiles with a kotlin implementation in component/ui/openinghour

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/api/app/TUMCabeAPIService.java
+++ b/app/src/main/java/de/tum/in/tumcampusapp/api/app/TUMCabeAPIService.java
@@ -27,6 +27,7 @@ import de.tum.in.tumcampusapp.component.ui.chat.model.ChatRoom;
 import de.tum.in.tumcampusapp.component.ui.news.model.News;
 import de.tum.in.tumcampusapp.component.ui.news.model.NewsAlert;
 import de.tum.in.tumcampusapp.component.ui.news.model.NewsSources;
+import de.tum.in.tumcampusapp.component.ui.openinghour.OpeningHours;
 import de.tum.in.tumcampusapp.component.ui.studyroom.model.StudyRoomGroup;
 import de.tum.in.tumcampusapp.component.ui.ticket.model.Event;
 import de.tum.in.tumcampusapp.component.ui.ticket.model.Ticket;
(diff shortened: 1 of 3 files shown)


app/src/main/java/de/tum/in/tumcampusapp/api/app/TUMCabeAPIService.java
```diff
@@ -209,6 +211,10 @@
 
     // TICKET SALE
 
+    // openingHours\
+    @GET("Api/openingtimes")
+    Observable<OpeningHours.Result> displayCategory(@Query("category") String category);
```
kordianbruck: There is not `category` filter on this endpoint. Please just download the whole file once and do any filtering on the clientside. Cache should be efficient enough, to not make this a problem.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/openinghour/OpeningApi.kt
```diff
@@ -0,0 +1,23 @@
...
+                    .addCallAdapterFactory(
+                            RxJava2CallAdapterFactory.create())
+                    .addConverterFactory(
+                            GsonConverterFactory.create())
+                    .baseUrl(BASE_URL)
+                    .build()
+
+            return retrofit.create(OpeningApi::class.java)
+        }
+    }
```
kordianbruck: You don't need to build your own API here - we already have all of this.

i-walker: Kannst du mir in der Zukunft ein Link Schicken, wenn du ueber eine File redest.

i-walker: [But this is not the Endpoint you created](https://github.com/i-walker/Campus-Android/blob/e65e242c65ef7d479910547694338c1e16ac9e7e/app/src/main/java/de/tum/in/tumcampusapp/api/app/TUMCabeClient.java#L74)

kordianbruck: Retrofit generates all paths for all interface defintions. You don't need to create one retrofit client per endpoint. You can simply use the already initialized instance variable (singleton pattern): https://github.com/TUM-Dev/Campus-Android/blob/master/app/src/main/java/de/tum/in/tumcampusapp/api/app/TUMCabeClient.java#L103

PR #1281, opened 2019-04-08
Get Opening Hours from Server

### LINKED ISSUE #491: Get rid of hard coded opening hours

### LINKED ISSUE #1225: Opening Times Cafeteria wrong
Similar to #1211 

> Campus App angezeigten Öffnungszeiten für die Mensa stimmen manchmal nicht (z.B. heute hatte die Mensa bis 16 Uhr auf, aber angezeigt wurde, dass die Mensa ab 14 Uhr geschlossen hatte).

jacqueline8711:
## Issue
Fixes #491 , Fixes #1225 

## Screenshot
Screenshot_20190408-205258 (2)
Screenshot_20190408-211111 (1)

we can update the opening hours at all times so the user will have more accurate opening hours


app/src/main/java/de/tum/in/tumcampusapp/component/ui/openinghour/model/Location.kt
```diff
@@ -25,8 +26,11 @@ data class Location(@PrimaryKey
                     var name: String = "",
                     var address: String = "",
                     var room: String = "",
+                    @SerializedName("transport_station")
                     var transport: String = "",
+                    @SerializedName("opening_hours")
                     var hours: String = "",
+                    @SerializedName("infos")
```
kordianbruck: I guess you did not change anything with any database model except this, so a migration is not needed right?

jacqueline8711: Yes, no migration is needed

PR #1285, opened 2019-04-19
Hide buy button and number of remaining tickets if no tickets available

AndreasChris:
## Issue

Up until now the "Buy Ticket" Button was always visible for future events and the "Remaining Tickets" section was always visible for future and past events.

The changes in these commits ensure that:
- The "Buy Ticket" button is only displayed for future events for which there are (still) tickets available, while the "View Ticket" button remains visible for all events for which a ticket has been purchased by the user.
- The "Remaining Tickets" Section displays the number of remaining tickets if it is positive and a text explaining that tickets may still be available at the evening box office if the number of remaining tickets is zero. If the number of remaining tickets is negative, the "Remaining Tickets" section is hidden (can be used for events that do not sell tickets).

## Screenshot

[see screenshot album on imgur](https://imgur.com/a/939jy5N)

This feature is especially useful for events that should be displayed as such, but do not want to sell tickets via the app (e.g. UNITY) or do not sell tickets at all (e.g. Diversityfilm, MaiTUM, TUNIX, GARNIX, ...). Otherwise people may end up confused wether they can still and/or need to purchase tickets.


jacqueline8711:
Continued in PR #1311

Issue #1286, opened 2019-04-19
Ticket view crashing when clicking "View Ticket" after reset of TUMonline token


AndreasChris:
In the process of testing the changes I made to the event section of the app, I purchased several tickets (for Mario and Drachenzähmen leicht gemacht) in order to check all corner cases. Inititally displaying the purchased tickets worked perfectly fine.

After everything was done, I deleted the app from my phone as well as the associated TUMonline token in my TUMonline account, reinstalled the app and reconnected it to TUMonline. Since then, a click on the "View Ticket" Button results in the App crashing (for both events I have purchased tickets for).

## Steps to Reproduce
1. Install the App and connect it to TUMonline
2. Purchase a ticket for any event
3. Uninstall the app
3.5. Delete old TUMonline token from your Account (might not be necessary)
4. Reinstall the App
5. Reconnect it to TUMonline with a newly generated token
6. Navigate to the event you purchased tickets for and click on "View Ticket"

## Expected Result
Ticket including QR-Code should be shown.

## Actual Result
The App crashes.

```
  E/AndroidRuntime: FATAL EXCEPTION: main
    Process: de.tum.in.tumcampus, PID: 25393
    java.lang.RuntimeException: Unable to start activity ComponentInfo{de.tum.in.tumcampus/de.tum.in.tumcampusapp.component.ui.ticket.activity.ShowTicketActivity}: java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
        at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2817)
        at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2892)
        at android.app.ActivityThread.-wrap11(Unknown Source:0)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1593)
        at android.os.Handler.dispatchMessage(Handler.java:105)
        at android.os.Looper.loop(Looper.java:164)
        at android.app.ActivityThread.main(ActivityThread.java:6541)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.Zygote$MethodAndArgsCaller.run(Zygote.java:240)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:767)
     Caused by: java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
        at java.util.ArrayList.get(ArrayList.java:437)
[...]

PR #1292, opened 2019-04-22
Temporarily fix show ticket crash

jacqueline8711:
## Issue
#1286

App crashes because the ticket types are not downloaded and I cannot find the actual bug right now. I checked that the download method is called and it recognizes the login. The tickets themselves are downloaded right after the login.

For now I am explicitly downloading the ticket types for the ticket that should be displayed since there are upcoming events and about 7 users have unsuccessfully tried to access their tickets according to Crashlytics.


kordianbruck:
@jacqueline8711 thanks for this hotfix and for pushing out the new release! :tada:

Issue #1287, opened 2019-04-20
Don't show new grades notification again after dismissing


jacqueline8711:
## Steps to Reproduce

1. Receive a new grade (on TUMonline)
2. Dismiss notification instead of clicking on it
3. Notification will pop up every hour or so

## Expected Result
It shouldn't show again after dismissing it (user probably has already checked the grade in TUMonline if they are dismissing it)

## Actual Result
The notification only stops showing after opening the grades view

### Environment
* Phone: Pixel XL
* OS version: 9
* TUM Campus App version: 3.2-dev
* Language: English

thellmund:
I worked on this some time ago, but didn’t make a pull request for it. Have now done so at #1295. Feel free to make changes if you see fit 😊

PR #1295, opened 2019-04-24
Prevent repeat grade notifications

thellmund:
## Issue
This fixes the following issue(s): #1287.


app/src/main/java/de/tum/in/tumcampusapp/component/tumui/grades/GradeNotificationDeleteReceiver.kt
```diff
@@ -0,0 +1,43 @@
...
+
+        val newGrades = intent.getStringArrayListExtra(KEY_NUMBER_OF_GRADES)
+        val existingGrades = gradesStore.gradedCourses
+        val updatedGrades = existingGrades + newGrades
+        gradesStore.store(updatedGrades)
+    }
+
+    companion object {
+
+        private const val KEY_NUMBER_OF_GRADES = "KEY_NUMBER_OF_GRADES";
```
jacqueline8711: This kind of sounds like an integer to me. Could you maybe rename it, e.g. KEY_NEW_GRADES_LIST ?

thellmund: Right. Fixed this now.

Issue #1153, opened 2018-10-22
MainActivity.onAlwaysHideCard() crashes with index -1


pfent:
>ArrayIndexOutOfBoundsException
>length=10; index=-1
>CardAdapter.remove (CardAdapter.java:104)
>MainActivity.onAlwaysHideCard (MainActivity.java:199)

probably do a sanity range check here

thellmund:
I’ve pushed a quick fix in #1155, but will investigate the actual cause.

thellmund:
Prevented with #1307.

PR #1307, opened 2019-05-08
Fix various bugs and allow better error tracing

thellmund:
This pull request includes some bug fixes and integrates some tools that should make testing easier: 
- Adds a global error handler for RxJava: We currently have some crashes that occur in download operations that run in the background. This should not crash the app, so we will catch those exceptions in the global error handler and log them to Firebase. 
- Fixes a crash in LecturesDetailsActivity (user clicked on button before lecture information was downloaded)
- Fixes an issue where receiving a notification could crash the app (wrong `Context`)
- Fixes an issue where storing news and news sources could crash the app (results from the API were null, resulting in a `NullPointerException`)
- Prevents `IndexOutOfBoundsException` when `adapterPosition` returned `-1` for in a `CardAdapter.ViewHolder`
- Adds `LeakCanary` for detecting leaks and `DebugDB` to debug databases and shared preferences

### DIFF
--- a/.idea/inspectionProfiles/profiles_settings.xml
+++ b/.idea/inspectionProfiles/profiles_settings.xml
@@ -1,7 +1,0 @@
-<component name="InspectionProjectProfileManager">
-  <settings>
-    <option name="PROJECT_PROFILE" value="Project Default" />
-    <option name="USE_PROJECT_PROFILE" value="true" />
-    <version value="1.0" />
-  </settings>
-</component>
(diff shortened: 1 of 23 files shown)


thellmund:
I’d say _prevent_ instead of _fixed_, but yes. I closed #1153.

thellmund:
Oddly enough, the build works just fine locally since the last commit 🤷‍♂

Issue #1298, opened 2019-04-24
Grades show 4.0 twice


kordianbruck:
![image](https://user-images.githubusercontent.com/298860/56661635-098e0100-6657-11e9-900e-108525ec4b6c.png)

Once in red and once in gray.

PR #1308, opened 2019-05-14
Remove second 4,0 from grades

thellmund:
## Issue

This fixes the following issue(s): #1298


jacqueline8711:
Actually there is an extra grade which means 'Ungraded' which is supposed to be grey and the black is for 5,0

thellmund:
How do you suggest we change it then?

PR #1312, opened 2019-05-28
Implement in-app updates

### LINKED ISSUE #1296: Add force update option
At some point we may not want or can support an old version of the app.

An implementation example can be found [here](https://medium.com/@sembozdemir/force-your-users-to-update-your-app-with-using-firebase-33f1e0bcec5a) which uses remote config from Firebase. 
Since this only works if the code is in the old version we don't want to support anymore we should implement this soon.
We can also use this to simply make the user aware that there is a new version if they haven't updated after a certain time and not necessarily force them to update if it is not necessary.

jacqueline8711:
## Issue

Fixes #1296 

Not sure about whether to put it in MainActivity, StartupActivity or BaseActivity (at least the onResume to cover all entry points)
Not tested yet. 

## Screenshot
TODO when testing

https://developer.android.com/guide/app-bundle/in-app-updates

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -113,6 +113,7 @@ dependencies {
     // Play services
     implementation 'com.google.android.gms:play-services-base:16.1.0'
     implementation 'com.google.android.gms:play-services-location:16.0.0'
+    implementation 'com.google.android.play:core:1.5.0'
 
     // Firebase
     implementation 'com.google.firebase:firebase-core:16.0.8'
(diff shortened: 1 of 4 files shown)


kordianbruck:
So `BaseActivity` does not seem to fit as we want to be very predictable when such notifications happen. I'd also think that `MainActivity` is too late. Ideally we would show the update prompt, before any other requests/refreshes are started. It might be harder to implement it during startup, but it would save us some headaches. I'd also just do the immediate upgrades, so that people get the newest version right away. 

Thoughts?

thellmund:
What’s the state of this? @jacqueline8711’s implementation looks good to me (maybe encapsulate in a custom class) and I’m not sure I understand @kordianbruck’s concerns

jacqueline8711:
@thellmund I think my main problem was that I didn't know how to test if the code actually works since I didn't know how to trigger the in-app update / choose if the flexible or immediate update should be shown

Issue #1322, opened 2019-06-24
Convert code base to Kotlin


jacqueline8711:
Overview of java files:
(please check if conversion is in progress and remove if it is done)
- [ ] api > app (networking) @thellmund 
- [x] (2) component > other > generic > adapter
- [x] (1) ImageViewTouchFragment.java
- [x] (2) Settings @thellmund 
- [x] (9) tumui > Calendar
- [x] (1) tumui > grades > ExamListAdapter.java @jacqueline8711 
- [x] (7) tumui > lectures
- [x] (1) tumui > person > PersonSearchSuggestionsProvider.java
- [x] (4) tumui > roomfinder
- [x] (1) tumui > tuitionfees @jacqueline8711 
- [x] (5) ui > barrierfree
- [x] (11) ui > cafeteria
- [x] (12) ui > chat @thellmund 
- [x] (2) ui > eduroam @thellmund 
- [x] (4) ui > onboarding @thellmund 
- [x] (5) ui > overview (cards)
- [x] (13) ui > ticket
- [x] (5) ui > transportation
- [x] (4) ui > tufilm
- [x] (1) database > TcaDB.java
- [x] (2) utils > sync
- [x] (2) utils
- [x] (2) App / TestApp

Liqs-v2:
Is this still an active issue and is the task list up to date?

CommanderStorm:
> Is this still an active issue and is the task list up to date?

Partially, see
https://github.com/TUM-Dev/Campus-Android/tree/master/app/src/main/java/de/tum/in/tumcampusapp/api/app

PersonSearchSuggestionsProvider is kotlin..

CommanderStorm:
@kordianbruck you can check that one of the list

kordianbruck:
Great, I think we can honestly close this out for now. The API client/networking isn't a big deal and we can do that eventually.

PR #1333, opened 2019-07-25
Convert MainFragment, CardAdapter, CardManager and remaining cards

jacqueline8711:
## Issue

#1322 

Also fixes 'Restore Cards' button


app/src/main/java/de/tum/in/tumcampusapp/component/ui/overview/CardAdapter.kt
```diff
@@ -0,0 +1,117 @@
...
+    private fun validatePosition(fromPosition: Int, toPosition: Int): Int {
+        val selectedCard = mItems[fromPosition]
+        val cardAtPosition = mItems[toPosition]
+
+        // If there is a support card, it should always be the first one
+        // except when it's been dismissed.
+        // Restore card should stay at the bottom
+        if (selectedCard is RestoreCard) {
+            return fromPosition
+        } else if (selectedCard is SupportCard) {
```
kordianbruck: Nice to have: these two could be merged into a single condition no?

PR #1336, opened 2019-07-25
Convert barrier-free classes

jacqueline8711:
## Issue

#1322


app/src/main/java/de/tum/in/tumcampusapp/component/ui/barrierfree/model/BarrierFreeContact.kt
```diff
@@ -5,7 +5,7 @@ import de.tum.`in`.tumcampusapp.component.other.generic.adapter.SimpleStickyList
 /**
  * The model used to display contact infromation in barrier free page
  */
-data class BarrierfreeContact(var name: String = "",
+data class BarrierFreeContact(var name: String = "",
```
kordianbruck: Does this not rename the db table?

Issue #1337, opened 2019-07-25
Searching doesn't work (PersonSearch and RoomFinder)


jacqueline8711:
## Problem
Search is never performed (no network requests)

### Cause
Broken since switch from Activities to Fragments (#1313)

kordianbruck:
Thanks Frank!

Issue #678, opened 2018-02-20
Migrate Searches to unified SearchActivity


kordianbruck:
Idea is to have one God-Like Search, that asynchronously searchs persons, mvv, lectures and others.

PR #1339, opened 2019-07-26
Fix search functionality (People and Rooms)

jacqueline8711:
## Issue
#1337


thellmund:
What’s still work-in-progress here, @jacqueline8711?

jacqueline8711:
@thellmund The recents in the room finder doesn't work, but that could be fixed in a different PR. It would be better anyway if the layout was the same as for Person search (the recents in the room finder were displayed as some grey overlay)

thellmund:
@jacqueline8711 Moving this to a separate PR would be great. Then we can merge this one and I can fix all code styling issues with ktlint (see #1340) and detekt (upcoming PR). 

Also, with regards to search, we should eventually consider @kordianbruck’s suggestion for a unified search (#678).

PR #1342, opened 2019-08-17
Deactivate ProGuard for debug builds

thellmund:
For some reason, we had ProGuard enabled for debug builds, which negatively impacts our build speed. I ran a few tests and had ~30% shorter build times with ProGuard turned off. So, this pull request deactivates ProGuard for debug builds and only activates it for release builds.


kordianbruck:
It was actually intended. We had issues, that only came up when using pro-guard. So we basically enabled it for developers, so those issues would not show up in release-only builds. (Cause nobody was testing them)

I'm happy to go back to disable it for debug, but I fear we will again have some of those release only bugs that don't get detected until released.

kordianbruck:
Closing stale PR.

PR #1346, opened 2019-10-12
Add in-app language selection

SchroDev:
This pull request adds the option to change the application's language inside the app itself. The default value is equal to the device language. After the language has been changed, the app is restarted for the change to take effect. Each language is noted in its own way of spelling.

## Screenshots

Screenshot
Screenshot2

* The application's language can be different from the device language.
* If the device language is not supported by the app, the user can decide himself/herself which language he/she would like to use. No one is forced into using English as the fallback language. Instead, one can use German, for example, if one's German skills are better.


app/src/main/java/de/tum/in/tumcampusapp/component/other/generic/activity/BaseActivity.kt
```diff
@@ -41,4 +48,30 @@ abstract class BaseActivity(
 
         return super.onOptionsItemSelected(item)
     }
+
+    // load language from user's preferences
+    private fun initLanguage(context: Context) {
+        val sharedPreferences = PreferenceManager.getDefaultSharedPreferences(context)
+        var lang = sharedPreferences.getString("language_preference", null)
```
kordianbruck: I think we can default this to "en" and just ignore the whole if block below. English will always be in there.

app/src/main/java/de/tum/in/tumcampusapp/component/other/settings/SettingsFragment.kt
```diff
@@ -192,6 +193,11 @@ class SettingsFragment : PreferenceFragmentCompat(), Preference.OnPreferenceClic
                 StartSyncReceiver.cancelBackground()
             }
         }
+
+        // restart app after language change
+        if (key == "language_preference") {
+            if (activity != null) (activity as SettingsActivity).restartApp()
```
kordianbruck: Please merge the two conditionals into one and && them.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/onboarding/StartupActivity.kt
```diff
@@ -180,6 +180,11 @@ class StartupActivity : BaseActivity(R.layout.activity_startup) {
 
     private fun openMainActivity() {
         val intent = Intent(this, BaseNavigationActivity::class.java)
+        // If the app has been restarted after a language change,
+        // the intent includes an extra which is passed on to BaseNavigationActivity to open up the settings again.
+        if (getIntent().getBooleanExtra("SETTINGS", false)) {
```
kordianbruck: I think there should be some kind of Const class. Can you merge all the "SETTINGS" there, so its used consistently?

kordianbruck:
Thanks for the PR! Much appreciated! Let me know, if you want to join the team.

A few smaller comments only, would be great if you could change those.

app/src/main/java/de/tum/in/tumcampusapp/component/other/settings/SettingsFragment.kt
SchroDev: Done.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/onboarding/StartupActivity.kt
SchroDev: I do not understand what you mean. Consts are constant. I use this extra to return to the settings activity if the app has been restarted after a language change. I am not sure how a Const class could help me here. Maybe you could explain what you mean or submit a commit yourself ;)

Maybe the name of the extra is just a little confusing. It is not actually a setting but just a kind of information which is only needed after the restart of the app.

kordianbruck: Sorry, I see how this was too ambiguous. 

In the `utils` package you will find a `Const.kt` class. There we keep track of strings that are used throughout the app. This helps with consistency and also avoids having naming conflicts. By giving the string a variable name, the code also becomes more readable.

app/src/main/java/de/tum/in/tumcampusapp/component/other/generic/activity/BaseActivity.kt
kordianbruck: Ah, makes perfect sense! Of course I did not think about that "initial" phase.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/onboarding/StartupActivity.kt
SchroDev: Ah, thanks. It's done.

kordianbruck:
Thanks for contributing!

SchroDev:
@kordianbruck Thanks, it would be an honor!

PR #1348, opened 2019-10-26
Interpret Event Location as HTML

mgranderath:
## Issue

The calendar events were showing escaped HTML entities. For example, "Hörsaal 1, &qout;Interims I&qout;".

## Screenshot
| Before | After |
| --------- | ------ |
| image | image |

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/model/Event.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/model/Event.kt
@@ -12,6 +12,7 @@ import de.tum.`in`.tumcampusapp.component.notifications.persistence.Notification
 import de.tum.`in`.tumcampusapp.component.other.locations.model.Geo
 import de.tum.`in`.tumcampusapp.utils.Const
 import de.tum.`in`.tumcampusapp.utils.DateTimeUtils
+import de.tum.`in`.tumcampusapp.utils.Utils
 import org.joda.time.DateTime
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/model/Event.kt
```diff
@@ -37,7 +38,7 @@ data class Event(
         return CalendarItem(
                 id ?: "", status ?: "", url ?: "", title,
                 description ?: "", startTime ?: DateTime(),
-                endTime ?: DateTime(), location ?: "", false
+                endTime ?: DateTime(), Utils.stripHtml(location ?: ""), false
```
kordianbruck: Shouldn't we rather decode them instead of removing them? They might be adding context to the text.

See: https://stackoverflow.com/questions/2918920/decode-html-entities-in-android

mgranderath: This is essentially what it does. `Utils.stripHtml` is a wrapper around `Html.fromHtml().toString()` so it converts the HTML to the actual string representation. I understand the confusion because of the function name though. Would it maybe be better to rename the `stripHtml` function since that is a wrong naming?

kordianbruck: Ah right, I didn't check. Its fine then. I might adapt the naming or if you really want, feel free to open up a second PR.

PR #1356, opened 2020-09-17
Update various dependencies

kordianbruck:
Fix breakages with new calendar version. Update to minSDK 24 because QR code scanner updated.


app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarFragment.kt
```diff
@@ -116,7 +114,7 @@ class CalendarFragment : FragmentForAccessingTumOnline<EventsResponse>(
                 todayButton.visibility = if (isToday) View.GONE else View.VISIBLE
                 TransitionManager.beginDelayedTransition(swipeRefreshLayout)
```
kordianbruck, line 115: I'm not even sure if this is needed anymore. Ideas on replacing this piece of code?

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/grades/GradesFragment.kt
```diff
@@ -161,7 +161,7 @@ class GradesFragment : FragmentForAccessingTumOnline<ExamList>(
 
             xAxis.apply {
                 granularity = 1f
-                setValueFormatter { value, _ -> grades[value.toInt()] }
+                //setValueFormatter { value, _ -> grades[value.toInt()] }
```
kordianbruck, line 164: It seems to be working just fine without this formatter.

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarFragment.kt
Bentipa: Do you know what this was for? Showing a separate "Today"-Button but hide it when the day is visible?

kordianbruck: Yes, it was hidden before. I've actually updated it now to always show that button. I don't think its so super useful to hide the button.

Issue #1357, opened 2020-10-02
Coloring of uncommon grades


logolive:
## Steps to Reproduce

1. Get (Simulate) a grade of (say) 1.5
2. In the app, navigate to "My Grades"
3. Look at the grade's background color

## Expected Result
As 1.5 lies between 1.4 and 1.7, the color should be somewhat greenish

## Actual Result
It shows up as grey.

## Problem
Grades outside the X.{0,3,4,7} scheme rarely appear.
Still they are possible whenever several grades are combined.
See e.g. the interdisciplinary project in M.Sc. Informatics,
which consists of two seperately graded parts worth 5 and 11 ECTS.

Based on the Color wheel at the top (which doesn't even show the grade even in it's grey color)
I assume that colors were assigned to individual grades rather than defined for the entire spectrum 1.0-5.0 .

## Suggested Enhancement :sparkles:
If this assumption holds, replace the color association by some continuous function.
E.g. use interpolation or something simple like that.

### Environment
* Phone: Blackview BV 8000 Pro
* OS version: Android 7.0
* TUM Campus App version: 3.5 (Version Code 571)
* Language: English (US)

kordianbruck:
Thansk to @joschahenningsen for fixing this!

PR #1358, opened 2020-10-09
Add colors for uncommon grades

joschahenningsen:
## Issue

This fixes the following issue(s): #1357 

## Screenshots Before/After:
piechartBeforeAfter
barchartBeforeAfter

## New color palette: 
colors

## Changes I made

- add colors for all grades with one decimal place (1.0, 1.1... 4.8, 4.9, 5.0)
- Round all grades with more decimal places when drawing it's colors

I'm not yet happy with the implementation of the bar-charts legend. If you feel comfortable with the MPAndroidChart API and have any suggestions regarding my solution, hit me up :)


pfent:
:+1:
Much better than the previous state. IMHO, we can even think about removing the legend altogether. The color gradient should be pretty much self explanatory (maybe except for the red-green colorblind). Another option would be to have a continuous legend instead of discrete steps, but I'm not aware, if MPAndroidChart supports that.

Also currently the color palette looks continuous from 1.0 to 5.0. However, I think we should have a break between 4.0 and everything higher, since that has the additional sematic of "fail". What do you think?

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/grades/GradesFragment.kt
```diff
@@ -134,7 +137,11 @@ class GradesFragment : FragmentForAccessingTumOnline<ExamList>(
             setDrawEntryLabels(false)
             legend.isWordWrapEnabled = true
             description = null
-
+            // the legend should not contain all possible grades but rather the most common ones
+            legend.setEntries(legend.entries.filter {
+                it.label != null && !it.label.contains(Regex("[1-5],[1245689]"))
```
pfent, line 142: We probably don't want to recompile the Regex for every entry we want to filter. Could you please move the regex object to a static / companion variable?

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/grades/GradesFragment.kt
```diff
@@ -159,13 +166,47 @@ class GradesFragment : FragmentForAccessingTumOnline<ExamList>(
...
+            val newLegend = ArrayList<LegendEntry>()
+            for (i: Int in 0..4) {
+                for (j: Int in listOf(0, 3, 7)) {
+                    newLegend.add(
+                            LegendEntry(
+                                    null, Legend.LegendForm.DEFAULT, 10f, 10f, null,
+                                    ColorTemplate.rgb(resources.getString(GRADE_COLORS[10 * i + j]))
+                            )
+                    )
+                    if (i == 4) {
```
pfent, line 197: that iteration logic isn't too obvious. I think just hardcoding it in an `DEFAULT_GRADE_COLORS` array might be simpler

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/grades/GradesFragment.kt
```diff
@@ -195,8 +236,13 @@ class GradesFragment : FragmentForAccessingTumOnline<ExamList>(
     private fun calculateGradeDistribution(exams: List<Exam>): ArrayMap<String, Int> {
         val gradeDistribution = ArrayMap<String, Int>()
         exams.forEach { exam ->
-            val count = gradeDistribution[exam.grade] ?: 0
-            gradeDistribution[exam.grade] = count + 1
+            // The grade distribution now takes grades with more than one decimal place into account as well
+            var cleanGrade = exam.grade!!
+            if (cleanGrade.contains(Regex("[1-4],[0-9][0-9]+"))) {
```
pfent, line 241: Same as above. Please avoid constructing the Regex object in the loop.

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/grades/GradesFragment.kt
joschahenningsen: Without the legend we won't need to worry about all this mess :+1:

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/grades/GradesFragment.kt
joschahenningsen: Yup, thanks for the tip, doing that!

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/grades/GradesFragment.kt
joschahenningsen: :+1:

joschahenningsen:
Thanks a lot for your suggestions!
I removed the legend from the bar chart all together since the chart is labeled already, that removes a lot of clutter in the code.
I also added new colors for every grade that's considered as failed. Now a 3.7 is still sort of orange and everything from 4.0 downwards gets progressively darker. Have a look at these screenshots: 

colors
screenshots

pfent (approved):
Looks good to me :+1:

pfent:
@joschahenningsen I just noticed this now misses the "grades without weights" disclaimer, we most probably don't want to get rid of that.
@kordianbruck What do you think? I think only you can approve / merge?

joschahenningsen:
> @joschahenningsen I just noticed this now misses the "grades without weights" disclaimer, we most probably don't want to get rid of that.
You are absolutely right, I'll add this back in later today. I also noticed, that that the vertical lines in the bar chart are not at the X,{0,3,7} marks anymore. I could get that back to the previous state as well.

joschahenningsen:
On a second thought, creating an irregular grid (larger gap between ,3 and ,7) might look weird. How does this look? 

barchart

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/grades/GradesFragment.kt
```diff
@@ -413,12 +433,23 @@ class GradesFragment : FragmentForAccessingTumOnline<ExamList>(
...
+                R.color.grade_2_0, R.color.grade_2_1, R.color.grade_2_2, R.color.grade_2_3,
+                R.color.grade_2_4, R.color.grade_2_5, R.color.grade_2_6, R.color.grade_2_7,
+                R.color.grade_2_8, R.color.grade_2_9,
+                R.color.grade_3_0, R.color.grade_3_1, R.color.grade_3_2, R.color.grade_3_3,
+                R.color.grade_3_4, R.color.grade_3_5, R.color.grade_3_6, R.color.grade_3_7,
+                R.color.grade_3_8, R.color.grade_3_9,
+                R.color.grade_4_0, R.color.grade_4_1, R.color.grade_4_2, R.color.grade_4_3,
+                R.color.grade_4_4, R.color.grade_4_5, R.color.grade_4_6, R.color.grade_4_7,
+                R.color.grade_4_8, R.color.grade_4_9,
+                R.color.grade_5_0, R.color.grade_default
```
kordianbruck, line 452: This makes me think, if it just wouldn't be simpler to round the grades to the closest known grade without the additional inbetween steps. But eh, now you already did the work so we can leave it as is.

kordianbruck (approved):
Looks good to me, thanks for working on this!

kordianbruck:
Hmm, I think I merged this too early and there is a commit missing or something. @joschahenningsen did you forgot to push the latest changes?

joschahenningsen:
Hi @kordianbruck, thanks! :sparkles: 

I actually forgot to push my latest changes which added the legend back to the bar chart.
(Difference: https://github.com/joschahenningsen/Campus-Android/commit/9e1fec72525a3277eb68b2c61c19c744ae8606ea)

Also the value formatter doesn't compile on my end suddenly, maybe because of changes in libraries you updated in the meantime? This would be fixed in https://github.com/joschahenningsen/Campus-Android/commit/d9e3b15e4434d92935fa1aaf60ab0950851c3c71

Should i open a new PR?

kordianbruck:
Yea, Please send a new PR if you can :)

PR #1366, opened 2020-11-20
Create recurring events

### LINKED ISSUE #1297: Reoccurring calendar entries
User requested to be able to create calendar entries, instead of single date entries. (Similar to how its possible in TUMonline)

I think we will need to adapt the API in TUMonline for this unfortunately.

joschahenningsen:
## Issue

This fixes #1297  

## Screenshot
Screenshot

## Note
~~Events created this way can only be deleted or edited one by one (Because there is no api endpoint for real event-series),
I could however add support for editing them all together but that might take a while depending on how my current courses play out :D~~

Any comments on the styling of the options? They might be improvable.


app/src/main/res/values-de/strings.xml
```diff
@@ -15,6 +15,10 @@
     <string name="no_internet_connection">Keine Internetverbindung</string>
 
     <!-- Calendar -->
+    <string name="ends_after">Ende nach</string>
+    <string name="ends_on">Ende am</string>
+    <string name="repeats">Wöchentlicher Termin</string>
+    <string name="create_event_too_many_error">Es können höchstens 25 Termine gleichzeitig erstellt werden.</string>
```
kordianbruck, line 21: Oder halt 6 monate

app/src/main/res/layout/activity_create_event.xml
```diff
@@ -111,6 +111,91 @@
...
+                                android:text="14"
+                                android:inputType="number"/>
+                            <TextView
+                                android:layout_width="wrap_content"
+                                android:layout_height="match_parent"
+                                android:textAppearance="@style/TextAppearance.AppCompat.Medium"
+                                android:clickable="true"
+                                android:focusable="true"
+                                android:gravity="center_vertical"
+                                android:textColor="@color/black"
```
kordianbruck, line 180: nit: all of this hardcoding of the text color will break the "darktheme" I think

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CreateEventActivity.kt
```diff
@@ -242,17 +283,63 @@ class CreateEventActivity : ActivityForAccessingTumOnline<CreateEventResponse>(R
         }
         event.description = description
         this.event = event
+        events = generateEvents()
 
-        val apiCall = apiClient.createEvent(event, null)
-        fetch(apiCall)
+        for (curEvent in events!!) {
+            val apiCall = apiClient.createEvent(curEvent, null)
+            fetch(apiCall)
```
kordianbruck, line 290: So what is supposed to happen when one of the requests fails? Will it fail everything? Will it try to complete the rest?

kordianbruck (commented):
I think we should add a visual note to the repeated fields to let folks know that it creates individual events. Just like a little heads up.

Other than that this looks pretty solid, just a few comments.

app/src/main/res/layout/activity_create_event.xml
joschahenningsen: :facepalm: You are right... how do we handle the dark mode? Is using something like color/text_dark_primary sufficient or do we need a custom style?

joschahenningsen:
Thanks a lot for your suggestions! I'll implement them as soon as possible

> I think we should add a visual note to the repeated fields to let folks know that it creates individual events. Just like a little heads up.
I implemented a way to keep track of events that belong to a series, they can now be deleted in one click (assuming one doesn't reinstall the App).

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CreateEventActivity.kt
joschahenningsen: Currently the ProgressActivity handles these errors and displays error messages/a retry dialog. If it's impossible to fetch all requests even after retrying, the CreateEventActivity would not automatically finish (Worst case: the user clicks on create again and events are created twice).

We could close the activity upon error and show a dialog advising the user to delete eventually created events and try again.

app/src/main/res/layout/activity_create_event.xml
SiKreuz: About the dark mode: The standard text color is defined in resources (v27) style:
```xml
<item name="android:textColorPrimary">@color/text_primary</item>
<item name="android:textColorSecondary">@color/text_secondary</item>
```
So, if this text does not need an own fixed color, just delete this line. It will take the default colors. And else, never use a specific color directly, but define it in the colors.xml. After adding the second colors.xml for dark mode it will automatically grab the correct color depending to the settings.

joschahenningsen: Thanks Simon! Looks a lot better now. If I have any further questions regarding this, I'll give you a call :)

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CreateEventActivity.kt
kordianbruck: Yea, so this retry behavior might be unexpected. With your addition of the new database column, you could insert the created entries one after each other, helping you to not create the same entry twice. Not sure that "advising" the user is the best UX :stuck_out_tongue:

joschahenningsen: I'm not entirely sure what you mean, my concern is the following: 10 events should be created, the 6th, 7th.. fail 5 events are created in the app and uploaded to tum online. 
My preferred solution would be to delete all previously created events but chances are that the API problem is likely not resolved (eg. bad internet connection) now half of the entries are created and can not be deleted from the remote calendar. 
Is that the kind of problem you are thinking about or am I missing something else? :)

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarDetailsFragment.kt
```diff
@@ -129,17 +130,25 @@ class CalendarDetailsFragment : RoundedBottomSheetDialogFragment() {
     }
 
     private fun displayDeleteDialog(eventId: String) {
+        val s = TcaDb.getInstance(requireContext()).calendarDao().getSeriesIdForEvent(eventId)
         val alertDialog = AlertDialog.Builder(requireContext())
                 .setTitle(R.string.event_delete_title)
                 .setMessage(R.string.delete_event_info)
                 .setPositiveButton(R.string.delete) { _, _ -> deleteEvent(eventId) }
-                .setNegativeButton(R.string.cancel, null)
-                .create()
```
kordianbruck, line 137: Does this work without `create` or `setNegativeButton` (in the case that its not a series)

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarDetailsFragment.kt
```diff
@@ -129,17 +130,25 @@ class CalendarDetailsFragment : RoundedBottomSheetDialogFragment() {
...
+        }
         alertDialog.show()
     }
 
+    private fun deleteEventSeries(seriesId: String){
+        val calendarItems = TcaDb.getInstance(requireContext()).calendarDao().getCalendarItemsInSeries(seriesId)
+        calendarItems.forEach{
+            deleteEvent(it.nr)
+        }
+    }
```
kordianbruck, line 150: This should probably be a function inside the calendarDao, so it can be potentially reused somewhere else.

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CreateEventActivity.kt
```diff
@@ -86,16 +95,46 @@ class CreateEventActivity : ActivityForAccessingTumOnline<CreateEventResponse>(R
...
             if (isEditing) {
                 editEvent()
             } else {
                 createEvent()
             }
         }
+
+        // edited events cannot repeat
+        if (isEditing) {
+            repeatingSwitch.layoutParams = LinearLayout.LayoutParams(0, 0)
```
kordianbruck, line 124: Is `LinearLayout.LayoutParams(0, 0)` a standard practice? If not, please add a comment about your intent, so its understandable what you wanted to happen.

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CreateEventActivity.kt
```diff
@@ -86,16 +95,46 @@ class CreateEventActivity : ActivityForAccessingTumOnline<CreateEventResponse>(R
...
+        } else {
+            repeatingSwitch.setOnCheckedChangeListener { _, isChecked ->
```
kordianbruck, line 126: nit: Is it smart to define a Change Listener within a `setOnClickListener`? I guess this works in Android, but for me as a Web Developer this just screams for a hard debuggable problem. (feel free to ignore this)

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CreateEventActivity.kt
```diff
@@ -86,16 +95,46 @@ class CreateEventActivity : ActivityForAccessingTumOnline<CreateEventResponse>(R
...
+                if (isChecked) {
+                    repeats = true
```
kordianbruck, line 128: nit: To be honest: this value should be read in the `createEvent` function if possible. All the data that is stored on the class level makes it hard to debug in a failure scenario, because there are so many paths to get to a certain state. (Feel free to ignore this - its just me being unhappy with android)

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CreateEventActivity.kt
kordianbruck: I think we are on the same page. 

What I'm trying to hint at is: if you insert the entries into the database as you do each call, then the set of events you have in the DB with the same series ID, will let you do a more graceful "retry" command. This is however a lot of work and as you say, chances are, the connection is still bad.

So I'd suggest in a failure scenario to do the following: inform the user about the failure with "X out of Y events were created" and just close the dialog. This will give the user actionable intel about what they need to do to fix it and it makes the control flow easier to understand in this case.

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CreateEventActivity.kt
```diff
@@ -242,17 +291,67 @@ class CreateEventActivity : ActivityForAccessingTumOnline<CreateEventResponse>(R
...
+    private fun generateEvents(): List<CalendarItem> {
+        if (!repeats) {
+            return listOf(this.event!!)
+        }
+        val items = ArrayList<CalendarItem>()
+        seriesId = generateSeriesId()
+
+        // event ends after n times
+        if (endAfterRadioBtn.isChecked) {
+            val numberOfEvents = eventRepeatsTimes.text.toString().toInt()
```
kordianbruck, line 315: You need a sanity check here, what if its `0`?

btw. you have the same code in `createEventButton.setOnClickListener`. I'd suggest only fetching the value once and then adding it as a function parameter to  `createEvent` - this makes debugging easier.

kordianbruck: Also: if you do it as a parameter, then the error checking becomes easier and your control flow in the onclicklisteneer can be reduced as well.

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CreateEventActivity.kt
```diff
@@ -242,17 +291,67 @@ class CreateEventActivity : ActivityForAccessingTumOnline<CreateEventResponse>(R
...
+                item.dtend = curDateEnd
+                curDateStart = curDateStart.plusWeeks(1)
+                curDateEnd = curDateEnd.plusWeeks(1)
+                items.add(item)
+            }
+        }
+        return items
     }
 
     override fun onDownloadSuccessful(response: CreateEventResponse) {
```
kordianbruck, line 343: nit: well, I guess the base class wasn't really build for this use case of doing multiple fetches. I guess it works just fine.

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CreateEventActivity.kt
```diff
@@ -242,17 +291,67 @@ class CreateEventActivity : ActivityForAccessingTumOnline<CreateEventResponse>(R
...
-        event?.let {
+        events?.get(apiCallsFetched++)?.let {
```
kordianbruck, line 344: So if you are sure that the fetch calls are synchronous, this works. Are you sure? ^^

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CreateEventActivity.kt
```diff
@@ -312,4 +411,18 @@ class CreateEventActivity : ActivityForAccessingTumOnline<CreateEventResponse>(R
...
+
+        /**
+         * generates a random id that is assigned to every CalendarItem in a series
+         */
+        private fun generateSeriesId(): String {
+            return (1..10)
+                    .map { kotlin.random.Random.nextInt(0, CHARPOOL.size) }
+                    .map(CHARPOOL::get)
+                    .joinToString("")
+        }
```
kordianbruck, line 426: why not use a standard method like: `UUID.randomUUID().toString()` ?

kordianbruck: It also has some more guarantees about collisions :)

app/src/main/res/values/strings.xml
```diff
@@ -17,6 +17,12 @@
     <string name="app_short_description" translatable="false">TUM information platform - from students for students</string>
 
     <!-- Calendar -->
+    <string name="ends_after">Ends after</string>
+    <string name="ends_on">Ends on</string>
+    <string name="repeats">Repeats Weekly</string>
+    <string name="create_event_too_many_error">Event series can not be longer than 6 months.</string>
+    <string name="create_event_too_little_error">Event series need to have at least two events.</string>
```
kordianbruck, line 24: `needs`

joschahenningsen:
> Sorry for the huge number of comments - I'm not trying to be picky and I hope you can also learn a thing. If it gets annoying, let me know please :)

Don't worry, I really appreciate the effort you put into this, plus I learn a lot :D

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarDetailsFragment.kt
joschahenningsen: In the Dao we can only delete local entries, since we have to iterate over all events to delete them remotely I think this might be the better solution, what do you think?

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CreateEventActivity.kt
joschahenningsen: Done :) I simulated this by injecting an error, looks like this and appears to work fine:

Screenshot_20201125-100455~2

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CreateEventActivity.kt
joschahenningsen: The sanity check takes place before this function is called (in `createEventButton.setOnClickListener`) this is properly not best practice, but taking the number as a parameter would also be quite annoying, since there would need to be parameters for when the event repeats n times and for when it ends after a specific date. I could however overload the method or use class attributes. Which option would you prefer?

kordianbruck (approved):

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarDetailsFragment.kt
kordianbruck: Sure, works either way. Probably won't need this a lot anywhere else.

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CreateEventActivity.kt
kordianbruck: Lets keep it as is and refine later. I think doing this with an overloaded method is the best way in java.

joschahenningsen:
For now this looks like it works fine, I'll keep refining this. If there are any problems you stumble across please assign me @kordianbruck :)

kordianbruck:
This PR is already huge (I usually like to keep it <500 lines), so may I recommend we merge now and you'll do one or two followup PRs for any outstanding work?

joschahenningsen:
Luckily 1200 of these lines are auto-generated but yes, that's what I thought as well.

kordianbruck:
It still bloats the diff :P

SiKreuz:
Did you test this well? Since this was merged (commit https://github.com/TUM-Dev/Campus-Android/commit/c11175fda76a2172e5800b49a55fd0ca062c5793) the app crashes already on startup. The previous commit is working. I get this error and similar ones for all the other cards on the overview:
```
E/WM-WorkerWrapper: Work [ id=8741e296-9f20-442f-b15c-fc89465fe5b5, tags={ de.tum.in.tumcampusapp.service.SendMessageWorker } ] failed because it threw an exception/error
    java.util.concurrent.ExecutionException: java.lang.IllegalStateException: Migration didn't properly handle: eventSeriesMappings(de.tum.in.tumcampusapp.component.tumui.calendar.model.EventSeriesMapping).
```

joschahenningsen:
This was tested. Which commit are you working on? The database was changed to version 6 and there is not migration from 6 back to 5, so going back to a previous version should indeed crash the app. What happens if you delete the apps data and install from the newest commit?

joschahenningsen:
Apparently there is a problem when upgrading the database in some cases. Thanks for noticing, I'll fix this.

PR #1371, opened 2020-11-28
Clean up repeating events

joschahenningsen:
cleaning up code introduced in #1366 

Changes:
- [x] Add Helper class that holds information on the repetition settings of the current event
- [x] Access user input in one place only
- [x] Refine handling of wrong user input
- [x] Events in a series that got updated stay in the series
- [x] remove mappings when series is deleted

Todo:

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarDetailsFragment.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarDetailsFragment.kt
@@ -146,6 +146,7 @@ class CalendarDetailsFragment : RoundedBottomSheetDialogFragment() {
         calendarItems.forEach {
             deleteEvent(it.nr)
         }
+        TcaDb.getInstance(requireContext()).calendarDao().removeSeriesIdMappings(seriesId)
     }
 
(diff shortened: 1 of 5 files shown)


app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarDetailsFragment.kt
```diff
@@ -136,17 +135,18 @@ class CalendarDetailsFragment : RoundedBottomSheetDialogFragment() {
...
     }
 
-    private fun deleteEventSeries(seriesId: String){
+    private fun deleteEventSeries(seriesId: String) {
         val calendarItems = TcaDb.getInstance(requireContext()).calendarDao().getCalendarItemsInSeries(seriesId)
-        calendarItems.forEach{
+        calendarItems.forEach {
             deleteEvent(it.nr)
         }
+        TcaDb.getInstance(requireContext()).calendarDao().removeSeriesIdMappings(seriesId)
```
kordianbruck, line 149: This again might lead to inconsistencies - but meh. Not sure its worth fixing.

kordianbruck (approved):
Nice splitting up! This is much more readable :+1:

PR #1373, opened 2020-12-02
fix database migration

joschahenningsen:
@SiKreuz noticed, that the current version of the app can be installed but occasionally crashes on upgrade due to the database changes I introduced. 

@SiKreuz, can you verify that this fixes the exact problems you had?


SiKreuz (approved):
Thank you Joscha. I can confirm, that this changes fix the upgrade. My build is not crashing anymore.

kordianbruck (approved):

kordianbruck:
Do you need to update the database json file?

joschahenningsen:
No, that only changes when the schema is updated (everything that is annotated with Entity for example) the migration doesn't affect it.

kordianbruck:
You know this stuff better by now probably than me :+1:

PR #1375, opened 2020-12-04
Library reservations

joschahenningsen:
## Screenshot

. | . 
--|--
Screenshot_20201204-120232 | Screenshot_20201204-120225

This is not yet ready, especially the styling, but it works and feels very useful. The reason I'm already opening this is, that we decided on integrating this into our backend. Currently the API client is mocked to use [the API I created](https://github.com/joschahenningsen/TUMBibAPI). @kordianbruck, would you feel comfortable adding me to Cabe so I can work on implementing it there or would you rather do that yourself?
Also I hope these PRs are not getting out of hand, this is not high priority so no need to review this too soon :)


joschahenningsen:
Works now as soon as the api changes are deployed.

joschahenningsen:
Screenshot_20201208-150637~2

Except for the large gap between the button and the time, I'm quite happy with this :)

joschahenningsen:
Library reservations are no more :(

Issue #1384, opened 2021-03-29
Average Grade ignores the credit in calculation and takes simple average


tatban:
## Steps to Reproduce

1. Take subjects with different credits
2. Get different grades

## Expected Result
Average Grade = sum(c_i * g_i) / sum(c_i)
where c_i = credit for ith passed subject, g_i = grade for ith passed subject

## Actual Result
Average Grade = sum(g_i) / n
where n = total no of passed subjects, c_i = credit for ith passed subject, g_i = grade for ith passed subjects

## Problem
It should take the credit into consideration while calculating the Average Grade, otherwise this calculation is currently treating grades for different workload subjects equally and the calculation is not matching with the TUM Online GPA calculation.

## Suggested Enhancement :sparkles:
Change the formula for Average Grade calculation to the formula mentioned in Expected result

### Environment
* Phone: Redmi Y2
* OS version: Android 9
* TUM Campus App version: v3.9
* Language: English

CommanderStorm:
Credit-rules are a bit more confusing, than you are thinking..

you have to account for how a grade is "GF"-rated. normally this is 1, but there are courses where this is 0.5 -> multiply with GF

you would also have to account for credits, that have no associated grade (pass-or-fail). -> filter both credits and grades by the ones that have a grade.

also some study subjects have rules how many Credits can you have in a certain subject-group (e.g. "Überfachliche Grundlagen" in the B.Sc. Informatics can have max. 7 credits.) If you have more credits than that it is usually the rule that the best grades get chosen and the other ones get discarded. 

there may be other things to concider.
I have no ideas what "ECTS-Credits (negativ)" ("CREN") are.

tatban:
Thanks a lot @CommanderStorm for your comment. Good to know different other scenarios, which of course I didn't know as I never took such subjects so far. However, my point was to make it consistent with the official GPA calculation instead of taking simple average. This is because when some professor updates the grade of a particular exam, it first comes to the App before it gets published to TUMOnline officially. So, seeing just simple average grade as 'Average Grade' can be misleading in a big way. 

So, I was just asking if it is possible for TUM-Dev team to collaborate with the TUMOnline team to implement the same logic such that the App shows the same 'Average Grade' as to be shown on TUMOnline later when official grades gets published. (Assuming the grade is not changed after exam review, and if changed, change should be reflected on both the the places). 

[...]

PR #1385, opened 2021-03-30
Add hint for average grade 

joschahenningsen:
This adds a hint on the average grade in the TCA indicating, that it's not the actual one from TUMOnline.

Seems to be all we can do regarding #1384 currently


kordianbruck (approved):

kordianbruck:
Thanks!

PR #1391, opened 2021-05-01
Fixed chat message style

SiKreuz:
The chat messages were oddly padded. I fixed this, did the todo in the code and set a max-width for chat messages and additionally added a timestamp for incoming messages (which was missing for no reason).

## Old design

image

## New design

image


kordianbruck (approved):

kordianbruck:
Awesome, thanks!

PR #1398, opened 2021-09-29
Parse html escape sequences in lecture text

jfoerste:
Parse html character escape sequences in lecture API responses into their respective characters.

## Issue

Html character escape sequences (ie. `&quot;` or `&amp;`) are sometimes present in lecture API responses and appear to the user instead of the represented characters (ie. `"` or `&`).

This shows lecture information as it is shown on the web at `campus.tum.de`.

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/api/tumonline/converters/NullableEscapedStringConverter.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/api/tumonline/converters/NullableEscapedStringConverter.kt
@@ -4,8 +4,7 @@ import android.os.Build
 import android.text.Html
 import com.tickaroo.tikxml.TypeConverter
 
-
-class NullableEscapedStringConverter : TypeConverter<String?>{
+class NullableEscapedStringConverter : TypeConverter<String?> {
 
     /**
      * Parses leftover html escape sequences
@@ -17,7 +16,7 @@ class NullableEscapedStringConverter : TypeConverter<String?>{
         // preserve newlines
         val str = value.replace("\n", "<br />")
 
-        return if(Build.VERSION.SDK_INT >= 24) {
+        return if (Build.VERSION.SDK_INT >= 24) {
             Html.fromHtml(str, Html.FROM_HTML_MODE_LEGACY).toString()
         } else {
             Html.fromHtml(str).toString()
@@ -34,5 +33,4 @@ class NullableEscapedStringConverter : TypeConverter<String?>{
     override fun write(value: String?): String? {
         return value
     }
-
 }
--- a/app/src/main/java/de/tum/in/tumcampusapp/api/tumonline/converters/NullableEscapedStringConverter.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/api/tumonline/converters/NullableEscapedStringConverter.kt
@@ -1,0 +1,38 @@
+package de.tum.`in`.tumcampusapp.api.tumonline.converters
+
+import android.os.Build
+import android.text.Html
+import com.tickaroo.tikxml.TypeConverter
+
+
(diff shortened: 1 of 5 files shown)


app/src/main/java/de/tum/in/tumcampusapp/api/tumonline/converters/NullableEscapedStringConverter.kt
```diff
@@ -0,0 +1,36 @@
...
+     *
+     * @param value String with leftover escape sequences
+     * @return unescaped String
+     */
+    private fun unescapeString(value: String): String {
+        // preserve newlines
+        val str = value.replace("\n", "<br />")
+
+        return if (Build.VERSION.SDK_INT >= 24) {
+            Html.fromHtml(str, Html.FROM_HTML_MODE_LEGACY).toString()
```
kordianbruck, line 20: Shouldn't this have an allow list of which HTML tags we actually want to support? Otherwise this might be a XSS entry point. Or am I missing something?

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/lectures/model/Lecture.kt
```diff
@@ -10,21 +12,21 @@ import de.tum.`in`.tumcampusapp.component.other.generic.adapter.SimpleStickyList
  */
 @Xml(name = "row")
 data class Lecture(
-    @PropertyElement(name = "dauer_info") val duration: String = "",
-    @PropertyElement(name = "org_kennung_betreut") val chairTumId: String? = null,
-    @PropertyElement(name = "org_name_betreut") val chairName: String? = null,
+    @PropertyElement(name = "dauer_info", converter = EscapedStringConverter::class) val duration: String = "",
```
kordianbruck, line 15: Maybe we should just make these either all null or all empty string?

kordianbruck (approved):
Thanks, just two minor things. Sorry for the long wait!

app/src/main/java/de/tum/in/tumcampusapp/api/tumonline/converters/NullableEscapedStringConverter.kt
jfoerste: As far as I am aware `Html.fromHtml` does not execute any javascript, does not support iframes and (since no handler is provided) does not load images. Any possible interactivity (like anker tags, which for example disappear completely) should be killed by the `toString()` anyways. Even if that fails and outbound connections were possible I see no possible way to make **authenticated** api calls from this environment, since there is no interaction with the java side of things.

This approach has the advantage that we don't have to rely on knowing all the escape sequences that might arise in the future. Also, this let's us rely on a proven parser implementation, instead of hacking our own, which would carry it's own xss risks.

kordianbruck: `Html.fromHtml` might not be executing the JS but the TextView that is displaying it might. In any case, I take your point and I think  you are right. This should't be a big deal.

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/lectures/model/Lecture.kt
jfoerste: From a brief overview of the consumers of the lecture api, this would not be my approach, as it is conceivable for an endpoint to return either null, indicating that an error message should be displayed, or an empty string which should be interpreted as just that: nothing to display.

On the other hand it might be wise to clean up our consumers to handle both cases the same way, but this would require a more in depth knowledge of the api and intended output which I do not have.

kordianbruck: Ack, got it.

Issue #1378, opened 2020-12-16
setting to switch between weekly and daily for calendar


AW-tum:
a user requested a setting to decide wether you see the calendar as weekly or as daily instead of always having to manually switch to weekly when opening the app

kordianbruck:
Probably we can just save the setting, when switching. This seems easy enough.

Liqs-v2:
I'm gonna pick this issue up, if thats okay.

kordianbruck:
Thats weird, we just shouldn't run firebase in tests as all. I'd need to take a closer look, but that seems super wrong.

Liqs-v2:
Okay thanks, for some more info on what I have tried so far in order to get this to work, check out the [pull request](https://github.com/TUM-Dev/Campus-Android/pull/1399), that I have opened for this issue.

joschahenningsen:
fixed in #1399

PR #1399, opened 2021-10-27
Issue/1378 calendar view preference

Liqs-v2:
## Issue 1378

This fixes the following issue(s): [Calendar view mode preference is not properly stored/retrieved](https://github.com/TUM-Dev/Campus-Android/issues/1378)

## Screenshot

Previously when opening the calendar activity and changing the view mode of the calendar, the change would be lost when exiting and reopening the calendar activity. The issue would also occur when restarting the entire app.
Broken_Calender_view

Now the prefered view mode of the user is properly retrieved from the PreferenceManager when the calendar fragment is created.
Fixed_Calendar_view

Being able to select a preference for the calendar's view mode once and having it be in the same state the next time the app is launched, saves everyone using the app time. Additionally this fix helps the app behave in a way the user would expect it to, by averting data loss and thus increasing usability.

## Open Problems

I have so far been unable to run the tests that were already present in the project when cloned, [see my comment on the issue.](https://github.com/TUM-Dev/Campus-Android/issues/1378#issuecomment-950571857)

### Testing the changed code
I am in the process of adding testing to verify the behaviour of the code. Testing the piece of code requires intrumentation tests, as the code is directly attached to CalendarFragment, which cannot be tests using Roboelectric or otherwise. 

**There seem to be issues with the setup of instrumentation tests regarding this project.** I have so far **not been able to launch any instrumentation test**. It seems like they are not registered by the project's runner.

#### What I have tried so far

- Using a fragment, that **does not** require TUMonline (eg.: SettingsFragment)
- Using a fragment, that **does** require TUMonline (eg.: CalendarFragment)
- Ensuring the **correct `fragmentArgs`** are passed into `launchFragmentInContainer(...)`
- Using `AndroidJUnit4ClassRunner` instead of `AndroidJunit4` in `@RunWith` [as per this thread.](https://stackoverflow.com/questions/52776716/androidjunit4-class-is-deprecated-how-to-use-androidx-test-ext-junit-runners-an)
- I have ensured, that **all necessary dependencies for intrumentation testing are available**, [as per this tutorial from Google.](https://developer.android.com/codelabs/advanced-android-kotlin-training-testing-test-doubles#1) The repository with the gradle file can be found there aswell.
- Creating a **small sample project** and **verifying the dependencies, code and setup** there.

There was no androidTest (so, intrumentation test) folder present in the project when I cloned it. I manually added it, my suspicion is that it might have something with this and in term the project setup. **Any help/pointers would be grately appreciated.**

I would also like to inquire **what the testing policy for this project is**, as I was unable to find any information in that regard.


Liqs-v2:
### I am now investigating the issues with tests, starting with this issue/PR.

@kordianbruck It seems like this line is the problem: FirebaseCrashlytica_disabled

As you can see, disabling it makes all tests that should pass, pass. (Except for one that is wrong)

**Is FirebaseCrashlytics something that is needed?** Because, if so, it seems like something that would be difficult to fix for me **[since FirebaseCrashlytics requires Firebase]**(https://firebase.google.com/docs/crashlytics/get-started?platform=android) and that means I **would need access to the projects' Firebase Console** to further investigate why this line is causing problems.

app/build.gradle
```diff
@@ -192,6 +192,24 @@ dependencies {
...
     // TicketSale
     implementation 'com.stripe:stripe-android:6.1.2'
+
+
+    // Dependencies for Android instrumented unit tests
+    androidTestImplementation "junit:junit:4.13.2"
+    // Testing code should not be included in the main code.
+    // Once https://issuetracker.google.com/128612536 is fixed this can be fixed.
+
+    debugImplementation "androidx.fragment:fragment-testing:1.3.6"
```
joschahenningsen, line 202: This dependency breaks my debug build: When I switch the activity (e.g. open the calendar), the top navigation buttons disappear, are you experiencing this as well?
Does `testImplementation` work for you as well?

joschahenningsen (commented):
Codewise looks good and works as expected. :ok_hand: 
I think my test-related question might be a bigger project that could be tackled in a separate pr.

app/build.gradle
Liqs-v2: Yes, I was also exeriencing this problem and both testImplementation and androidTestImplementation fixed it. [However **according to Google** one is supposed to use **debugImplementation.**](https://developer.android.com/guide/fragments/test#declare-dependencies)

Since testing is a bigger issue anyways (as discussed in Discord), I would **rollback app/build.gradle** and merge my existing PRs as is, without further testing/dependencies and take care of that in another PR. Would that be okay with you?

joschahenningsen: That sounds like the best approach.

joschahenningsen (approved):
Thanks for fixing this!

Issue #1363, opened 2020-11-05
Implement In-App Review API


kordianbruck:
https://developer.android.com/guide/playcore/in-app-review

would be nice to get some more good ratings.

Liqs-v2:
Is someone working on this?

Liqs-v2:
If not I'd pick this up too.

Liqs-v2:
I have implemented this feature and opened a PR for it, would be grateful for a review.

kordianbruck:
Awesome, I guess this can now be closed :)

Thanks a lot!

PR #1400, opened 2021-11-01
Issue/1363 Google Play store in-app reviews

Liqs-v2:
## Issue

This fixes the following issue(s): [Implement In-App Review API#1363](https://github.com/TUM-Dev/Campus-Android/issues/1363)

## Screenshot
In_App_Review_Implemented

## Implementation
The user is **prompted for an in-app review once they have:**
- **connected to TUMOnline**
- **opened the calendar view**
- **opened the grades view**

The user is **prompted again after approximately 6 months**. To have some sort of basis, upon which a decision on whether or not to prompt a review can be reached, some fields have been added to the SharedPreferences.

I have decided against using the database, since the extra overhead of another table that will only have one row anyways, did not seem justified to me. I am aware however, that **storing data that is used for logic in the application "preferences" might not be the best solution**, I am **open to suggestions in this regard**.

### Testing
**Currently there are no tests added to this feature.** The interaction with the **Google Play service handling the review can not be tested directly**, the only thing one could test is the response of the application to a successful prompt. [(see documentation)](https://developer.android.com/guide/playcore/in-app-review/test#test-fakereviewmanager)

**I will add such tests once my questions in regards to testing have been answered** [(refer to my other PR)](https://github.com/TUM-Dev/Campus-Android/pull/1399). If you are willing to merge without tests, please also let me know.

Prompting user's for feedback does not only benefit the developers, but also the entire userbase as a whole. By **generating feedback on a larger scale, bugs, issues  can be surfaced and in turn fixed more quickly** resulting in a better user experience. The additional feedback also makes it possible to **extract feature requests**, resulting in new and or improved functionality and a better experience for the users.

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/ui/overview/MainFragment.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/ui/overview/MainFragment.kt
@@ -102,13 +102,12 @@ class MainFragment : BaseFragment<Unit>(
         }
 
         // Triggers a Google Play store review if the user has experienced some key features of the app
-        // The earliest possible retrigger of a review prompt occurs after half a year
+        // The earliest possible re-trigger of a review prompt occurs after half a year
         val dateFormatter = SimpleDateFormat("MM-yyyy", Locale.GERMANY)
-        // Will be null if this is the first launch of the app and the preference LAST_REVIEW_PROMPT has not been set yet
-        val lastReviewDate = dateFormatter.parse(Utils.getSetting(requireContext(), Const.LAST_REVIEW_PROMPT, ""))
-        val halfYearInSeconds = 183*24*60*60
+        val lastReviewDate = dateFormatter.parse(Utils.getSetting(requireContext(), Const.LAST_REVIEW_PROMPT, "01-01-1970"))
+        val halfYearInSeconds = 6*30*24*60*60
 
-        if (lastReviewDate == null || Date().time > lastReviewDate.time + halfYearInSeconds &&
+        if (Date().time > lastReviewDate!!.time + halfYearInSeconds &&
                 Utils.getSetting(requireContext(), Const.LRZ_ID, "").isNotEmpty() &&
                 Utils.getSettingBool(requireContext(), Const.HAS_VISITED_GRADES, false) &&
                 Utils.getSettingBool(requireContext(), Const.HAS_VISITED_CALENDAR, false)) {
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/ui/overview/MainFragment.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/ui/overview/MainFragment.kt
@@ -7,7 +7,6 @@ import android.net.ConnectivityManager
 import android.net.Network
 import android.net.NetworkRequest
 import android.os.Bundle
-import android.util.Log
 import android.view.View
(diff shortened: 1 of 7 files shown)


app/src/main/java/de/tum/in/tumcampusapp/component/ui/overview/MainFragment.kt
```diff
@@ -95,6 +100,19 @@ class MainFragment : BaseFragment<Unit>(
         viewModel.cards.observe(viewLifecycleOwner) {
             it?.let { onNewCardsAvailable(it) }
         }
+
+        // Triggers a Google Play store review if the user has experienced some key features of the app
+        // The earliest possible re-trigger of a review prompt occurs after half a year
+        val dateFormatter = SimpleDateFormat("MM-yyyy", Locale.GERMANY)
```
joschahenningsen, line 106: Unless I'm missing something, instead of serializing and parsing the last review date you can just save `Date().time` (the timestamp in milliseconds) and compare it to the current time, right?

app/src/main/java/de/tum/in/tumcampusapp/component/ui/overview/MainFragment.kt
```diff
@@ -95,6 +100,19 @@ class MainFragment : BaseFragment<Unit>(
...
+        val lastReviewDate = dateFormatter.parse(Utils.getSetting(requireContext(), Const.LAST_REVIEW_PROMPT, "01-01-1970"))
+        val halfYearInSeconds = 6*30*24*60*60
```
joschahenningsen, line 108: This could be a constant (in a companion object)

joschahenningsen (approved):
Works as expected :+1: 

> storing data that is used for logic in the application "preferences" might not be the best solution

Using shared preferences for this seems straightforward, no complaints there 

What would happen when a user already reviewed the app and we toggle the prompt? Will it be shown?

app/src/main/java/de/tum/in/tumcampusapp/component/ui/overview/MainFragment.kt
joschahenningsen: BTW, you can also use `org.joda.time.*`, it has a few fancy methods that could be handy like `DateTime.now().minusMonths(6).isBefore(...)`

Liqs-v2:
The prompts are **restricted via a quota from Google**. We can show a prompt at most once every 1-2 months, how often exactly is an implementation detail according to Google. So if a prompt were to be triggered **twice in quick succession, the second prompt would not be let through by the quota set up by Google and thus no prompt will be displayed.**

app/src/main/java/de/tum/in/tumcampusapp/component/ui/overview/MainFragment.kt
Liqs-v2: Thanks for that tip, that does indeed make the code a lot more readable and also allows removal of that ugly constant. I wasn't aware of that date helper class.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/overview/MainFragment.kt
Liqs-v2: These constants are no longer present in the code due to the other suggestion, that you made (except for lastReviewDate). Should I move that to a companion object still?

joschahenningsen (approved):
Looks good, thanks! :beers:

Issue #1392, opened 2021-05-03
Calendar time doesn't respect users current time zone


sobuj53:
## Steps to Reproduce

1. Move to a different time zone
2. Open TUM Campus app
3. Home as well as calendar will not show event time based on current time zone. 

## Expected Result
The app should detect current time zone and show event time accordingly 
## Actual Result
App is showing event time based event's time zone.  
## Problem
Many international students are currently staying on their home country due to online lecture format. But TUM Campus app can't detect our time zone and show event time accordingly. The app would prove more productive if it can show lecture and other events time based on our current time zone instead of standard CET.  
## Suggested Enhancement :sparkles:
Please implement a feature to detect current time zone of TUM Campus app and process event time relative to that information.
### Environment
* Phone: Xiaomi mi5
* OS version: Android 11
* TUM Campus App version: v3.11
* Language: English

joschahenningsen:
Feel free to pick it up if you want :)

Liqs-v2:
I am picking this issue up :)

Liqs-v2:
I have implemented a fix and opened a PR for it.

Liqs-v2:
Fixed in #1401

PR #1401, opened 2021-11-03
Issue/1392 calendar doesnt adopt user timezone

Liqs-v2:
## Issue

This fixes the following issue(s): [Calendar time doesn't respect users current time zone #1392 ](https://github.com/TUM-Dev/Campus-Android/issues/1392)

## Screenshot
When the **user leaves the german timezone, the start and end times stay fixed and do not reflect the "actual" start and end times from the perspective of the user** in another timezone:
Broken_Timezone_fix

**Changing timezones immediately updates the start and end times of events displayed in the calendar**, even if the user has only tabbed out of the app:
Working_Timezone_fix

## Implementation
**When fetching calendar data from the TUMOnline API** the application **checks if the user is currently in the german timezone**, if they are not the timezone offset is computed and the start and end times for the events displayed in the calendar are adjusted.

### Testing
**Currently there are no tests added to this feature.** I will add tests once my questions in regards to testing have been** answered [(refer to my other PR)](https://github.com/TUM-Dev/Campus-Android/pull/1399).

**International students that are traveling outside of the german timezone** (eg to visit family during the holidays), **currently have to manually convert the scheduled times of lectures** to the scheduled times in their current timezone. W**ith this fix, the TUM Campus App automatically detects when a user/device is not in the same timezone as the german timezone and updates all events that are fetched via the TUMOnline API to their respective timezone**, allowing for a better overview of their timetable and increasing the usability of the app.

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarFragment.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarFragment.kt
@@ -104,7 +104,6 @@ class CalendarFragment : FragmentForAccessingTumOnline<EventsResponse>(
         isWeekMode = Utils.getSettingBool(requireContext(), Const.CALENDAR_WEEK_MODE, false)
 
         disableRefresh()
-        loadEvents(CacheControl.USE_CACHE)
 
         // Tracks whether the user has used the calendar module before. This is used in determining when to prompt for a
         // Google Play store review
@@ -117,7 +116,7 @@ class CalendarFragment : FragmentForAccessingTumOnline<EventsResponse>(
 
         // In case the timezone changes when reopening the calendar, while the app is still open, this ensures
         // that the lectures are still adjusted to the new timezone
-        loadEvents(CacheControl.USE_CACHE)
+        loadEvents(CacheControl.BYPASS_CACHE)
     }
 
     override fun onRefresh() {
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarFragment.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarFragment.kt
@@ -135,38 +135,6 @@ class CalendarFragment : FragmentForAccessingTumOnline<EventsResponse>(
         var events = response.events ?: return
         scheduleNotifications(events)
 
-        // If the device is in a different timezone, get the daylight savings corrected device offset that is
-        // then used to adjust the event startTime and endTime
-        val deviceTimeZone = TimeZone.getDefault()
-        val germanTimeZone = TimeZone.getTimeZone("Europe/Berlin")
-
-        if(!deviceTimeZone.equals(germanTimeZone)) {
-            // 1000 * 60 * 60, converts from milliseconds to hours, since 10^(-3) seconds * 10^(3) = 1 second
-            //      and there are 60 seconds in a minute and 60 minutes in an hour
-            val millisecondsToHoursFactor = 1000 * 60 * 60
-            val offset = (deviceTimeZone.getOffset(Date().time) - germanTimeZone.getOffset(Date().time)) / millisecondsToHoursFactor
-
-            // A new variable is needed to be able to update event.startTime and event.endTime, since these are
-            // values and changing them to variables will break code in Event.kt
-            var timeZoneCorrectedEvents = emptyList<Event>()
-            for(event in events){
-                timeZoneCorrectedEvents = timeZoneCorrectedEvents.plusElement(Event(
-                        event.description,
-                        // If a null value is present in the database simply propagate it forward
-                        if (event.startTime != null) event.startTime.plusHours(offset) else null,
-                        if (event.endTime != null) event.endTime.plusHours(offset) else null,
-                        event.geo,
-                        event.location,
-                        event.id,
-                        event.status,
-                        event.title,
-                        event.url
-                ))
-            }
-
-            events = timeZoneCorrectedEvents
-        }
-
         compositeDisposable += Completable
                 .fromAction { calendarController.importCalendar(events) }
                 .subscribeOn(Schedulers.io())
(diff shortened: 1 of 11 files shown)


app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarFragment.kt
```diff
@@ -126,9 +128,41 @@ class CalendarFragment : FragmentForAccessingTumOnline<EventsResponse>(
...
+                        event.id,
+                        event.status,
+                        event.title,
+                        event.url
+                ))
+            }
+
+            events = timeZoneCorrectedEvents
+        }
+
```
joschahenningsen, line 165: The best solution for this might be to change the Event's attributes `startTime` and `endTime` to be an instance of `LocalDateTime` and parse them with German Locale from the API. We'd then be able to just display them in the device's timezone. This might be a little annoying database-wise though.

Another simple solution would be to manipulate the `toCalendarItem` method of the `Event` to convert the times to german local times and then convert them back to dateTimes in the devices timezone like this: 

```Kotlin
val tzGer = DateTimeZone.forID("Europe/Berlin")
val startInDeviceLocation = LocalDateTime(startTime, tzGer).toDateTime(DateTimeZone.getDefault())
val endInDeviceLocation = LocalDateTime(endTime, tzGer).toDateTime(DateTimeZone.getDefault())
```

Liqs-v2: I think I'll go with the second suggestion, unless you would really prefer to go with the first. I missed that there already was a loop going through the events hidden in that method.

I can sadly not test this right now though, because TUMOnline seems to be having issues and wont let me log in with my token.

joschahenningsen:
This looks good :) 
I never used the event notifications and the silence mode, but they might be wrong in other timezones as well, this needs some investigation.
Maybe it would make sense to put the converting logic into their own methods (`event.getStartInDeviceTimezone()`...) so we can use it for the notifications as well, what do you think @Liqs-v2?

https://github.com/TUM-Dev/Campus-Android/blob/1ef43061f757cbbb83111016382b25f4ef37982e/app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/model/Event.kt#L53
https://github.com/TUM-Dev/Campus-Android/blob/105b81fd97e4ccbf3c79e0ece1c765334743f02d/app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarDao.kt#L16
(for the silence service, the current lectures are retrieved from the database, this could also be done by filtering all events if this is easier than working with timezones in SQL)

If you don't feel like picking this up as well, I can do a follow up :grinning:

Liqs-v2:
Well I was originally gonna work on the tests after I got these issues fixed, but with how that is going it might be better to just go ahead and fix some more bugs.

I ll let you know later this week. In principle I'd be okay with fixing this follow up too.

Liqs-v2:
Since I am having a lot of trouble with adding tests, I will instead investigate the followup that you suggested @joschahenningsen

kordianbruck (approved):
Pretty cool, thanks!

Liqs-v2:
@joschahenningsen is there even a possibility to activate calendar notifications? Because `CalendarFragment` checks for `calendarController.hasNotificationsEnabled()` before scheduling any notifications, which looks like this:

`override fun hasNotificationsEnabled() = Utils.getSettingBool(context, "card_next_phone", false)`

However this preference is never set anywhere, is this feature unfinished or is this some sort of bug?

Liqs-v2:
Also there was a bug with the adjustment of timezones when staying in the same timezone, different from the german one and loading the events multiple times that I fixed.

joschahenningsen:
> is there even a possibility to activate calendar notifications? 

That's a very good question actually :D assuming we use LocalDateTimes consistently though, it shouldn't make any difference (implementation wise).

There is a setting under next-lecture->Notifications, I'll check out what exactly it does tomorrow.

joschahenningsen:
Funny... 
https://github.com/TUM-Dev/Campus-Android/blob/a6b36a8a2bea1b8f8d9fe94af51cdc819499ac12/app/src/main/res/xml/settings.xml#L204
The actual preference should be called `card_next_lecture_phone`, not `card_next_phone`. This would mean the notification feature was implemented for > two years but never scheduled any notifications :thinking: am I missing something @kordianbruck ?

joschahenningsen:
Yup, changing the key fixes the notifications :D can you please fix this here as well while you're at it? @Liqs-v2

Liqs-v2:
On which revision did you fix the notifications? Because I just found another bug causing the custom created events by `CreateEventActivity.kt` to not be stored in `TcaDb`. 

```
// finish when all events have been created
        if (apiCallsFetched == events.size) {
            setResult(Activity.RESULT_OK)
            finish()
        }
        if (apiCallsFetched + apiCallsFailed == events.size) {
            finishWithError()
        }
```

The problem is that the caller of finish, will run to completion. Finish is only run once the control is handed back to android. And in this case, if all api calls succeed, then both conditions are true, as `apiCallsFetched + 0 = apiCallsFetched == event.size`. This in leads to error flags being set and the event being rolled back and removed from `TcaDb` again.

Adding return statments fixes this.

In addition to that I would also suggest, using `CacheControl.BYPASS_CACHE` when returning to the CalendarFragment, as I found it rather unintuitive, than one has to manually refresh, to see an event that was just manually added.

Liqs-v2:
I had another look at this, and actually the problem with the returns, does not cause this. It simply has to do with the manualy refresh. I would however still fix this, as it could lead to unexpected behaviour down the line.

Also, using `CacheControl.USE_CACHE` when getting data for the fragment to initially display, it will load events that were already deleted, because the Cache is apparently not flushed when an event is deleted, so this should definitely be changed to `CacheControl.BYPASS_CACHE` and potentially be investigated down the line.

Liqs-v2:
There is also a **bug when creating a new calendar entry in another timezone**. But in order t**o fix that one would have to change how events are stored**, because one must **be able to tell in which timezone an event has originate**d. Either that or all events are always mapped into the new timezone and also stored as that, but that would cause issues with the TUM events, which are downloaded in german time.

**Would you like me to also take a look at this in this PR?**

Liqs-v2:
The notifications and silent service work with the adjusted timezones aswell.

Notification_silent_service_timezone

joschahenningsen (approved):
:partying_face:

PR #1406, opened 2021-11-26
Issue/toolbar broken after dependency upgrade

Liqs-v2:
## Issue

This fixes the following issue(s): The recent upgrade to gradle 7 other upgrades to dependencies introduced breaking changed, causing the toolbar or items in the toolbar to disappear as the user navigates the app.

**This PR fixes the most severe** of the bugs introduced making the app usable again for the most part.

## Screenshot

Initial state, completely broken toolbar:
toolbar_broken

Current state, works again in most cases:
toolbar_partial_fix

The toolbar is an essential part of the app, without it navigation becomes a hassle and confusing. Features become inaccessible, both for users and developers.

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarActivity.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarActivity.kt
@@ -1,16 +1,14 @@
 package de.tum.`in`.tumcampusapp.component.tumui.calendar
 
 import android.os.Bundle
-
 import de.tum.`in`.tumcampusapp.R
-import de.tum.`in`.tumcampusapp.component.other.generic.activity.ActivityForAccessingTumOnline
-import de.tum.`in`.tumcampusapp.component.tumui.calendar.model.EventsResponse
+import de.tum.`in`.tumcampusapp.component.other.generic.activity.BaseActivity
 import de.tum.`in`.tumcampusapp.utils.Const
 
 /**
  * Activity showing the user's calendar. Calendar items (events) are fetched from TUMOnline and displayed as blocks on a timeline.
  */
-class CalendarActivity : ActivityForAccessingTumOnline<EventsResponse>(R.layout.activity_calendar) {
+class CalendarActivity : BaseActivity(R.layout.activity_calendar) {
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/TicketDao.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/TicketDao.kt
@@ -10,6 +10,7 @@ interface TicketDao {
     @Insert(onConflict = OnConflictStrategy.REPLACE)
     fun insert(vararg ticket: Ticket)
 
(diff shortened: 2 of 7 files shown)


Liqs-v2:
Alright, so as requested I removed the commits from #1401 polluting this branch. @joschahenningsen

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarActivity.kt
```diff
@@ -1,16 +1,14 @@
...
-import de.tum.`in`.tumcampusapp.component.other.generic.activity.ActivityForAccessingTumOnline
-import de.tum.`in`.tumcampusapp.component.tumui.calendar.model.EventsResponse
+import de.tum.`in`.tumcampusapp.component.other.generic.activity.BaseActivity
 import de.tum.`in`.tumcampusapp.utils.Const
 
 /**
  * Activity showing the user's calendar. Calendar items (events) are fetched from TUMOnline and displayed as blocks on a timeline.
  */
-class CalendarActivity : ActivityForAccessingTumOnline<EventsResponse>(R.layout.activity_calendar) {
+class CalendarActivity : BaseActivity(R.layout.activity_calendar) {
```
joschahenningsen, line 11: What's the reason for this change?

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/TicketDao.kt
```diff
@@ -13,6 +10,8 @@ interface TicketDao {
     @Insert(onConflict = OnConflictStrategy.REPLACE)
     fun insert(vararg ticket: Ticket)
 
+    @SuppressWarnings(RoomWarnings.CURSOR_MISMATCH)
+    @Transaction
```
joschahenningsen, line 14: I assume, Room changed its behavior and complaints about this query now? If yes this could use a //todo

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarActivity.kt
Liqs-v2: Initially I thought only the CalendarActivity was having issues with the toolbar and that was the difference to the other activities, which were already moved to this class from the **deprecated ActivityForAccessingTumOnline**. This is also why I left it in, because the original Parent Class is marked as deprecated.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/ticket/TicketDao.kt
Liqs-v2: Yes, Room is complaining. I will add a TODO.

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarActivity.kt
Liqs-v2: deprecated_activityforaccessingtumonline

joschahenningsen: Alright I see, thanks :)

joschahenningsen (approved):
Thanks! This might need some more fine-tuning but for now this is perfect 👌

PR #1409, opened 2021-12-02
Issue/1349 mvv card settings

Liqs-v2:
## Issue

This fixes the following issue(s): When the preferred stop for a campus location is changed, the change is not reflected in the MVVCard that is displayed in the app's homescreen.

### Broken version
https://user-images.githubusercontent.com/9247500/144461001-e2ba6fb2-0ecb-4dc8-953b-e0ced3d184c3.mp4

### Fixed version
https://user-images.githubusercontent.com/9247500/144458358-51182d5b-5006-4981-8743-3688121e6688.mp4

## Followup suggestion
Changing the default campus location also seems broken in regards to the MVVCard. I'll look into this. **This has been fixed, see below.**

## Other
It seems as though the [incorrect namespace for preference attributes was used](https://developer.android.com/guide/topics/ui/settings/) for the settings. Everything seems to still work, but I find this to be confusing. Perhaps some clarification would be good.

Being able to see public transport departures for the public transport station that one is actually going to use, is a lot more useful, than just getting information regarding departures for the default stationthat was selected by the developers for a specific campus.

### DIFF
--- a/app/src/main/res/xml/settings.xml
+++ b/app/src/main/res/xml/settings.xml
@@ -380,7 +380,7 @@
             android:key="employee_mode"
             android:summary="@string/employee_mode_info"
             android:title="@string/employee_mode" />
-        <androidx.preference.Preferenced
+        <androidx.preference.Preference
             android:icon="@drawable/ic_outline_exit_to_app_24px"
             android:key="button_logout"
             android:title="@string/logout" />
--- a/app/src/main/res/xml/settings.xml
+++ b/app/src/main/res/xml/settings.xml
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <androidx.preference.PreferenceScreen xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:app="http://schemas.android.com/apk/res-auto"
     android:key="cards_pref_container">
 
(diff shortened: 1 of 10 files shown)


Liqs-v2:
Changing the default campus requires a restart currently for the MVVCard to be updated.

Liqs-v2:
Changing the default campus no longer breaks the MVVCard now:

### Previous verion
https://user-images.githubusercontent.com/9247500/144565667-2cecd22e-bdd7-458d-9036-4d8327a18354.mp4



### Fixed version
https://user-images.githubusercontent.com/9247500/144565691-de105f14-185a-41de-8337-6597355cc26f.mp4

app/src/main/res/xml/settings.xml
```diff
@@ -380,7 +380,7 @@
             android:key="employee_mode"
             android:summary="@string/employee_mode_info"
             android:title="@string/employee_mode" />
-        <androidx.preference.Preference
+        <androidx.preference.Preferenced
```
joschahenningsen, line 383:
```suggestion
        <androidx.preference.Preference
```

Liqs-v2: This seems to be a type, fixed that.

joschahenningsen (approved):
Thanks :)

PR #1411, opened 2021-12-05
fix double url encoding and cleaner coord-api error handling

jfoerste:
## Issue

1. double URL encoding:
	When calling the room-coordinate api the identifiers were url-encoded two times: once manually using `ApiHelper.encodeUrl()` and once automatically by retrofit. The result does not decode to the original anymore and the api returns an error.
Example: `02.09.014@5609` -> `01.06.020%405606` -> `01.06.020%25405606` 
This is fixed by eliminating the first, manual encoding.
2. room-coordinate api error handling
	When the api returns an error (with http code 200), it is now caught and logged, instead of trying to decode empty Strings.
	example:


joschahenningsen (approved):
Thanks!

PR #1412, opened 2021-12-05
replace "by lazy" view references in BaseFragment

jfoerste:
## Issue

The `BaseFragment` initalizes all references to views by using the `by lazy {}`  construct. This works in activites but is problematic in  fragments because when replaced, their views are destroyed but the fragments themselves might be reused. `by lazy{}` is inadequate in this scenario because
1. all views are only initialized once, which might lead to all kinds of crazy behavior
2. (more overtly) `by lazy{}` captures references to the views which are not removed when the view is destroyed. This may leak the referenced views

To fix this I replaced these initializations with a safer albeit uglier method which allows for nulling the references in `onDestroyView()`. Example: 
```kotlin
private var _errorLayout: LinearLayout? = null
    get() = if(field == null) requireActivity().findViewById<LinearLayout>(R.id.error_layout).also { e -> field = e } else field

private val errorLayout: LinearLayout
    get() = _errorLayout!!
```

I am very open to an abstraction of this boilerplate or even an entirely different approach. For example, it would be easiest if every subclass only references and cleans up the views it actually interacts with.

<details>
<summary>Example of typical LeakCanary report</summary>

```lang-none
┬───
│ GC Root: System class
│
├─ leakcanary.internal.InternalLeakCanary class
│    Leaking: NO (BaseNavigationActivity↓ is not leaking and a class is never leaking)
│    ↓ static InternalLeakCanary.resumedActivity
├─ de.tum.in.tumcampusapp.component.other.generic.activity.BaseNavigationActivity instance
│    Leaking: NO (MainFragment↓ is not leaking and Activity#mDestroyed is false)
│    ↓ BaseNavigationActivity.mFragments
├─ androidx.fragment.app.FragmentController instance
│    Leaking: NO (MainFragment↓ is not leaking)
│    ↓ FragmentController.mHost
├─ androidx.fragment.app.FragmentActivity$HostCallbacks instance
│    Leaking: NO (MainFragment↓ is not leaking)
│    ↓ FragmentActivity$HostCallbacks.mFragmentManager
├─ androidx.fragment.app.FragmentManagerImpl instance
│    Leaking: NO (MainFragment↓ is not leaking)
│    ↓ FragmentManagerImpl.mFragmentStore
├─ androidx.fragment.app.FragmentStore instance
│    Leaking: NO (MainFragment↓ is not leaking)
│    ↓ FragmentStore.mActive
├─ java.util.HashMap instance
│    Leaking: NO (MainFragment↓ is not leaking)
│    ↓ HashMap.table
├─ java.util.HashMap$Node[] array
│    Leaking: NO (MainFragment↓ is not leaking)
│    ↓ HashMap$Node[].[1]
├─ java.util.HashMap$Node instance
│    Leaking: NO (MainFragment↓ is not leaking)
│    ↓ HashMap$Node.value
├─ androidx.fragment.app.FragmentStateManager instance
│    Leaking: NO (MainFragment↓ is not leaking)
│    ↓ FragmentStateManager.mFragment
├─ de.tum.in.tumcampusapp.component.ui.overview.MainFragment instance
│    Leaking: NO (Fragment#mFragmentManager is not null)
│    ↓ MainFragment.swipeRefreshLayout$delegate
│                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~
├─ kotlin.SynchronizedLazyImpl instance
│    Leaking: UNKNOWN
│    ↓ SynchronizedLazyImpl._value
│                           ~~~~~~
├─ androidx.swiperefreshlayout.widget.SwipeRefreshLayout instance
│    Leaking: YES (View detached and has parent)
│    mContext instance of de.tum.in.tumcampusapp.component.other.generic.activity.BaseNavigationActivity with mDestroyed = false
│    View#mParent is set
│    View#mAttachInfo is null (view detached)
│    View.mID = R.id.swipeRefreshLayout
│    View.mWindowAttachCount = 1
│    ↓ SwipeRefreshLayout.mParent
╰→ androidx.coordinatorlayout.widget.CoordinatorLayout instance
​     Leaking: YES (ObjectWatcher was watching this because de.tum.in.tumcampusapp.component.ui.overview.MainFragment received Fragment#onDestroyView() callback (references to its views should be cleared to prevent leaks))
​     key = c13eac11-cb25-4c63-8cb7-29be74ba35f8
​     watchDurationMillis = 22128
​     retainedDurationMillis = 17127
​     mContext instance of de.tum.in.tumcampusapp.component.other.generic.activity.BaseNavigationActivity with mDestroyed = false
​     View#mParent is null
​     View#mAttachInfo is null (view detached)
​     View.mWindowAttachCount = 1

METADATA

Build.VERSION.SDK_INT: 26
Build.MANUFACTURER: samsung
LeakCanary version: 2.4
App process name: de.tum.in.tumcampus
Analysis duration: 10673 ms
```
</details>

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/ui/overview/MainFragment.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/ui/overview/MainFragment.kt
@@ -7,9 +7,7 @@ import android.net.ConnectivityManager
 import android.net.Network
 import android.net.NetworkRequest
 import android.os.Bundle
-import android.view.LayoutInflater
 import android.view.View
-import android.view.ViewGroup
 import androidx.lifecycle.ViewModelProviders
 import androidx.recyclerview.widget.ItemTouchHelper
 import androidx.recyclerview.widget.LinearLayoutManager
@@ -62,6 +60,8 @@ class MainFragment : BaseFragment<Unit>(
 
     private val binding by viewBinding(FragmentMainBinding::bind)
 
+    override val swipeRefreshLayout get() = binding.swipeRefreshLayout
+
     @Inject
     lateinit var viewModelProvider: Provider<MainActivityViewModel>
     private val viewModel: MainActivityViewModel by lazy {
@@ -77,15 +77,14 @@ class MainFragment : BaseFragment<Unit>(
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
 
-        swipeRefreshLayout?.setOnRefreshListener(this)
-        swipeRefreshLayout?.isRefreshing = true
-        swipeRefreshLayout?.setColorSchemeResources(
-                R.color.color_primary,
-                R.color.tum_A100,
-                R.color.tum_A200)
-
-
         with(binding) {
+            swipeRefreshLayout.setOnRefreshListener(this@MainFragment)
+            swipeRefreshLayout.isRefreshing = true
+            swipeRefreshLayout.setColorSchemeResources(
+                    R.color.color_primary,
+                    R.color.tum_A100,
+                    R.color.tum_A200)
+
             registerForContextMenu(cardsRecyclerView)
 
             cardsRecyclerView.layoutManager = LinearLayoutManager(requireContext())
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/other/generic/activity/ProgressActivity.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/other/generic/activity/ProgressActivity.kt
@@ -53,7 +53,7 @@ abstract class ProgressActivity<T>(
     }
 
     private val errorLayout: LinearLayout by lazy {
-        findViewById<LinearLayout>(R.id.error_layout)
+        findViewById<LinearLayout>(R.id.layout_error)
     }
 
     private val errorIconImageView: ImageView by lazy {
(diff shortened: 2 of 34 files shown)


app/src/main/java/de/tum/in/tumcampusapp/component/ui/overview/MainFragment.kt
```diff
@@ -160,6 +160,11 @@ class MainFragment : BaseFragment<Unit>(
         super.onDestroy()
     }
 
+    override fun onDestroyView() {
+        cardsRecyclerView.adapter = null
```
joschahenningsen, line 164: With the latest changes from #1417 this only compiles this way:

```suggestion
        binding.cardsRecyclerView.adapter = null
```
Which leads to this error though
```
java.lang.IllegalStateException: Should not attempt to get bindings when Fragment views are destroyed.
```

Can you check out how to handle this with viewBindings?

joschahenningsen (changes requested):
Clever catch :+1: ! Just one quick question

jfoerste:
This approach gets rid of the lazy initializiation in `BaseFragment` by providing two attributes which have to to be overridden when the `BaseFragment`  is supposed to interact with the `ScrollRefreshLayout` and the error layout respectively:

```kotlin
protected open val swipeRefreshLayout: SwipeRefreshLayout? = null
protected open val layoutAllErrorsBinding: LayoutAllErrorsBinding
    get() = throw NotImplementedError()
```

This utilizes the migration to viewbinding and leverages polymorphism. Additionally calls to `findViewById()` are reduced and the memory leak resulting from the use of `by lazy` in fragment should be banished.

joschahenningsen (approved):
Thanks!

And also sorry for the delay :see_no_evil:

PR #1415, opened 2021-12-15
Issue/1414 refactor utils

### LINKED ISSUE #1414: Reduce code rot in Utils.kt
## Problem
`Utils.kt` currently uses the *deprecated `android.preference.PreferenceManager`* instead of the new `androidx.preference.PreferenceManager`.

Since preferences are at the core of the application the deprecation can lead to problems in the future. Thus an upgrade is advisible.

The Utils.kt object could do with some refactoring in general, to reduce code rot.

Liqs-v2:
## Issue

This fixes the following issue(s): fixes #1414 

Keeping code rot to a minimum ensures that future features can be added more easily to TCA and that bugs are found faster. I also reduces the time it takes future developers to get into the project and deliver code that matters to the users.

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/utils/Utils.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/utils/Utils.kt
@@ -3,12 +3,9 @@ package de.tum.`in`.tumcampusapp.utils
 import android.annotation.TargetApi
 import android.app.Activity
 import android.content.Context
-import android.content.Intent
-import android.content.IntentFilter
 import android.content.pm.PackageManager
 import android.graphics.Bitmap
 import android.graphics.Canvas
-import android.os.BatteryManager
 import android.os.Build
 import android.text.Html
 import android.text.Spanned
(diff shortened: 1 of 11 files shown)


Liqs-v2:
I fixed the PR so that changes already in TUM-Dev:master don't pollute it. @joschahenningsen please review this when you find time for it :)

app/src/main/java/de/tum/in/tumcampusapp/utils/Utils.kt
```diff
@@ -246,32 +224,13 @@ object Utils {
...
     }
 
-    /**
-     * Returns a String[]-List from a CSV input stream
-     *
-     * @param fin CSV input stream
-     * @return String[]-List with Columns matched to array values
-     */
-    @JvmStatic
-    fun readCsv(fin: InputStream): List<Array<String>> {
```
kordianbruck, line 263: I was sure this function was still used :joy:

kordianbruck (approved):

PR #1419, opened 2021-12-21
Issue/1416 migrate tumui to viewbinding

Liqs-v2:
## Issue

This fixes the following issue(s): Second part for fixing #1416 

## Why this is useful for all students
Having a working app that can easily be maintained and expanded upon using the newest android features will improve the experience for users and developers alike.

### component.tumui
- [x] `calendar`
- [x] `feedback`
- [x] `grades`
- [x] `lectures`
- [x] `person`
- [x] `roomfinder`
- [x] `tuitionfees`

PR #1417, opened 2021-12-15
Issue/1416 migrate to viewbindings

Liqs-v2:
## Issue

This fixes the following issue(s): First part for #1416 

## Why this is useful for all students
Having a working app that can easily be maintained and expanded upon using the newest android features will improve the experience for users and developers alike.

### component.ui
- [x] `overview`
- [x] `onboarding`
- [x] `openinghour`
- [x] `news`
- [x] `eduroam`
- [x] `chat`
- [x] `cafeteria`
- [x] `barrierfree`
- [x] `alarm`
- [x] `studyroom`
- [x] `ticket`
- [x] `transportation`
- [x] `tufilm`
- [x] `updatenote`

I will **open a new PR for the tumui component migration** to keep this already sizable PR managable.

PR #1420, opened 2021-12-22
Issue/1416 migrate other to viewbinding

### LINKED ISSUE #1416: Migrate from kotlin synthetics to view binding
## Problem
Currently the project uses **kotlin synthetics** which have been **(deprecated)[https://developer.android.com/topic/libraries/view-binding/migration#groovy)** in favor of **view binding.**

Migrate the codebase from kotlin synthetics to view binding, to make sure that the app does not completely break once kotlin synthetics are no longer supported.

This can potentially be done on a per module basis.

Liqs-v2:
## Issue

This fixes the following issue(s): fixes #1416 

Having a working app that can easily be maintained and expanded upon using the newest android features will improve the experience for users and developers alike.

### component.other
- [x] `settings`
- [x] `generic`

This is the last part of the 3 PRs for migrating the project to view bindings. See #1417  and #1419  for the other PRs.

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/other/settings/SettingsFragment.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/other/settings/SettingsFragment.kt
@@ -40,7 +40,6 @@ import io.reactivex.Single
 import io.reactivex.android.schedulers.AndroidSchedulers
 import io.reactivex.disposables.CompositeDisposable
 import io.reactivex.schedulers.Schedulers
-import kotlinx.android.synthetic.main.activity_setup_eduroam.*
 import org.jetbrains.anko.defaultSharedPreferences
 import org.jetbrains.anko.notificationManager
 import java.util.concurrent.ExecutionException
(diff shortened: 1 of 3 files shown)


Liqs-v2:
This should also be mergable right away @joschahenningsen

app/src/main/java/de/tum/in/tumcampusapp/component/other/settings/SettingsFragment.kt
```diff
@@ -40,7 +40,6 @@ import io.reactivex.Single
 import io.reactivex.android.schedulers.AndroidSchedulers
 import io.reactivex.disposables.CompositeDisposable
 import io.reactivex.schedulers.Schedulers
-import kotlinx.android.synthetic.main.activity_setup_eduroam.*
```
kordianbruck, line 43: This doesn't look correct? Do we no longer have the setup eduroam setting?

Liqs-v2: I also felt like this was suspicious, but I checked and this was marked as unused by the IDE even in 29ef2c56b47383ef48101e72b666a8d0c7a67ebe (Release 3.12), which was pushed half a year ago.

Liqs-v2: Okay so it seems like this import was introduced in c60552d6cff11d0f1d6db0640c3dd6d8a1e3d2ec about a year ago. Seems like it is just a residue from someone playing around with a view and removing that before committing, but forgetting to remove the import aswell.

kordianbruck: Ack, thanks for fishing it out!

kordianbruck (approved):

PR #1428, opened 2022-04-22
Fixup latest release and bump to SDK32

kordianbruck:
Fixes #1427

PR #1429, opened 2022-04-23
Add FLAG_IMMUTABLE to remaining PendingIntents

### LINKED ISSUE #1427: Sdk version bump issues
The sdk version bump in 564059be65025d9a9d1645d815409ee868e69037 broke the install and introduced a crash as some new requirements are not met.

1. LeakCanary version must be bumped accordingly, as its Manifest is outdated
2. CompileSdkVersion should propable also be bumped
3. `PendingIntent`s now require explicitly specifying `FLAG_IMMUTABLE` or `FLAG_MUTABLE`, else the app crashes

jfoerste:
Fixes #1427 and extends #1428 

All remaining instantiations of `PendingIntent` in the project have `FLAG_IMMUTABLE` added. Additionally `work-runtime-ktx` is added as a direct dependency as without it some androidx libraries internally rely on an outdated `WorkManager` which created `PendingIntent`s without `FLAG_IMMUTABLE` added yet.

If any of the `PendingIntent`s turn out to have to be mutable (which I haven't noticed) the flag would have to be replaced. But since without this change the call would lead to a crash in any case that should not be a hindrance.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -144,6 +144,8 @@ dependencies {
     implementation 'androidx.core:core-ktx:1.7.0'
     // Note: upgrading this and or androidx.appcompat:appcompat will break the toolbar
     implementation 'androidx.fragment:fragment-ktx:1.2.5'
+    // Note: fix for internal androidx libraries using outdated WorkManager causing a crash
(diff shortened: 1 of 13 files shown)


kordianbruck (approved):
Fantastic, thanks a ton!

Issue #1408, opened 2021-11-28
ActionBar blank after dependency bump


jfoerste:
10a625b185ae04063d3b319f6f75d5e51216ee69 seems to have caused a regression.

## Steps to Reproduce

1. Open any fragment via the drawer

## Expected Result
Everything appears normally.

## Actual Result
The action bar (and the drawer button) become completely blank.

## Screenshot
![](https://i.jfoe.de/KCTX)

Liqs-v2:
Hello, I have deployed a temporary fix for now (see #1409) it is already merged into master aswell.

PR #1434, opened 2022-04-29
Update appcompat, fragment libraries and fix findViewById() in Fragments

jfoerste:
## Issue

Because of #1408 and as a result of the fix #1406 the versions of the dependencies `androidx.fragment:fragment-ktx` and `androidx.appcompat:appcompat` were locked.

This updates the dependencies to the latest version and fixes the underlying issue.

The problem is that when the `NavigationManager` replaces the Fragment in the `FrameLayout` often times the views of the old `Fragment` are not destroyed yet when `onStart()` is called on the new Fragment. This behavior does not seem to be clearly documented as the documentation claimes that `replace()` behaves "essentially the same" as discrete `remove()` and `add()` operations. In this example that is evidently false as the toolbar works fine when separating the transactions.

This results in `findViewById()` often yielding the previous and invalid `Toolbar` instance when setting up the toolbar following the onStart, onResume, ... event. 

The solution is to call `findViewById()` on the new fragments root view directly which should be great for performance anyways.

During the cleanup I also fixed a small memory leak in `CalendarFragment`. I was also not sure why the title is set on every resume event so for now I resorted to setting it in `onViewCreated()` which should be the more appropriate point.


kordianbruck (approved):

kordianbruck:
I think this looks good, but I'd like Joscha to also take a look if he has some time.

joschahenningsen (approved):
Looks good!

Issue #1432, opened 2022-04-29
Frequent App Crashes on Android 12


ctsk:
## Steps to Reproduce

1. Open App and Log into TUM Online
2. Navigate to My Lectures
3. Navigate to Chat rooms

(Lots of other navigation flows produce the same crash)

## Expected Result
See the chat rooms

## Actual Result
App crashes

## Problem
Seems connected to this issue: https://github.com/emilsjolander/StickyListHeaders/issues/477

Stack Trace:
```
E/AndroidRuntime: FATAL EXCEPTION: main
    Process: de.tum.in.tumcampus, PID: 20326
    java.lang.IllegalStateException: Handling non empty state of parent class is not implemented
        at se.emilsjolander.stickylistheaders.StickyListHeadersListView.onSaveInstanceState(StickyListHeadersListView.java:1098)
        at android.view.View.dispatchSaveInstanceState(View.java:20984)
        at android.view.ViewGroup.dispatchSaveInstanceState(ViewGroup.java:3975)
        at android.view.ViewGroup.dispatchSaveInstanceState(ViewGroup.java:3981)
        at android.view.ViewGroup.dispatchSaveInstanceState(ViewGroup.java:3981)
        at android.view.ViewGroup.dispatchSaveInstanceState(ViewGroup.java:3981)
        at android.view.View.saveHierarchyState(View.java:20967)
        at androidx.fragment.app.FragmentStateManager.saveViewState(FragmentStateManager.java:466)
        at androidx.fragment.app.FragmentStateManager.saveBasicState(FragmentStateManager.java:441)
        at androidx.fragment.app.FragmentStateManager.saveState(FragmentStateManager.java:400)
        at androidx.fragment.app.FragmentStore.saveActiveFragments(FragmentStore.java:143)
        at androidx.fragment.app.FragmentManager.saveAllState(FragmentManager.java:2398)
        at androidx.fragment.app.FragmentController.saveAllState(FragmentController.java:152)
        at androidx.fragment.app.FragmentActivity.onSaveInstanceState(FragmentActivity.java:512)
        at androidx.appcompat.app.AppCompatActivity.onSaveInstanceState(AppCompatActivity.java:544)
        at android.app.Activity.performSaveInstanceState(Activity.java:2200)
        at android.app.Instrumentation.callActivityOnSaveInstanceState(Instrumentation.java:1520)
        at android.app.ActivityThread.callActivityOnSaveInstanceState(ActivityThread.java:5830)
[...]

PR #1435, opened 2022-05-02
Switch to up-to-date fork of stickylistheaders

ctsk:
## Issue

This fixes the following issue(s): #1432 

Should fix those crashes for other users of Android O. (I have no clue how widespread this issue is or whether it's only happening on my phone ......)

------
I'm unsure about possible regressions. I've used the app with the patched version of stickylistheaders for a couple of days now and it seems to be working well.


joschahenningsen (approved):
Works as expected, thanks for the fix :)

Issue #1421, opened 2022-01-11
MVV Widget TUM don't work


JunkyDeveloper:
I found that widget MVV Widget TUM but it don't works. I drop it on the home screen but it stay where for a sec and then disappear. Important after the disappearing the home screen editor close.

## Steps to Reproduce

1. Homescreen
2. Open home screen editor
2. Add Widget MVV Details
3. nothing shows

## Expected Result
Show the MVV Details on the home screen

## Actual Result
Not showing the MVV Details on the home screen

## Problem
I don't know sorry

### Environment
* Phone: OnePlus Nord
* OS version: 11 Buildnumber: AC2003_14_211014
* TUM Campus App version: v3.13
* Language: device German, App: English

CommanderStorm:
closed, due to being completed by #1421

PR #1442, opened 2022-05-10
Fix for issue #1421 MVV Widget TUM don't work

PiotrKedra:
## Issue

This fixes the following issue(s): #1421

## Description

There were two problems with the widget:
- in API 31 FLAG_UPDATE_CURRENT requires also FLAG_IMMUTABLE or FLAG_MUTABLE, so I add it
- there was a problem with the unnecessary id for toolbar in the mvv widget layout. Because of that, the toolbar was mapped to AppBar which was throwing an error, so I removed this id

Besides this when I was testing if the widget is working again, I noticed that it was not refreshing (when option auto-reload is enabled). It was because after the user changed it to auto-reload, the alarm which should refresh the widget was not set properly.


kordianbruck (approved):

kordianbruck:
You'll have to please rebase onto the latest master, in order to get the build fixed.

PiotrKedra:
I rebase it

kordianbruck (approved):
Still looks good.

PR #1443, opened 2022-05-12
open nav.tum.sexy to show location

### LINKED ISSUE #1437: Use OpenStreetMaps for Roomfinder
I have seen [this app](https://play.google.com/store/apps/details?id=de.unigoettingen.lageplan) and would like to work on this. 
My idea would be to use something like [this](https://github.com/CartoDB/mobile-sdk) to solve the problem.

kmodexc:
## Issue

This only opens nav.tum.sexy and does not have an in-app map. However i think that ok for now and solves the issue for me. 

On step for #1437


joschahenningsen:
As long as the in app roomfinder does not work this seems to be a reasonable way to go 👌
@CommanderStorm are you fine with this? Will the API URL this way?

CommanderStorm:
Obviously yes.
We should however not keep this as a permanent fix (which should also be obvious)

@kmodexc is working on that too, so I see no problem in merging this

kordianbruck:
I'd probably recommend to instead do this in an in-app webview instead of opening an external browser: https://developer.android.com/guide/webapps/webview

kmodexc:
@kordianbruck Good idea. i will do this

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/NavigatumActivity.kt
```diff
@@ -0,0 +1,21 @@
+package de.tum.`in`.tumcampusapp.component.tumui.calendar
+
+import android.annotation.SuppressLint
+import androidx.appcompat.app.AppCompatActivity
+import android.os.Bundle
+import android.webkit.WebView
+
+class NavigatumActivity : AppCompatActivity() {
```
CommanderStorm, line 8:
```suggestion
class NavigaTUMActivity : AppCompatActivity() {
```

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarDetailsFragment.kt
```diff
@@ -188,10 +187,10 @@ class CalendarDetailsFragment : RoundedBottomSheetDialogFragment() {
     }
 
     private fun onLocationClicked(location: String) {
-        val findStudyRoomIntent = Intent()
-        findStudyRoomIntent.putExtra(SearchManager.QUERY, Utils.extractRoomNumberFromLocation(location))
-        findStudyRoomIntent.setClass(requireContext(), RoomFinderActivity::class.java)
-        startActivity(findStudyRoomIntent)
+        val sendIntent: Intent = Intent(getContext(),NavigatumActivity::class.java).apply {
```
CommanderStorm, line 190:
```suggestion
        val sendIntent: Intent = Intent(getContext(),NavigaTUMActivity::class.java).apply {
```

app/src/main/AndroidManifest.xml
```diff
@@ -371,6 +371,12 @@
                 android:value=".component.ui.overview.MainActivity" />
         </activity>
 
+        <activity
+            android:name=".component.tumui.calendar.NavigatumActivity"
+            android:label="Navigatum"
```
CommanderStorm, line 376:
```suggestion
            android:name=".component.tumui.calendar.NavigaTUMActivity"
            android:label="NavigaTUM"
```

CommanderStorm (commented):
I dont have a strong oppinion on these suggestions. These are suggestions, not demands.

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/NavigatumActivity.kt
CommanderStorm: `app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/NavigatumActivity.kt` could be changed also, if you decide that these suggenstions are valid

joschahenningsen (approved):
Looks good to me. Happy to work on #1451 if no one picked it up already :)

Issue #1418, opened 2021-12-16
Logging out in dark theme causes app to crash


Liqs-v2:
## Steps to Reproduce

1. Be logged in (connected to TUMOnline)
2. Open the Settings via the drawer menu
3. Scroll down and press the **"Log out"**-Button

## Expected Result
The user is logged out, the app reloads and the user sees the updated SettingsFragment without the "Log out"-Button.

## Actual Result
The user **is logged out**, but the **app crashes while reloading**. When the app is reopened manually, the user is logged out.

### Screenrecording on an emulated Android device
![dark_theme_logout_broken](https://user-images.githubusercontent.com/9247500/146401315-8c890757-3f8e-4eaa-93af-e15dcc6f36bd.gif)

## Problem
The app crashes when users with phones in **dark theme attempt to log out**.

### Stacktrace
```
2021-12-16 16:34:19.831 15990-15990/de.tum.in.tumcampus E/AndroidRuntime: FATAL EXCEPTION: main
    Process: de.tum.in.tumcampus, PID: 15990
    java.lang.IllegalStateException: Fragment SettingsFragment{57a2e84} (f7f6f36a-f989-453a-abfd-57c5aae19571)} not attached to a context.
        at androidx.fragment.app.Fragment.requireContext(Fragment.java:805)
        at de.tum.in.tumcampusapp.component.other.settings.SettingsFragment$themeProvider$2.invoke(SettingsFragment.kt:52)
        at de.tum.in.tumcampusapp.component.other.settings.SettingsFragment$themeProvider$2.invoke(SettingsFragment.kt:52)
        at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
        at de.tum.in.tumcampusapp.component.other.settings.SettingsFragment.getThemeProvider(SettingsFragment.kt:52)
        at de.tum.in.tumcampusapp.component.other.settings.SettingsFragment.onSharedPreferenceChanged(SettingsFragment.kt:202)
        at android.app.SharedPreferencesImpl$EditorImpl.notifyListeners(SharedPreferencesImpl.java:637)
        at android.app.SharedPreferencesImpl$EditorImpl.apply(SharedPreferencesImpl.java:511)
        at androidx.preference.PreferenceManager.setNoCommit(PreferenceManager.java:513)
        at androidx.preference.PreferenceManager.inflateFromResource(PreferenceManager.java:142)
[...]

PR #1448, opened 2022-05-22
Logout crash fix

PiotrKedra:
## Issue

This fixes the following issue(s): #1418 

## Description

There was an issue that ThemeProvider was initialising before context was available, but actually, the function getTheme() that was used, does not need context. So to fix it, I moved getTheme() to a companion object.


kordianbruck (approved):
Thanks!

PR #1343, opened 2019-08-25
Add centralized search

thellmund:
## Issue

This fixes the following issue(s): #678 

## Screenshot
| Navigation drawer  | Search screen |
| ------------- | ------------- |
| Screenshot_1566729266 | Screenshot_1566729305 |

## Why this is useful for all students
- Single, easily accessible point for searching
- Avoid usage of ugly `SearchView` component

## Still to do
- [ ] List with recent search queries

Feedback welcome!

PR #1457, opened 2022-06-07
Centralized search v2

PiotrKedra:
## Issue

Fixes: #678 

## Description

More or less I want to create a centralized search, that was started in #1343 PR. Why am I doing it from scratch instead of using a branch from existing PR? Since this code was created 3 years ago I had a hard time merging it with the current master; there were too many conflicts. Therefore, I decided to start over again.

Additional to what was proposed in #1343 PR I am thinking of adding the history of searches, and maybe some tags (lecture, person, room, ...) for quicker navigation in search results.


PiotrKedra:
Search should be done. Here it the final look of it:

pic 1 pic 1
pic 1 pic 1
pic 1 pic 1

## Discussion

If anyone has other suggestions on what more can be done here, I am open to adding them.
- I was thinking of adding a "My lectures" type next to the "Lecture" type, but I am unsure if this is necessary.
- Also, now search starts when the user clicks 'enter' (this is how it worked in previous searches).  I was thinking that maybe it would be better if the search starts when the user is still typing the query, but then there will be lots of API calls to TUMOnline. My question is: is this a problem?

CommanderStorm:
I tested your code and everything works as expected.
Looks awesome.

Just one small issue I noticed:
From a development perspective it would be nice, if the enter key does also run the search, not only the search-icon on the virtual keyboard.

From a performance perspective: @joschahenningsen what can our current backend handle?
Room-search should be no real problem for this one, since searching is infrequent and the current comparison is just a query to the database.
Though, I have no idea how this looks on the Persons/Lectures side.

joschahenningsen:
I have no worries, backend wise here :)

CommanderStorm:
> I have no worries, backend wise here :)

Just to make shure, that there has not been a miscomunication.
I noticed, that my statement may have been misleading:
Seach-as-you-type (as referenced in the discussion part) is okay with you?

joschahenningsen:
> Seach-as-you-type (as referenced in the discussion part) is okay with you?

Right that should be no issue

PiotrKedra:
I have added search-as-you-type. The enter key still does not run the search, but now with search-as-you-type, I think it is not needed.

kordianbruck (approved):

Issue #1456, opened 2022-06-06
Allow to open Roomfinder location in other apps like Google Maps


PiotrKedra:
## Problem

I think it would be handy for users to allow them to open the Google Map app from Roomfinder. Why? For instance, a user could open a lecture location in Google Maps for car navigation. What is more, we could also allow users to open the location in other apps like MVV or Uber.

## Expected Result

Alongside Roomfinder Screen at the top, there is a menu where users can choose to open a location in Google Maps/MVV/Uber ...

CommanderStorm:
Related issues in another Project of ours:

https://github.com/TUM-Dev/navigatum/issues/87

MVV/Uber might be a good strech-goal for https://github.com/TUM-Dev/navigatum/issues/86

PiotrKedra:
So I guess it is better to implement it in navigatum project, rather than here?

CommanderStorm:
No, my comment was just indendet to group related issues together across different projects.
This issue right where it belongs, This should not need any support from any backend.
(Note that the campusapps are not yet using Navigatum https://github.com/TUM-Dev/Campus-Android/issues/1451)

But this feature will be done at https://nav.tum.sexy sooner, as making changes to a website is easier.
See https://github.com/TUM-Dev/navigatum/pull/136 for the discussion on what/how we want to implement this.

Btw, I saw that you have a some Kotlin experience..
Would you like to contribute on this feature? I am currently a bit strapped for time if I am honest..

PiotrKedra:
You mean #1451? I am currently working on #1457, but when I finished with this, then I might have a look at migration to NavigaTUM 🙂

Issue #1451, opened 2022-05-26
Migrate the Roomfinder to NavigaTUM


CommanderStorm:
Currently the used roomfinder has a few problems. Namely #1338, #1337, #1182  and #415 
A migration to [NavigaTUM](https://github.com/TUM-Dev/navigatum) could solve some of these problems.

The suggested plan to fix this would be:
1. reviewing and merging https://github.com/TUM-Dev/Campus-Android/pull/1443 to have an intermediate fix
2. solving #1437, to get a map to display coordinates
3. implementing the more information `nav.tum.sexy/api/get/:id` api-endpint (replacing the current endpoit)
4. implementing the search `nav.tum.sexy/api/search?q=querry` api-endpoint (replacing the current endpoit)

Optional:
5. implement the feedback api

All of these steps require the previous step to be completed, to work to the full extend possible, but dont require any of the further steps to be completed to be shipped (except, that once 3 is done, 1. will become unnessesary and we may want to remove this code).
We have renamed some rooms in the roomfinder, but those are edgecases, that I assume most of our users will not notice.

You can find our api documentation at https://editor.swagger.io/?url=https://raw.githubusercontent.com/TUM-Dev/navigatum/main/openapi.yaml
If you find any dificultys/anything is unclear/you found a bug, please feel free to open an issue.

CommanderStorm:
Sorry for the long delay:
@kmodexc is working on #1437, but this will take a bit longer, since we are currently in the middle of the semester and stuff is a bit stressful currently. #1437 is a bigger issue than initial anticipated 😉

Help with 3,4 would be greatly appreciated. ❤️ 

There are two stopgap solutions until #1437 is properly done I can imagine (one is more work, one is less), since until then we won't have an interactive map.
- implement the "Site Maps"/"Lagepläne" as seen on https://nav.tum.sexy/room/5407.01.760D
- just implement 4. and redirect to nav.tum.sexy via #1443

Issue #1338, opened 2019-07-26
Room maps not loading


jacqueline8711:
Maps are loaded ("Sorry, something went wrong")
Tested on current develop and last release. 

Example call [https://tumcabe.in.tum.de/Api/roomfinder/room/availableMaps/101%25405620](https://tumcabe.in.tum.de/Api/roomfinder/room/availableMaps/101%25405620)
```
Warning: file_get_contents(): SSL operation failed with code 1. OpenSSL Error messages: error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed in /var/www/tca-portal/App/Functions.php on line 119

Warning: file_get_contents(): Failed to enable crypto in /var/www/tca-portal/App/Functions.php on line 119

Warning: file_get_contents(http://portal.mytum.de/campus/roomfinder/): failed to open stream: operation failed in /var/www/tca-portal/App/Functions.php on line 119

Notice: Undefined offset: 1 in /var/www/tca-portal/App/Modules/Api/Room.php on line 165

Notice: Undefined index: ze in /var/www/tca-portal/App/Modules/Api/Room.php on line 232
Error: empty URL for xml rpc call supplied
```

jacqueline8711:
I’m quite sure this worked when we introduced Picasso. I don’t know the
file type though. Since it’s most likely the backend I’ll try to fix it
before the semester starts.

CommanderStorm:
Works on Older Android Versions

- [Android 5.1.1, App-Version 3.4, Huawei Y6, no VPN/...], where it DOES work
Proof:
![WhatsApp Image 2019-10-23 at 10 16 20](https://user-images.githubusercontent.com/26258709/67381524-54e48600-f58c-11e9-9c5b-f9c6147f1c43.jpeg)


- [Android 9, App-Versin 3.4, Samsung S8, no VPN/...] and here it does NOT work
Proof:
![Screenshot_20191023-120040_Campus](https://user-images.githubusercontent.com/26258709/67381922-f2d85080-f58c-11e9-8354-c30cd5a80fc8.jpg)

julianostarek:
@CommanderStorm Can't confirm. Room 02.13.007, too, fails for me on a SDK 22 emulator. 

I've spent some time diving into the thrown exceptions and it seems there's something funky about one of the servers' SSL certificates. The SSL handshake fails with the following (standard) exception:

[...]

Issue #1182, opened 2018-10-30
Roomfinder shows wrong locations


xsrf:
## Steps to Reproduce

1. Search for a room, e.g. "0.01.24"
2. Search for the same room in TUMonline and click on the location
3. compare locations shown on the picture

## Expected Result

locations should match

## Actual Result

locations don't match

## Problem

locations don't match

## Suggested Enhancement :sparkles:

fix it ;) I guess TCA only shows the location of the building

### Environment
* Phone: Nexus 5X
* OS version: ...
* TUM Campus App version: 2.3
* Language: de

xsrf:
Buidl

![2018-10-29 21 57 06](https://user-images.githubusercontent.com/8081353/47725409-ce3ae880-dc58-11e8-8ed6-6297a1c9bc06.png)

Honeybyte:
It seems like that the bug is in the the server which is redirecting to the actual map URL. Searched for room 1200B:

Result in version 2.1-dev
`ImageViewTouchFragment: room finder url: https://tumcabe.in.tum.de/Api/roomfinder/room/defaultMap/1200B@0502`
After switching maps
`ImageViewTouchFragment: room finder url: https://tumcabe.in.tum.de/Api/roomfinder/room/map/1200B@0502/12`

Result in recent version 3.2-dev
`ImageViewTouchFragment: room finder url: https://tumcabe.in.tum.de/Api/roomfinder/room/defaultMap/1200B@0502`
Switching maps not possible in recent version as menu is not shown (intended?)

Both URLs (…/room/defaultMap/… and  …/room/map/…) are redirecting to the same URL https://portal.mytum.de/campus/roomfinder/getBuildingMap?b_id=0502&mapid=12
But the non default one should redirect to https://portal.mytum.de/campus/roomfinder/getRoomMap?mapid=12&roomid=1200B@0502

kordianbruck:
@Honeybyte thanks for pointing this out. That the map selection is not visible is certainly a separate issue. I'll follow up once I had time to check out the backend and data stored there.

PR #1462, opened 2022-07-04
Migration to Navigatum

PiotrKedra:
## Issue

- Resolves: #1451
- Resolves: #1456 
- Resolves: #1338
- Resolves: #415
- Resolves: #1182

Depending on merging centralized search #1457, I decided to work on this PR rather than the master branch because I used some room finder logic there.

## Description
Migrate to:
- more information nav.tum.sexy/api/get/:id api-endpint 
- the search nav.tum.sexy/api/search?q=querry api-endpoint


PiotrKedra:
Migration more or less should be ready. I have migrated three roomfinder endpoints: 
- fetchAvailableMaps()
- fetchRooms()
- fetchCoordinates()

Into two NavigaTum endpoints (search navigation and navigation details)

### What is still to be done?
1. remove old roomfinder logic (didn't remove it yet because of the below problems)
2. besides migration I want to slightly improve the navigation details UI (now it only displays sitemap plan), to sth like on the NavigaTum website. Also, I want to allow to open location in other apps #1456.

### Problems/questions
1. I heard that NavigaTum is not stable yet, therefore can it be merged now?
2. Old roomfinder API has an endpoint that fetches RoomFinderSchedule. However, I couldn't make this code run in the App, so I am not sure if this is still supported or it can be removed?
3. Barrier-free API fetches rooms that have ID from the old roomfinder API and using NavigaTum with this ID does not work. Therefore, now when users click on the Barrier-free facility, the app redirects to the search screen with this facility address as a query input. Is this solution ok?

CommanderStorm:
> 1. I heard that NavigaTum is not stable yet, therefore can it be merged now?
If we comment on https://github.com/TUM-Dev/navigatum/issues/130 about the change timeline, it is stable ;)

> 2. Old roomfinder API has an endpoint that fetches RoomFinderSchedule. However, I couldn't make this code run in the App, so I am not sure if this is still supported or it can be removed?
@joschahenningsen historically What is the status on this? Was this ever used in the app?

> 3. Barrier-free API fetches rooms that have ID from the old roomfinder API and using NavigaTum with this ID does not work. Therefore, now when users click on the Barrier-free facility, the app redirects to the search screen with this facility address as a query input. Is this solution ok?

@joschahenningsen Can we migrate the backend (add a new field?)

joschahenningsen:
> @joschahenningsen historically What is the status on this? Was this ever used in the app?

I think this might be used in the calendar somewhere. I'll need to do some more digging though.

> @joschahenningsen Can we migrate the backend (add a new field?)

Sure, as long as it does not break old clients :) Do you have a way of mapping those ids? 

On another note: Do y'all need some help with the merge conflicts? :D

PiotrKedra:
## RoomFinderSchedule

I did some digging into this secret RoomFinderSchedule logic ;p
> 2. Old roomfinder API has an endpoint that fetches RoomFinderSchedule. However, I couldn't make this code run in the App, so I am not sure if this is still supported or it can be removed?

I find out that this logic was not available to execute after these two commits:
- [moved to Picasso on Apr 27, 2018](https://github.com/TUM-Dev/Campus-Android/commit/f03650e9654ea677123392ca28b444e5dca78a60#diff-c3a8aa040517eb62a994a98d9e6cb86d4aeb851a1bb64162c948d65f6be3d065)
- [Update various dependencies on Oct 13, 2020](https://github.com/TUM-Dev/Campus-Android/commit/5236de171266a6b79b5c24ce68fdbcfe12bf19a5#diff-1c4f58e643463b7977c699461a8ae3db68743711905b3fa0ad1f3037d4b3fe4c)

In both these commits, it looks like this might be removed accidentally, but cannot be sure. However, it looks like, since April 2018 this RoomFinderSchedule feature does not work, so I think it would be best to deal with this problem in a separate issue.  Also since it was not available for so long, there is a question if anybody needs it.

## Barrier-free API

Another case is what shall we do with this Barrier-free API
> 3. Barrier-free API fetches rooms that have ID from the old roomfinder API and using NavigaTum with this ID does not work. Therefore, now when users click on the Barrier-free facility, the app redirects to the search screen with this facility address as a query input. Is this solution ok?

Should we wait for https://github.com/TUM-Dev/Campus-Backend/pull/80 to be merged and ready to use, so it can be implemented here as well? Or we just merge this PR now with my temporary solutions and take care of Barrier-free in another one? @CommanderStorm

CommanderStorm:
> Should we wait for [TUM-Dev/Campus-Backend#80](https://github.com/TUM-Dev/Campus-Backend/pull/80) to be merged and ready to use, so it can be implemented here as well? Or we just merge this PR now with my temporary solutions and take care of Barrier-free in another one? @CommanderStorm

Actually, we don't need to wait for that PR, as the current backend does already provide us with the exact information we need:
image
When looking in the network tab, `room_code` is the ID, NavigaTUM uses, so probably we can simplify this UX a bit.  
The study rooms also one place where we already have this field. (it is named `code` in the data class)
Thus, we can redirect directly to showing the room.

Doing so should cover all cases, since [the fixes we added](https://github.com/TUM-Dev/navigatum/blob/main/data/sources/15_patches-rooms_tumonline.yaml) for invalid `room_code`s do not relate to study rooms or barierfree. Redirect to the details page for the search results sould be save here..

However, if you like, we can use your solution and migrate all occurrences of the `room_id` to `room_code` in a later PR…

PiotrKedra:
Thanks for finding the room_code parameter @CommanderStorm. I have implemented it and it works as expected. I have also removed not used room finder code and merged master. There were lots of conflicts, but after merging all seems to work (I hope 😅).

From my side, I think PR is ready to be merged.  I will create a separate issue for the RoomFinderSchedule feature as described earlier.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/search/adapter/RecentSearchesAdapter.kt
```diff
@@ -60,25 +60,36 @@ class RecentSearchesAdapter(
...
+                RecentsDao.NAVIGA_TUM_ROOMS -> {
+                    try {
+                        val navigationEntity = NavigationEntity.fromRecent(recentSearch)
+                        titleText.text = navigationEntity.getFormattedName()
+                    } catch (exception: Exception) {
+                        titleText.setText(R.string.not_available_search)
+                    }
+                    icon.setImageResource(R.drawable.ic_room)
                     icon.setBackgroundResource(R.drawable.search_result_icon_background)
                     profilePicture.setImageDrawable(null)
```
CommanderStorm, line 94: Since this is 2x basically the same code except for the setImageResource calls, we should simplify this and have an interiour when clause for this

app/src/main/java/de/tum/in/tumcampusapp/component/ui/search/SearchFragment.kt
```diff
@@ -135,8 +140,9 @@ class SearchFragment : BaseFragment<Unit>(
         try {
             when (recent.type) {
                 RecentsDao.PERSONS -> openPersonDetails(Person.fromRecent(recent))
-                RecentsDao.ROOMS -> openRoomDetails(RoomFinderRoom.fromRecent(recent))
                 RecentsDao.LECTURES -> openLectureDetails(Lecture.fromRecent(recent))
+                RecentsDao.NAVIGA_TUM_BUILDINGS -> openNavigationDetails(NavigationEntity.fromRecent(recent))
+                RecentsDao.NAVIGA_TUM_ROOMS -> openNavigationDetails(NavigationEntity.fromRecent(recent))
```
CommanderStorm, line 145:
```suggestion
                RecentsDao.NAVIGA_TUM_BUILDINGS, RecentsDao.NAVIGA_TUM_ROOMS -> openNavigationDetails(NavigationEntity.fromRecent(recent))
```

CommanderStorm, line 145:
```suggestion
                RecentsDao.NAVIGA_TUM_BUILDINGS, RecentsDao.NAVIGA_TUM_ROOMS -> openNavigationDetails(NavigationEntity.fromRecent(recent))
```

app/src/main/res/values/strings.xml
```diff
@@ -771,18 +771,35 @@ Signature: %5$s</string>
...
     <string name="clear_search_history_request">Do you want to clear search history?</string>
 
+    // Navigation details
+    <string name="location_details">Location details</string>
+    <string name="loading_details">loading</string>
+    <string name="site_plans">Site plans</string>
+    <string name="interactive_map">Interactive map</string>
+    <string name="open_location_in_other_app">Open location in other app</string>
+    <string name="available_maps">Available site plans:</string>
+    <string name="map">Map</string>
```
CommanderStorm, line 795: map does not seem to be used anywhere else. Is this intentional?

app/src/main/res/values-de/strings.xml
```diff
@@ -719,14 +719,32 @@ Signatur: %5$s</string>
...
     <string name="clear_search_history_request">Möchten Sie den Suchverlauf löschen?</string>
 
+    // Navigation details
+    <string name="location_details">Location details</string>
+    <string name="loading_details">loading</string>
+    <string name="site_plans">Site plans</string>
+    <string name="interactive_map">Interactive map</string>
+    <string name="open_location_in_other_app">Open location in other app</string>
+    <string name="available_maps">Available maps:</string>
+    <string name="map">Map</string>
```
CommanderStorm, line 740:
```suggestion
    <string name="location_details">Details zum Ort</string>
    <string name="loading_details">lädt</string>
    <string name="site_plans">Lagepläne</string>
    <string name="interactive_map">Interaktive Karte</string>
    <string name="open_location_in_other_app">Ort in anderer Anwendung öffnen</string>
    <string name="available_maps">Verfügbare Karten:</string>
    <string name="map">Karte</string>
```

app/src/main/java/de/tum/in/tumcampusapp/api/app/TUMCabeClient.java
```diff
@@ -280,24 +279,12 @@ public List<BuildingToGps> getBuilding2Gps() throws IOException {
                 .body();
     }
 
+    @Deprecated // use NavigaTUM api instead
```
CommanderStorm, line 282:
```suggestion
    @Deprecated // This API has been deprecated. Use the equivalent NavigaTUM API instead
```

CommanderStorm (approved):
Thanks for all of this work. 😄 
I have commented on some small things inline. Once these are resolved, IMO we can merge this. 

Given the size of this PR, I would like somebody else to double-check me @Bentipa, would you have the time to look through this PR?

CommanderStorm (approved):
Thanks for all of this work. 😄 
I have commented on some small things inline. Once these are resolved, IMO we can merge this. 

Given the size of this PR, I would like somebody else to double-check me @Bentipa, would you have the time to look through this PR?

Bentipa:
> Thanks for all of this work. 😄 
> 
> I have commented on some small things inline. Once these are resolved, IMO we can merge this. 
> 
> 
> 
> Given the size of this PR, I would like somebody else to double-check me @Bentipa, would you have the time to look through this PR?

Hey,

as you may have noticed I did not have time yet to also answer/review the other things as I am sick at the moment. Although I will take a look once I am back, if you need a review for this one faster we should maybe ask someone else.

app/src/main/java/de/tum/in/tumcampusapp/api/navigatum/domain/NavigationEntity.kt
```diff
@@ -0,0 +1,54 @@
...
+
+    fun getFormattedName(): String {
+        return removeHighlight(name)
+    }
+
+    fun getFormattedSubtext(): String {
+        return removeHighlight(subtext)
+    }
+
+    private fun removeHighlight(field: String): String {
```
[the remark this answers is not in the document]
PiotrKedra: Actually, I haven't thought of it much. Mb one concern is that in centralized search there are ppl and lectures fetched as well and they do not have any highlights, but I am not sure if this is wrong. I think we can add highlights it just needs additional logic that handles it.

app/src/main/java/de/tum/in/tumcampusapp/api/navigatum/model/details/RoomFinderMapDto.kt
```diff
@@ -0,0 +1,15 @@
+package de.tum.`in`.tumcampusapp.api.navigatum.model.details
+
+import com.google.gson.annotations.SerializedName
+
+data class RoomFinderMapDto(
+    var id: String = "",
+    var name: String = "",
+    @SerializedName("file")
```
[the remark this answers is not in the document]
PiotrKedra: I think it can stay like it is. Didn't know that there are multiple sizes ;d

app/src/main/java/de/tum/in/tumcampusapp/component/other/locations/LocationManager.kt
```diff
@@ -311,16 +262,20 @@ class LocationManager @Inject constructor(c: Context) {
         var loc = roomTitle
         if (loc.contains("(")) {
             loc = loc.substring(0, loc.indexOf('('))
-                    .trim { it <= ' ' }
+                .trim { it <= ' ' }
```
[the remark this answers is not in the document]
PiotrKedra: Yeah probably u are right, but I think I didn't touch these lines, they were the same before ;p

app/src/main/res/layout/toolbar_search.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<com.google.android.material.appbar.AppBarLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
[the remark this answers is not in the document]
PiotrKedra: We don't handle these gestures here I guess, but if we maybe want in future, then I think your way is better. And yeah, I did it due to runtime error, didn't come up with your solution. I guess I need to improve my skills ;D

app/src/main/res/values/strings.xml
PiotrKedra: My mistake, I should have remove it

PiotrKedra:
Thanks for the extensive review :) WIth most of the comments I agree. I think it would be better If I aggregate these changes into one commit, rather than one by one by clicking here on GitHub (there will be long history ;d)? But I don't know what is better, I see these commit suggestions for the first time ^^

When it comes to ordering since there are four different objects fetched in a centralized search. The "All results" sort alphabetically all of them, but when you choose a specific object type, then the ordering is as it was fetched from APIs. I added sorting because I thought it was slightly better to mix the results alphabetically than display groups of objects one by one (firstly all people, then all lectures, then all buildings and rooms). This was my reasoning, what do you think? :D

CommanderStorm:
> Thanks for the extensive review :) WIth most of the comments I agree. I think it would be better If I aggregate these changes into one commit, rather than one by one by clicking here on GitHub (there will be long history ;d)? But I don't know what is better, I see these commit suggestions for the first time ^^

What works for you…
You don't have to worry about a long history, since all changes are squashed into one commit anyway 😉 

> When it comes to ordering since there are four different objects fetched in a centralized search. The "All results" sort alphabetically all of them, but when you choose a specific object type, then the ordering is as it was fetched from APIs. I added sorting because I thought it was slightly better to mix the results alphabetically than display groups of objects one by one (firstly all people, then all lectures, then all buildings and rooms). This was my reasoning, what do you think? :D

For rooms/buildings, this definitively does not make sense, as we have a priority between our items.
For people and lectures I dont know this for certain, but I assume they are also sorted somehow.

We should therefore just drop this sorting. Finding a total order between these items is not trivial and should be its PR

From an ordering, I would therefore prefer:
- people
- building/room (in the order we communicate if possible)
- lectures

Assumptions being,
- that there are few human names that map intersect with buildings/rooms and therefore this can be first without causing much of a bad impact.
- Searching for your lectures is something you rarely do in a campus app

PiotrKedra:
Ok, I will fix the ordering as u suggest, but probably I will do it next week after my exams

Bentipa:
I will take another look once @PiotrKedra has implemented the changes, just tag me again then please :)

app/src/main/java/de/tum/in/tumcampusapp/api/navigatum/domain/NavigationEntity.kt
```diff
@@ -0,0 +1,54 @@
...
+    var type: String = "",
+    @SerializedName("name")
+    var name: String = "",
+    @SerializedName("subtext")
+    var subtext: String = ""
+) : Serializable {
+
+    fun getFormattedName(): String {
+        return removeHighlight(name)
+    }
```
octycs, line 21:
```suggestion
    @SerializedName("subtext")
    var subtext: String = "",
    @SerializedName("parsed_id")
    var parsedId: String? = null,
) : Serializable {

    fun getFormattedName(): String {
        return if (parsedId === null) {
            removeHighlight(name)
        } else {
            removeHighlight(parsedId!!) + " ➤ " + removeHighlight(name)
        }
    }
```

I think it would be helpful to add the `parsed_id` field as well. This way it would look like this:



We could also use a different arrow symbol, e.g. `»`:



If you think this makes the search results too long, I would suggest showing only the `parsed_id`, because it's only included if we consider it more helpful than the normal name.

(You can also change the sources, I've never written Kotlin so far)

app/src/main/res/values/strings.xml
PiotrKedra: Ah, it is used in fragment_navigation_details spinner ^^

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/roomfinder/NavigationDetailsFragment.kt
```diff
@@ -0,0 +1,211 @@
...
+    private fun showAvailableMaps(navigationDetails: NavigationDetails) {
+        val availableMaps = navigationDetails.availableMaps
+        if (availableMaps.isNotEmpty()) {
+            val map = availableMaps[0]
+            loadMapImage(map)
+
+            val adapter = ArrayAdapter<String>(requireContext(), android.R.layout.simple_spinner_item)
+            adapter.setDropDownViewResource(android.R.layout.simple_spinner_dropdown_item)
+            binding.availableMapSpinner.adapter = adapter
+            availableMaps.forEach {
```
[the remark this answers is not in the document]
PiotrKedra: Spinner shows when the details object is being loaded. There is no spinner for loading the map itself.

app/src/main/java/de/tum/in/tumcampusapp/api/navigatum/domain/NavigationEntity.kt
PiotrKedra: I have added ➤ option

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/roomfinder/NavigationDetailsFragment.kt
```diff
@@ -0,0 +1,211 @@
...
+                .into(binding.photoView)
+        }
+
+        binding.interactiveMapBtn.setOnClickListener {
+            showInteractiveMapDialog()
+        }
+    }
+
+    private fun loadMapImage(map: NavigationMap) {
+        val pinDrawable = ResourcesCompat.getDrawable(resources, R.drawable.ic_place, null)
```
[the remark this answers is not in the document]
PiotrKedra: I have changed it to tum_orange

PiotrKedra:
I have implemented changes, and also left comments on some conversations
@Bentipa

CommanderStorm (approved):
imo this can be merged. All my major comments have been adressed.
@Bentipa do you want to review it, or should I merge it?

CommanderStorm:
@PiotrKedra Thanks for the wonderful work you did ✨

Bentipa:
@CommanderStorm Just gave it a quick look cause of time reasons, but looks good overall to me. So we can merge it.

PR #1463, opened 2022-07-04
Miscelaneous dependency upgrades

CommanderStorm:
## Issue

While looking through the build I noticed, that we have quite a lot of deprecated/outdated things on board.

This issue tries to update the dependencys, where it makes sense to do so in a cumulative PR (i.e. only the trivial stuff).

I decided to merge these changes into one issue, because I think that it is more reviewable this way.


kordianbruck (approved):
Really hopes this all works :P

PR #1466, opened 2022-07-05
Migration of `kotlin-android-extensions`

### LINKED ISSUE #1445: kotlin-android-extensions is deprecated
This warning appears in the build. Does anyone know what this means and how to fix it?

```
The 'kotlin-android-extensions' Gradle plugin is deprecated. Please use this migration guide (https://goo.gle/kotlin-android-extensions-deprecation) to start working with View Binding (https://developer.android.com/topic/libraries/view-binding) and the 'kotlin-parcelize' plugin.
```

CommanderStorm:
## Issue
This migrates the `kotlin-android-extensions` Gradle plugin.

Changes:
- `Parcelisise`-ation now comes from the `kotlin-parcelise` package
- we no longer use synthetics

Current Blocker:
we included in #1417 the following block:
```gradle
androidExtensions {
    experimental = true
}
```
@Liqs-v2 what is the reasoning behind this?
Is this still nessesary?
(there is a gradle bug caused by this that I can not, for the live of me, debug..)

This fixes the following issue(s): #1445 

The warning is confusing for new developers and should be squashed

resolves #1445

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -73,9 +73,6 @@ android {
     sourceSets {
         androidTest.assets.srcDirs += files("$projectDir/schemas".toString())
     }
-    androidExtensions {
-        experimental = true
-    }
     lint {
         abortOnError false
         checkAllWarnings true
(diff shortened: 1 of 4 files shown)


Liqs-v2:
It's been to long to remember this off of the top of my head. It might just be debugging residue. What happens if u disable it?

CommanderStorm:
> What happens if u disable it?

The build does not fail in the first step with some really undebuggable error

PR #1467, opened 2022-07-06
Consistent dialogue boxes

### LINKED ISSUE #1465: Inconsistent Dialogue box Design
## Steps to Reproduce

1.Open Dialogue boxes

## Expected Result
A consistent design

## Actual Result
Screenshot_20220704_013222
Screenshot_20220704_020038

## Problem
Both designs are problematic:
1. has bad accessibility
2. looks bad due to missing margins

I think 1. with our primary color would be ideal

CommanderStorm:
## Issue
Currently the dialogs are less consistent and less accessible than they could be. This PR fixes this:

- [x] removal of unnessesary code
- [x] stylistic unification of buttons
- [x] stylistic unification of dialogs

Accessible and consistent dialogs

resolves #1465


app/src/main/java/de/tum/in/tumcampusapp/component/other/settings/SettingsFragment.kt
```diff
@@ -276,15 +276,11 @@ class SettingsFragment : PreferenceFragmentCompat(), Preference.OnPreferenceClic
 
     private fun showLogoutDialog(title: Int, message: Int) {
         AlertDialog.Builder(requireContext())
-                .setTitle(title)
-                .setMessage(message)
-                .setPositiveButton(R.string.logout) { _, _ -> logout() }
-                .setNegativeButton(R.string.cancel, null)
-                .create()
-                .apply {
-                    window?.setBackgroundDrawableResource(R.drawable.rounded_corners_background)
```
kordianbruck, line 285: No more rounded corners? Sad!

kordianbruck (approved):

CommanderStorm:
The need for `setBackgroundColor()` here is because the settings seem to not want to cooperate with this.. Any ideas?

I am pretty unshure about the colors.. we should probably change those.. any ideas?

app/src/main/java/de/tum/in/tumcampusapp/component/other/settings/SettingsFragment.kt
CommanderStorm: No, they are still included, but now on every popup

kordianbruck (approved):

PR #1468, opened 2022-07-06
Work through some of our Deprecations

CommanderStorm:
## Issue

This issue works through some of our deprecations. This was sparked by the ammount of deprecations noticed while doing #1463


CommanderStorm:
I know that these are not even close to all of our deprecations. I will create follow-up PRs, tackeling the rest but for now I think getting this reviewed is the better step

kordianbruck (approved):

PR #1471, opened 2022-07-07
Fix the current seeing position in calendar

### LINKED ISSUE #1469: Fix the current seeing position in Calendar
## Problem
When "show canceled event" check box is changed, or when button for weekly-daily calendar is clicked, current position of calendar always go to today. (The most left side)

I want the keep the currently seeing position in calendar even though options are changed.

yujinkimmn:
## Issue & Why this is useful for all students

The position in the calendar always go to the current Date and Time. But it will be more useful to keep the current seeing position in the calendar even when the options are changed to check the schedule of the date that user want to see. 

## Screenshot
This is the changes.

First screen when user go to "Calendar" category, it is same as before.

When user scroll the calendar to the date when user want to check, it will be like this(same as before so far)

If user click the button for changing to daily calendar from weekly one, the first seeing date(7/12) is maintained.

In addition, when user want to check canceled events on specific date and change the check box "show canceled events", the current seeing date will also be kept. 

Fixes #1469

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarFragment.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarFragment.kt
@@ -438,7 +438,13 @@ class CalendarFragment : FragmentForAccessingTumOnline<EventsResponse>(
         showDate?.let {
             binding.weekView.goToDate(it.toGregorianCalendar())
             binding.weekView.goToHour(it.hourOfDay)
-        } ?: binding.weekView.goToCurrentTime()
+        } ?: run {
+            binding.weekView.firstVisibleDate?.let{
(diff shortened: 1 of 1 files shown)


Bentipa:
Hey @yujinkimmn thanks for the contribution!
Works perfectly, I am only awaiting @CommanderStorm view on it then I guess we should be able to merge it :)

yujinkimmn:
> I think from a readability side, this tripple-elvis is really hard to parse for humans..
> 
> I have experimented with other ways to write this code, but they also look horrible, so I guess this is the best way I can think of..
> 
> If you (@yujinkimmn) can think of a better solution that would be preferable, but if not we can merge this as is.

Sorry:( I don't have better solution now. If I come up with, i could make another issue.

kordianbruck:
Thanks :beers:

Issue #1473, opened 2022-07-10
Barrier-free facilities are not loaded


PiotrKedra:
## Steps to Reproduce

1. Open Menu
2. Click Barrier-free Information
3. Chose Facilities

## Expected Result

The app shows a list of barrier-free facilities.

## Actual Result

The app does not load barrier-free facilities and a blank screen is displayed. Sometimes instead of a blank screen "Unknow error" was shown with the "Retry" button, and then facilities get "sometimes" loaded. Idk why it has such strange behaviour.

## Problem

On the current master branch, the app does not display the list of barrier-free facilities and the below error appears in the logs. However, the current version of the app 3.18 that is on the Googe Play Store sometimes works, but not always ... ;D

## Logs

W/System.err: android.view.ViewRootImpl$CalledFromWrongThreadException: Only the original thread that created a view hierarchy can touch its views.
W/System.err:     at android.view.ViewRootImpl.checkThread(ViewRootImpl.java:9806)
W/System.err:     at android.view.ViewRootImpl.requestLayout(ViewRootImpl.java:1890)
W/System.err:     at android.view.ViewExtImpl.hookRequestLayout(ViewExtImpl.java:509)
W/System.err:     at android.view.View.requestLayout(View.java:25905)
...

### Environment
* Phone: One Plus 9 Pro
* OS version: Android 12
* Language: English

PiotrKedra:
Yeah, I just reported it, feel free to take it ^^

Bentipa:
Closed via #1474

PR #1474, opened 2022-07-12
Fix Bug in Barrier Activity + Add back button to toolbar

Bentipa:
Fixed wrong thread ui error

## Issue
Barrier free facilities were not loaded.
(+ No Back Button in Activity)

#1473 

Back-Button: Improved UX
Bugfix: Usability restored :D

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/ui/barrierfree/BarrierFreeFacilitiesActivity.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/ui/barrierfree/BarrierFreeFacilitiesActivity.kt
@@ -4,6 +4,8 @@ import android.content.Intent
 import android.os.Bundle
 import android.view.View
 import android.widget.AdapterView
+import androidx.appcompat.widget.Toolbar
+import androidx.core.app.NavUtils
 import de.tum.`in`.tumcampusapp.R
 import de.tum.`in`.tumcampusapp.component.other.general.RecentsDao
 import de.tum.`in`.tumcampusapp.component.other.general.model.Recent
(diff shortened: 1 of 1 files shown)


app/src/main/java/de/tum/in/tumcampusapp/component/ui/barrierfree/BarrierFreeFacilitiesActivity.kt
```diff
@@ -36,6 +38,16 @@ class BarrierFreeFacilitiesActivity : ActivityForAccessingTumCabe<List<RoomFinde
...
 
+        val toolbar = findViewById<Toolbar>(R.id.toolbar)
+        setSupportActionBar(toolbar)
+        supportActionBar?.let {
+            val parent = NavUtils.getParentActivityName(this)
+            if (parent != null) {
+                it.setDisplayHomeAsUpEnabled(true)
+                it.setHomeButtonEnabled(true)
+            }
+        }
```
kordianbruck, line 49: Can you please add a comment why this is needed? It seems like "busy code" that we used to have copied into every and all activities. Did you follow some documentation that points to this being required?

Bentipa: Originally, this code is in the BaseActivity class, which therefore should normally inherited by the BarrierFreeFacilitiesActivity.
I debugged it for ~1hour and tested every possible change of layout files / etc..
The code in BaseActivity is run, though the back button is not enabled. I guess this is because the layout is inflated inside of the BarrierFreeFacilitiesActivity after super.onCreate.
I can check again but this was the only workaround getting the back button to work.

CommanderStorm (approved):
Works as expected. I dont like the code duplicatio between `BarrierFreeFacilitiesActivity` and `BaseActivity`, but this probably cannot be avoided

Bentipa:
@kordianbruck @CommanderStorm 
I found a solution which does not require the code duplication anymore.
In addition, this PR now also closes #1482 which I just found during testing. :)

Issue #1433, opened 2022-04-29
Navigation drawer is blank in some Activities


ctsk:
## Steps to Reproduce

1. Open App & Connect to TUMOnline
2. Navigate to Chat rooms
3. Open any chat room
4. Swipe from left to open the navigation drawer

----

1. Open App
2. Open TU Film card
3. Swipe from left to open the navigation drawer

## Expected Result
See Navigation

## Actual Result
Drawer is blank

<details>
<summary>Screenshot</summary>

![scrot-drawer](https://user-images.githubusercontent.com/9384305/165911695-edcd65c4-cb4b-4fe6-b6b4-d177d19c79af.png)

</details>

### Environment
* Phone:  Pixel
* OS version: 12
* TUM Campus App version: v3.13
* Language:  English

Bentipa:
I guess the correct solution here would be to completely disable the drawer when being inside such a child view.

PR #1475, opened 2022-07-12
Removed drawer where not necessary

Bentipa:
## Issue
Empty drawer could be opened from certain views

#1433 

## Screenshot
(No more drawer from the left possible)

Bugfix

### DIFF
--- a/app/src/main/res/layout/activity_chat.xml
+++ b/app/src/main/res/layout/activity_chat.xml
@@ -74,6 +74,4 @@
 
     </LinearLayout>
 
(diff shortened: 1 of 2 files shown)


kordianbruck:
Shouldn't we just populate the drawer properly in those views? It would make sense to have it so users can navigate faster.

Bentipa:
@kordianbruck It does not really make sense in those detail views, as the top left icon is a back button and not the drawer menu icon. It would not be intuitive that are drawer exists there.

Issue #1480, opened 2022-07-18
Migrate new central Search to ViewBindings


CommanderStorm:
## Steps to Reproduce

1. Click on 
![image](https://user-images.githubusercontent.com/26258709/179557887-8469f102-0214-4b03-8e4e-c1019f1e150e.png)

2. Currently the Search view fails on opening

## Expected Result
The View should lod

## Actual Result

<details>

  ```
  2022-07-18 18:17:36.856 5170-5170/de.tum.in.tumcampus E/AndroidRuntime: FATAL EXCEPTION: main
      Process: de.tum.in.tumcampus, PID: 5170
      java.lang.ClassCastException: android.widget.LinearLayout cannot be cast to com.google.android.material.appbar.AppBarLayout
          at de.tum.in.tumcampusapp.databinding.ToolbarSearchBinding.bind(ToolbarSearchBinding.java:86)
          at de.tum.in.tumcampusapp.component.ui.search.SearchFragment$bindingToolbar$2.invoke(SearchFragment.kt:54)
          at de.tum.in.tumcampusapp.component.ui.search.SearchFragment$bindingToolbar$2.invoke(SearchFragment.kt:54)
          at com.zhuinden.fragmentviewbindingdelegatekt.FragmentViewBindingDelegate.getValue(FragmentViewBindingDelegate.kt:68)
          at de.tum.in.tumcampusapp.component.ui.search.SearchFragment.getBindingToolbar(SearchFragment.kt:54)
          at de.tum.in.tumcampusapp.component.ui.search.SearchFragment.access$getBindingToolbar(SearchFragment.kt:45)
          at de.tum.in.tumcampusapp.component.ui.search.SearchFragment$handleStateChange$2.emit(SearchFragment.kt:297)
          at de.tum.in.tumcampusapp.component.ui.search.SearchFragment$handleStateChange$2.emit(SearchFragment.kt:287)
          at kotlinx.coroutines.flow.StateFlowImpl.collect(StateFlow.kt:398)
          at de.tum.in.tumcampusapp.component.ui.search.SearchFragment.handleStateChange(SearchFragment.kt:287)
          at de.tum.in.tumcampusapp.component.ui.search.SearchFragment.access$handleStateChange(SearchFragment.kt:45)
          at de.tum.in.tumcampusapp.component.ui.search.SearchFragment$onViewCreated$1.invokeSuspend(SearchFragment.kt:93)
[...]

PR #1481, opened 2022-07-18
Fixed incorrect migration

CommanderStorm:
## Issue

Resolves: #1480 

## Screenshot
image

Idk, less bugs and stuff

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/ui/search/SearchFragment.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/ui/search/SearchFragment.kt
@@ -51,7 +51,6 @@ class SearchFragment : BaseFragment<Unit>(
     lateinit var viewModelProvider: Provider<SearchViewModel>
 
     private val binding by viewBinding(FragmentSearchBinding::bind)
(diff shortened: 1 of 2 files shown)


Bentipa (approved):
The bug is fixed, although I am not sure if there would also be a way to get that view binding to work without putting the included layout directly into the main layout file.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/search/SearchFragment.kt
```diff
@@ -4,7 +4,9 @@ import android.app.SearchManager
 import android.content.Context
 import android.content.Intent
 import android.os.Bundle
+import android.view.LayoutInflater
 import android.view.View
+import android.view.ViewGroup
```
CommanderStorm, line 9:
```suggestion
import android.view.View
```

CommanderStorm, line 9:
```suggestion
import android.view.View
```

Bentipa:
@CommanderStorm yeah looks good :D

Issue #1476, opened 2022-07-12
Enhance UX for TUFilm ViewPager


Bentipa:
## Steps to Reproduce

1. Open TU-Film Card
2. Do not recognize you can swipe left & right

## Problem
It might not be intuitive for the users that you can swipe to the left and right to switch between movies

## Suggested Enhancement :sparkles:
Either we add a preview of the next and last item:
<img src="https://user-images.githubusercontent.com/11789478/178519393-55d16ee7-64f8-4d12-b894-d8f447c75931.png" width="250px"/>

Or we add an overlay on first open of the view that shortly shows that you can swipe to the left and right.
Example:
![image](https://user-images.githubusercontent.com/11789478/178519650-fabf9e2c-7122-4058-bdb2-6a2b0aca064c.png)

Bentipa:
@kordianbruck @CommanderStorm @joschahenningsen maybe you guys have an opinion on this? :)

joschahenningsen:
I didn't know this was possible either 😀
Option one seems quite intuitive!

Bentipa:
@CommanderStorm maybe another opinion before I work on it?

CommanderStorm:
> I didn't know this was possible either :grinning: Option one seems quite intuitive!

I agree with joscha.

PR #1483, opened 2022-07-19
Fixed kino activity and added preview on the sides

Bentipa:
## Issue

Resolves: #1476 and it seems like the recent viewbinding change actually broke the tu film activity.
Fixed that as well.

## Screenshot
image

Being able to see that there are actually movies before and after the entry :)


app/src/main/java/de/tum/in/tumcampusapp/component/ui/tufilm/KinoActivity.kt
```diff
@@ -21,7 +22,7 @@ class KinoActivity : ProgressActivity<Void>(R.layout.activity_kino) {
 
     @Inject
     internal lateinit var viewModelProvider: Provider<KinoViewModel>
-    private lateinit var binding: ActivityKinoBinding;
+    private lateinit var viewPager: ViewPager;
```
CommanderStorm, line 25:
```suggestion
    private lateinit var viewPager: ViewPager
```

app/src/main/java/de/tum/in/tumcampusapp/component/ui/tufilm/KinoActivity.kt
```diff
@@ -56,7 +56,11 @@ class KinoActivity : ProgressActivity<Void>(R.layout.activity_kino) {
             return
         }
 
-        binding.kinoViewPager.adapter = KinoAdapter(supportFragmentManager, kinos)
-        binding.kinoViewPager.currentItem = startPosition
+        // Disable clip to padding
+        viewPager.clipToPadding = false;
+        viewPager.setPadding(60, 0, 60, 0);
+        viewPager.pageMargin = 20;
```
CommanderStorm, line 62:
```suggestion
        viewPager.clipToPadding = false
        viewPager.setPadding(60, 0, 60, 0)
        viewPager.pageMargin = 20
```

PR #1484, opened 2022-07-20
Fix the build

CommanderStorm:
## Issue

This PR fixes the Build.. Again

@kordianbruck or @joschahenningsen can we enable merge restrictions to only allow PRs with green checkmarks to be merged?


Bentipa:
Agree. Let's make that a must. I guess if they do not run reliable they need to be adapted to do so

PR #1494, opened 2022-07-24
Bug/Feature issue template

CommanderStorm:
This PR adds a Bug and a feature template


tobiasjungmann (commented):
From my perspective, this looks good. Can we also include similar templates for creating pull requests? The current one is also a bit minimalistic, or would that be another PR?

joschahenningsen (approved):
Looks good, a PR template would be a good follow-up (especially with some 'how-to-test ') hints :)

CommanderStorm:
> Looks good, a PR template would be a good follow-up (especially with some 'how-to-test ') hints :)

Actually with most of the things I see here, that is pretty obvious how to test them, so I do not think, that this is nessesary..

Issue #1497, opened 2022-07-27
Delete unavailable items in search page


yujinkimmn:
### **Is your feature request related to a problem? Please describe.**
When I click the upper left menu icon, and enter the 'Search page', I can see the result of 'Recent searches'. 
However, even when each item is unknown and unavailable, it still appears on the list.
I think it could be more convenient to delete them, instead of showing them as 'Not available'.     
   

### **Describe the solution you'd like**
Delete the 'Not available' items on Recent searches list.    

   
 
### **Describe alternatives you've considered**
Just get rid of them from the list and provide convenience to user in terms of UI.   
    
   

    

### **Additional context**
Current page
<image src="https://user-images.githubusercontent.com/75295578/181286024-09e57128-7717-4eff-a7a8-f8530bf09bca.png" width="300" height="650">

PR #1498, opened 2022-07-27
Delete unavailable items on Recent Searches list 

yujinkimmn:
## Issue

- Resolves: #1497 

## Screenshot

> Before 
<image src="https://user-images.githubusercontent.com/75295578/181288428-22ed56fe-f650-44d9-af27-e49d92b401cb.png" width="300" height="650"> 

> After 
<image src="https://user-images.githubusercontent.com/75295578/181288456-e9e42053-b2f8-4326-8d00-6b1ea7fd1c5b.png" width="300" height="650"> 

There is no need to show the unavailable items on this page. By doing this, we could increase the usability of this page.


yujinkimmn:
Sorry for messy commit histories :( If I try to drop them, there is a merge conflict.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/search/adapter/RecentSearchesAdapter.kt
```diff
@@ -54,8 +54,7 @@ class RecentSearchesAdapter(
                             .load(person.getFullImageUrl())
                             .into(profilePicture)
                     } catch (exception: Exception) {
-                        titleText.setText(R.string.not_available_search)
-                        profilePicture.setImageDrawable(null)
+                        onRemove(recentSearch)
```
CommanderStorm, line 57:
```suggestion
                        Utils.log(exception)
```
When Picasso throws an error in dev and not just silently hide it, or what is the other way this can fail?

app/src/main/java/de/tum/in/tumcampusapp/component/ui/search/adapter/RecentSearchesAdapter.kt
```diff
@@ -65,7 +64,7 @@ class RecentSearchesAdapter(
                         val room = RoomFinderRoom.fromRecent(recentSearch)
                         titleText.text = room.formattedAddress
                     } catch (exception: Exception) {
-                        titleText.setText(R.string.not_available_search)
+                        onRemove(recentSearch)
```
CommanderStorm, line 67: I cannot reproduce to get an error from `fromRecent`. Is there a way this can fail that I am missing?
Can we just remove the try catch?

app/src/main/java/de/tum/in/tumcampusapp/component/ui/search/adapter/RecentSearchesAdapter.kt
```diff
@@ -76,7 +75,7 @@ class RecentSearchesAdapter(
                         val lecture = Lecture.fromRecent(recentSearch)
                         titleText.text = lecture.title
                     } catch (exception: Exception) {
-                        titleText.setText(R.string.not_available_search)
+                        onRemove(recentSearch)
```
CommanderStorm, line 78: I cannot reproduce to get an error from `fromRecent`. Is there a way this can fail that I am missing?
Can we just remove the try catch?

app/src/main/java/de/tum/in/tumcampusapp/component/ui/search/adapter/RecentSearchesAdapter.kt
CommanderStorm:
```
import de.tum.`in`.tumcampusapp.utils.Utils
```
(is required)

yujinkimmn: When I tried to debug that part, the variable is showed as 'Cannot find local variable 'room''. So it throws error.
Is that what you asked about?

app/src/main/java/de/tum/in/tumcampusapp/component/ui/search/adapter/RecentSearchesAdapter.kt
yujinkimmn: I'm also looking for some way to reproduce it. 
If you think that would be better, we can do like that!

yujinkimmn: Just removing the try catch would be good??

app/src/main/java/de/tum/in/tumcampusapp/component/ui/search/adapter/RecentSearchesAdapter.kt
CommanderStorm: I again tried to reporoduce this: I cannot reproduce this. Do you know any hints on how to reproduce this?

CommanderStorm:
I again tried to reporoduce this: 
I cannot reproduce the cases, for which the catches are needed.
@yujinkimmn Do you know any hints on how to reproduce them?

@PiotrKedra do you have a clue on this issue?

PR #1501, opened 2022-07-28
Cafeteria ratings

### LINKED ISSUE #1490: Feature: Cafeteria Rating
We all like to eat food, especially if it tastes great. Therefore, we should add a rating system for cafeterias which helps by choosing the best cafeteria or meal as well as warning against meals or to recommend particularly good ones.

The backend for this is almost done and will hopefully be merged in the next days (TM). The corresponding [PR](https://github.com/TUM-Dev/Campus-Backend/pull/70) describes all possible endpoints. Alternatively there is also a description in Swagger, which describes the individual parameters and their function in greater detail. 

## What is the overall functionality
1. Create a rating for a cafeteria or a meal in a cafeteria.
	1. Add points (range 1 to 5).
	2. add an optional photo.
	3. add an optional comment (max 256 chars).
	4. add a list of rating tags (rating tags allow to rate a certain aspect quickly with points (1-5) to improve the information of the rating while keeping the time users need to a minimum).
2. Query ratings for cafeterias or a meal in a cafeteria.
	1. Includes average/max/min points as well as the standard deviation.
	2. Includes the last n ratings with everything the user put in.
	3. Includes a list of name tags with average points (name tags are automatically assigned in the backend and group the average points for similar meals like all pasta meals or all meals with pork.)

Remark: There are different rating tags available for cafeteria and meal ratings, only these will be accepted by the backend.

## Suggested Implementation

This is an early draft to just show a potential layout. All the small things like margins etc will be improved in later versions.

Users can choose a cafeteria and afterwards a meal which is available at this day (or no meal if the cafeteria itself should be rated). Additionally, there should be functionality to add a photo, a comment and the tag ratings. If a user accidentally selected points for a tag, these can be reset with the icon on the left (this will only be displayed after the tag was rated.), since the lowest option in the rating bar which can be selected by a user is 1. If a tag has been moved, the font is set darker so that this tag stands out more clearly.

Displaying ratings can happen in multiple form factors: The most intuitive way will be only the average points right next to meal in the meal plan overview. However, users potentially want to go through some of the past ratings, meal tags/ cafeteria tags or name tags to get additional information. Such an overview is ideally directly reachable from the activity where the meals are presented. My idea would be to use the cafeteria name and the ratings at the individual meals as clickable views which navigate to a detailed view. This allows the correct configuration (Cafeteria & Meal) to be loaded directly without the user having to search for it again.

Any new ideas to improve the UI is welcome, especially, if we could manage to reduce the effort for  students, which will ideally lead to more ratings in the long run.

tobiasjungmann:
## Issue

This implements the cafeteria Rating system as it was proposed in this [issue](https://github.com/TUM-Dev/Campus-Android/issues/1490). 
Resolves #1490

## Screenshot
Details will follow once more than the UI is implemented, until now, most details are described in the corresponding issue.


tobiasjungmann:
Will be directly integrated in the flutter app

PR #1506, opened 2022-07-31
Bugfix/grades bar chart

### LINKED ISSUE #1505: [BUG] Grade Bar Chart is scaled wrongly
**Describe the bug**
The Bar Chart in the Grades Fragment has a weird scaling on the x-axis (0 - 41 instead of 1.0 - 5.0).
image

**To Reproduce**
1. Go to "My Grades"
2. Switch to the Bar Chart

**Expected behavior**
The scaling on the x-axis should be 1.0 - 5.0.
image
(I didn't manage to get it completely right, but you probably understand what I feel it should look like)

**Smartphone:**
* Phone: Samsung M51
* OS version: Android 12
* Language: English

**Additional considerations:**

* I don't understand why the bar-values (labels) are formatted as X.0 where X represents the count/amount of such grades. As you can see in the screenshot above, I would suggest switching to just X (or maybe even show the value (like 1.3) instead of the amount).
* I feel like B ratings can be skipped in the graph. I don't see a point in showing them in the bar chart, especially as representing them as something worse than 5.0 is not accurate. One could even think about also removing them from the pie chart, but especially in the bar chart, they don't make much sense.
* "Consider credits and weights in the diagram" ([here](https://github.com/TUM-Dev/Campus-Android/blob/4f75a327aa88a1773ce468aab0743923284b6406/app/src/main/res/layout-land/fragment_grades.xml#L30) and [here](https://github.com/TUM-Dev/Campus-Android/blob/4f75a327aa88a1773ce468aab0743923284b6406/app/src/main/res/layout/fragment_grades.xml#L74)) should not be checked by default. I think activating this feature by default is confusing to most users, especially when they have not weighted their grades (yet). 
image

tobiasjungmann:
This PR resolves #1505 for the grade diagram styling.

Now, the y-axis is removed if weights are used in the diagram.
Additionally, the x-axis now shows the grades instead of "some" numbers and there is a small description for exams whch are marked as passed.

Chart diagram with weights:

Chart diagram without weights:

Since most phone screens are to small to display the values on top of the individual bars, I have replace them with horizontal lines to the y-axis.
Additionally, the button to toggle between the diagrams is not shown in the edit mode anymore.


app/src/main/java/de/tum/in/tumcampusapp/component/tumui/grades/GradesFragment.kt
```diff
@@ -289,41 +290,27 @@ class GradesFragment : FragmentForAccessingTumOnline<ExamList>(
...
-                description = null
-                setTouchEnabled(false)
-
-                axisLeft.granularity = 1f
-                axisRight.granularity = 1f
+                xAxis.granularity = 1f
+                legend.isEnabled = true
 
-                description = null
+                // description = null
```
Bentipa, line 308: remove comment

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/grades/GradesFragment.kt
```diff
@@ -672,13 +675,18 @@ class GradesFragment : FragmentForAccessingTumOnline<ExamList>(
             binding.floatingButtonAddExamGrade.visibility = View.VISIBLE
             binding.chartsContainer.visibility = View.GONE
             binding.checkboxUseDiagrams.visibility = View.VISIBLE
+            barMenuItem?.isVisible = false
+            pieMenuItem?.isVisible = false
             param.setMargins(0, ((32 * density + 0.5f).toInt()), 0, 0)
             binding.gradesListView.setPadding(0, 0, 0, 0)
         } else {
             showExams(exams)
+
```
Bentipa, line 684:
```suggestion
```

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/grades/GradesFragment.kt
```diff
@@ -672,13 +675,18 @@ class GradesFragment : FragmentForAccessingTumOnline<ExamList>(
...
             binding.frameLayoutAverageGrade?.visibility = View.VISIBLE
             binding.floatingButtonAddExamGrade.visibility = View.GONE
             binding.chartsContainer.visibility = View.VISIBLE
+            pieMenuItem?.isVisible = !(binding.barChartView.visibility == View.GONE)
```
Bentipa, line 688:
```suggestion
            pieMenuItem?.isVisible = binding.barChartView.visibility != View.GONE
```

app/src/main/res/values/strings.xml
```diff
@@ -113,7 +113,7 @@
     <string name="add_grade_error_invalid_weight_less_than_zero">Invalid amount of credits. Must be greater than 0.</string>
     <string name="add_exam_dialog_title">Add a New Exam</string>
     <string name="add_exam_dialog_message">Grades are normally added automatically. In case special grades should be added, this can be done here. Manually added Exams can be deleted later on.</string>
-
+    <string name="grade_passed_annotation">Exam Passed Without a Grade</string>
```
Bentipa, line 116:
```suggestion
    <string name="grade_passed_annotation">Exam passed without a grade</string>
```

Bentipa (changes requested):
In addition to the code comments one question.
Why don't you show y-axis makers in the normal view (with weights)?
So previously you can see how many ects per grade you have right? 
Now you can not see it at all anymore?
And what is directly the difference between the diagrams "with" and "without weights"?


PS: I guess you mean bar and not chart diagram.

tobiasjungmann:
Yes, I mean the bar-chart. The bar chart shows not only ects*grade but also includes a weight (e.g. PGDP has 6 ects and a weight of 0.5). Therefore, the y axis wouldn't directly translate to credits, but more as a general relation between them. The version "without weights" only counts exams with this grade, the version with weights computes weight*ects*grade.

Bentipa:
I am wondering what benefit this visualization this would now have for a user.
He knows then that in relation his 1.0 grades might count less than some 2.0 grades right?

For me it's hard to interpret that without knowing what is shown in the diagram at all.
So either we add some explanation why it's useful or we make it more simple (e.g., just showing the amount of grades per grade again). But maybe someone else has another opinion on that? @CommanderStorm

tobiasjungmann:
Yes, it would only show the relation you described, without any real correlation. One idea that came to my mind would be to show percentages on the y axis in this case to visualize the proportion of one grade (with weighting) from the total set of grades and weightings. Then we would have a meaningful(lish) label for the diagram.

CommanderStorm:
> But maybe someone else has another opinion on that? @CommanderStorm

I will review this once my exams are done next week.

CommanderStorm:
This branch crashes on my device. Can you reproduce this?

tobiasjungmann:
The crash was connected to inconsistent database version from switching between this branch and the chat removal. You have to manually uninstall the app. Before relaunching the version from the other branch. Depending on whether the chat removal PR gets merged soon, this can be merged afterwards or directly without the rebasing commits.

CommanderStorm:
@tobiasjungmann sorry for the long delay on this one. Could you rebase onto main? 😅

tobiasjungmann:
Its finally rebased and manageable small again. Looking for your code reviews again!

kordianbruck (approved):

kordianbruck:
Awesome, thanks so much!

PR #1507, opened 2022-08-06
removed "resolveActivity" check on open camera

tobiasjungmann:
## Issue
This PR is a fix for a small bug I stumbled on. Since I found a solution rather quickly, I've skipped the creating an additional issue.

The problem was that the camera view in the feedback activity could not be opened.

## Solution
This code (FeedbackPresenter) finished the method too early:

```
if (takePictureIntent.resolveActivity(context.packageManager) == null) {  
    return  
}
```

This check returns null even if successful in more recent API versions. This check is now removed and the potential exception will be caught if the activity does actually not exist. 

## How to test
Open Feedback tab and add a new image with the camera. Camera view should open.

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/tumui/feedback/FeedbackPresenter.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/tumui/feedback/FeedbackPresenter.kt
@@ -1,6 +1,7 @@
 package de.tum.`in`.tumcampusapp.component.tumui.feedback
 
 import android.Manifest.permission.*
+import android.content.ActivityNotFoundException
 import android.content.Context
 import android.content.Intent
 import android.content.pm.PackageManager
@@ -12,6 +13,7 @@ import android.provider.MediaStore
 import android.util.Patterns
 import androidx.annotation.RequiresApi
 import androidx.core.content.ContextCompat
+import androidx.core.content.ContextCompat.startActivity
 import androidx.core.content.FileProvider
 import de.tum.`in`.tumcampusapp.R
 import de.tum.`in`.tumcampusapp.api.app.TUMCabeClient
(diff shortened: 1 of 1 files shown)


CommanderStorm:
I will review this once my exams are done next week.

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/feedback/FeedbackPresenter.kt
```diff
@@ -127,8 +125,11 @@ class FeedbackPresenter @Inject constructor(
         val authority = "de.tum.in.tumcampusapp.fileprovider"
         val photoURI = FileProvider.getUriForFile(context, authority, photoFile)
         takePictureIntent.putExtra(MediaStore.EXTRA_OUTPUT, photoURI)
-
-        view?.openCamera(takePictureIntent)
+        try {
+            view?.openCamera(takePictureIntent)
+        } catch (e: ActivityNotFoundException) {
+            photoFile.delete()
```
Bentipa, line 131: why is this required? is the file already created when getting the Uri?

tobiasjungmann: The file is created before trying to open the new activity (line 115). If the activity cannot be opened, the file won't be used. Therefore, it should be deleted again. Previously, the file would not even be created if the check which I have removed failed.

CommanderStorm (approved):

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/feedback/FeedbackPresenter.kt
Bentipa: Ah sorry, did not open the collapsed section :D

Issue #1518, opened 2022-10-21
[Feature] An option to share a link of the room


nilsreichardt:
**Is your feature request related to a problem? Please describe.**
Someone asked where exactly a room for our lecture. I opened the campus app, searched for the room and tried to find a share button to share a link to the room.

**Describe the solution you'd like**
I would like to have a share button of the showed page to share the link to the room. In the case of MW 2001 this could be this link: https://nav.tum.sexy/room/5510.02.001 (or another roomfinder link).

**Describe alternatives you've considered**
An alternative is to just send a screenshot of the app.

**Additional context**
![screenshot-of-room-finder-mw-2001](https://user-images.githubusercontent.com/24459435/197276341-9705b152-b444-4acc-bafa-06208e8a44e3.jpg)

PR #1528, opened 2022-11-14
Implement location sharing for rooms

ctsk:
## Issue

- Resolves: #1518

## Screenshot

image

It allows communicating locations on the campus even when the peer does not have the app themselves.

## Draft

This is currently a draft, as it merely allows to share rooms. When the viewed item is not a room, the button is not visible. I'd like to add the ability to share "buildings" (and other location types that "nav.tum.sexy" supports before merging.


ctsk:
So this pretty much works now, however the button placement is not satisfactory yet.

ctsk:
Update: Moved the button into the toolbar:

image

TODO: Find a proper android share icon

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/roomfinder/NavigationDetailsFragment.kt
```diff
@@ -85,13 +85,29 @@ class NavigationDetailsFragment : BaseFragment<Unit>(
...
+        setupShareLocationButton(navigationDetails)
         setOpenInOtherAppBtnListener(navigationDetails)
 
         showNavigationDetailsProperties(navigationDetails)
 
         showAvailableMaps(navigationDetails)
     }
 
+    private fun setupShareLocationButton(navigationDetails: NavigationDetails) {
+        val url = "https://nav.tum.sexy/${navigationDetails.type}/${navigationDetails.id}";
```
CommanderStorm, line 97:
```suggestion
        val url = "https://nav.tum.sexy/${navigationDetails.type}/${navigationDetails.id}"
```

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/roomfinder/NavigationDetailsFragment.kt
```diff
@@ -85,13 +85,29 @@ class NavigationDetailsFragment : BaseFragment<Unit>(
...
+
+        binding.toolbarNav.shareButton.setOnClickListener {
+            val sendIntent = Intent().apply {
+                action = Intent.ACTION_SEND
+                putExtra(Intent.EXTRA_TEXT, url)
+                type = "text/plain"
+            }
+
+            val shareIntent = Intent.createChooser(sendIntent, null);
+            startActivity(shareIntent);
```
CommanderStorm, line 107:
```suggestion
            val shareIntent = Intent.createChooser(sendIntent, null)
            startActivity(shareIntent)
```

CommanderStorm (commented):
Formatting fixes

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/roomfinder/NavigationDetailsFragment.kt
```diff
@@ -89,13 +89,29 @@ class NavigationDetailsFragment : BaseFragment<Unit>(
...
+        setupShareLocationButton(navigationDetails)
         setOpenInOtherAppBtnListener(navigationDetails)
 
         showNavigationDetailsProperties(navigationDetails)
 
         showAvailableMaps(navigationDetails)
     }
 
+    private fun setupShareLocationButton(navigationDetails: NavigationDetails) {
+        val url = "https://nav.tum.sexy/${navigationDetails.type}/${navigationDetails.id}"
```
CommanderStorm, line 101: This is sadly not 100% correct. Either we do this:
```suggestion
        val url = "https://nav.tum.sexy/view/${navigationDetails.id}"
```

Or steal this code:
https://github.com/TUM-Dev/NavigaTUM/blob/main/webclient/src/views/view/view-view.js#L24-L35

app/src/main/res/values-de/strings.xml
```diff
@@ -781,6 +781,7 @@ Signatur: %5$s</string>
     <string name="loading_details">lädt</string>
     <string name="site_plans">Lagepläne</string>
     <string name="interactive_map">Interaktive Karte</string>
+    <string name="share_location">Ort teilen</string>
```
CommanderStorm, line 784: This translation is never used. 
Is this a mistake?

CommanderStorm (changes requested):
I have two changes,
Also: you did remove the share functionality in the last commit. Was that intentional

app/src/main/res/values-de/strings.xml
ctsk: Yes, a mistake. It's a leftover from the first iteration where there was a button with text. (see screenshot in original pull request)

ctsk:
> Also: you did remove the share functionality in the last commit. Was that intentional?

Did I? It's there on my end (when checking out the PR). I removed the large button and instead added an icon to the toolbar.

CommanderStorm:
Ah, now I noticed it…

Honestly, not noticing this is probably something that most people will do.
I agree that two stacked buttons take up a lot of space.
Have you experimented with two horizontally stacked buttons with less text (f. ex. `share` and `open in`, maybe with an icon)?

ctsk:
I've experimented with a bunch of layouts, and I think I like this one the most:

<details>
<summary> Screenshot </summary>
image
</details>

ctsk:
Not sure how to say "Open in Maps" in German without it getting too long. Perhaps "Open with ..."?

CommanderStorm:
> Unrelated: Is there a particular reason to use nav.tum.sexy and not nav.tum.de directly?

We got nav.tum.de like 2-3 Weeks ago, this is just an alias for now (nav.tum.sexy returns a permanent redirect, which the app follows)
Exchanging this string should probably be another PR 😉


> Not sure how to say "Open in Maps" in German without it getting too long. Perhaps "Öffnen mit ..."?

I would probably say `Maps öffnen` is the better option, but I am not a german expert (I barely got passing grades in the german Abitur...)

CommanderStorm:
> I've experimented with a bunch of layouts, and I think I like this one the most

I think this is better, but I would like to experiment with a few other ideas (in terms of fewer margins, or maybe [something like this](https://medium.com/nerd-for-tech/how-to-add-extended-floating-action-button-in-android-android-studio-java-481cc9b3cdcb)). Could you push this version?

ctsk:
<details>

<summary> Screenshot </summary>

image

</details>

Any ideas on how to avoid clashing with the site plans selection?

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/roomfinder/NavigationDetailsFragment.kt
```diff
@@ -89,13 +89,29 @@ class NavigationDetailsFragment : BaseFragment<Unit>(
...
+        setupShareLocationButton(navigationDetails)
         setOpenInOtherAppBtnListener(navigationDetails)
 
         showNavigationDetailsProperties(navigationDetails)
 
         showAvailableMaps(navigationDetails)
     }
 
+    private fun setupShareLocationButton(navigationDetails: NavigationDetails) {
+        val url = "https://nav.tum.sexy/view/${navigationDetails.id}"
```
kordianbruck, line 101: nav.tum.app pls

app/src/main/res/layout/toolbar_navigation.xml
```diff
@@ -1,16 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
```
kordianbruck, line 1: Is this an unrelated cleanup? I'd appreciate to put this into a separate PR, so its easier to roll back in case this actually still used somewhere.

app/src/main/java/de/tum/in/tumcampusapp/component/tumui/roomfinder/NavigationDetailsFragment.kt
CommanderStorm: Since we serve at `nav.tum.de`,
I think this would be better:
```suggestion
        val url = "https://nav.tum.de/view/${navigationDetails.id}"
```

CommanderStorm (approved):

CommanderStorm:
I Agree.
Works as intended => merging

Thanks for this contribution

Issue #1522, opened 2022-10-24
[BUG] Caching issues in the Roomfinder


octycs:
**Describe the bug**
The Roomfinder functionality caches the map from previous searches. If I search for a new room that would be on this map, without closing the app, I will get the old map with the old coordinates

**To Reproduce**
1. Open the menu and click on search
2. Search for "MW0001" and click on the result
3. Go back to search
4. Search for "MW1801" and click on the result

**Expected behavior**
The map displayed should show the position of MW1801 and not the cached map for MW0001

**Screenshots**
<img src="https://user-images.githubusercontent.com/7429408/197502607-7b611bb1-0d61-4d28-8e02-1bdc3ad9b895.png" width=" 400px">
(Should be on the left part of the building)

**Smartphone (please complete the following information):**
* Phone: Samsung
* OS version: Android 11 (Lineage OS 18.1)
* Language: German

**Additional context**
Restarting the app removes the cache

ctsk:
Yup, can reproduce!

PR #1530, opened 2022-11-14
Fix map image pin caching

ctsk:
## Issue

- Resolves: #1522

## Screenshot
image
image


CommanderStorm (approved):
Thank you for this contribution ❤️ 
Works as expected

Issue #1531, opened 2022-11-14
[Devel Bug] ClassCastException when running in emulator without Play Store


ctsk:
When the app is running in the emulator where the play store isn't available, the app crashes when requesting a review. The stack trace reveals these lines as the culprit:

https://github.com/TUM-Dev/Campus-Android/blob/9d075dd4d587e3f205ac8fb72a96a7523397f621/app/src/main/java/de/tum/in/tumcampusapp/component/ui/overview/MainFragment.kt#L188-L191

Stack trace:
```
Process: de.tum.in.tumcampus, PID: 19000
    java.lang.ClassCastException: com.google.android.play.core.internal.zzat cannot be cast to com.google.android.play.core.review.ReviewException
        at de.tum.in.tumcampusapp.component.ui.overview.MainFragment.triggerReviewPrompt$lambda-2(MainFragment.kt:190)
        at de.tum.in.tumcampusapp.component.ui.overview.MainFragment.$r8$lambda$m8fOp0j6HyX2ulP4M4cVAfmKVBg(Unknown Source:0)
        at de.tum.in.tumcampusapp.component.ui.overview.MainFragment$$ExternalSyntheticLambda0.onComplete(Unknown Source:4)
        at com.google.android.play.core.tasks.zza.run(com.google.android.play:core@@1.10.3:1)
        at android.os.Handler.handleCallback(Handler.java:942)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loopOnce(Looper.java:201)
        at android.os.Looper.loop(Looper.java:288)
        at android.app.ActivityThread.main(ActivityThread.java:7898)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:936)
```

Since the result of that cast isn't used, I suggest removing it. Objections?

PR #1532, opened 2022-11-14
Remove cast to ReviewException as result is unused

ctsk:
## Issue

- Resolves: #1531

## Why this is ~useful for all students~ an improvement

Could be inconsequential, could be that the bug is also encountered "in the wild". At the least it helps developers :)

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/ui/overview/MainFragment.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/ui/overview/MainFragment.kt
@@ -185,9 +185,6 @@ class MainFragment : BaseFragment<Unit>(
                 flow.addOnCompleteListener {
                     Utils.setSetting(requireContext(), Const.LAST_REVIEW_PROMPT, Date().time.toString())
                 }
-            } else {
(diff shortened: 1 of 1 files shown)


CommanderStorm:
@Liqs-v2 what was the intention with this?
Shouldn't this info be used instead show a modal?

XSpielinbox:
As this bug also effects startup of the app for normal usage, e.g. on  Custom ROMs, I would be grateful, if this could be removed until there is new code that also works well on all devices.

CommanderStorm (changes requested):
What do you think about this change?

MainFragment.kt
````
} else {
    // There was some problem, log or handle the error code.
    Utils.showToast(requireContext(),R.string.in_app_review_failed_to_start)
    task.exception?.let { Utils.log(it) }
}
```

strings.xml
en: `<string name="in_app_review_failed_to_start">In App ReviewFlow failed to start</string>`
de: `<string name="in_app_review_failed_to_start">In App ReviewFlow konnte nicht gestartet werden</string>`

CommanderStorm:
closed in favor of #1564

Issue #1535, opened 2022-11-16
[BUG] Very long search query breaks entry field


ctsk:
**Describe the bug**
When entering a very long search, the entry field expands behind the  "x" button.

**To Reproduce**
1. Go to 'search'
2. Enter a very long query

**Expected behavior**
The entry field shouldn't expand beyond the "x" button

**Screenshots**

![image](https://user-images.githubusercontent.com/9384305/202151383-0b3f87bc-31e2-45eb-a2e2-b889f0d1529a.png)

**Smartphone (please complete the following information):**
* Phone: Pixel
* OS version: 12
* Language: english

PR #1537, opened 2022-11-16
fixed overlapping ui

tobiasjungmann:
## Issue

- Resolves: [#1535](https://github.com/TUM-Dev/Campus-Android/issues/1535)

Buttons do not overlap anymore, text will stop before the button is reached.

### DIFF
--- a/app/src/main/res/layout/toolbar_search.xml
+++ b/app/src/main/res/layout/toolbar_search.xml
@@ -12,13 +12,13 @@
         android:theme="@style/TumToolbar"
         app:popupTheme="@style/TumToolbarPopUp">
 
-        <FrameLayout
+        <androidx.constraintlayout.widget.ConstraintLayout
             android:layout_width="match_parent"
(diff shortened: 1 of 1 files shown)


ctsk:
Fixes the issue for me. The whitespace is weird though imo.

<details>
<summary> Screenshot </summary>

image

</details>

tobiasjungmann:
Oh, I forgot the screenshot myself. Which whitespace do you mean? The one left or right to the "x"?

ctsk:
I guess it's more of the mismatch between left and right side of "x", and the difference to the gap between the spacing between the input and "<-" ...

tobiasjungmann:
The spacing between the text and the "back error" is the same as the activity titles in the other activities. Changing this would look more inconsistent in my opinion. Using the same space on the right would reduce the size of the search query unnecessarily.

joschahenningsen (approved):
Looks good :+1:

Issue #1454, opened 2022-06-02
Adding monthly calendar 


yujinkimmn:
## Problem
On Calendar menu, only the weekly calendar is available.
If we have monthly calendar as well, it would be more convenient. 

## Expected Result :sparkles:
Add monthly calendar option, so if user clicks button, monthly calendar appears.

Marius1501:
Could you assign me to this issue, please? I would like to work on it. 😃

PR #1539, opened 2022-11-25
Provisional month view

Marius1501:
## Issue

- Resolves: #1454 

## Screenshot
Month view:
grafik

Month icon:
grafik

Adds month view to the calendar

## Not complete yet
Is this implementation good? If yes I would work on the month view further and would try to implement also events to it.


Marius1501:
I implemented a month view from scratch. It would be nice to get some feedback. I know that it is not perfect, but maybe the appraoch is good. I have some things that should be implemented or that should be improved. 
- Sliding to the previous or the next month with an animation
- Adding a touchListener to the events
- Improving the appearance of the events
Screenshot_20230224_225526

CommanderStorm:
I have changed the design in this way:

image
image

What do you think?

CommanderStorm:
I see these functionality improvements (these should probably be separate PRs):
- how eazy would it to include the days of previous months in those spaces where they belong (see Google calendar)
- how eazy would it to link to days if they are clicked on
- how eazy would it to change the current button navigation to scroll navigation

Do you agree?

Marius1501:
> I have changed the design in this way:
> 
> image image
> 
> What do you think?

Yes, your approach looks better. 👍🏻

Marius1501:
> I see these functionality improvements (these should probably be separate PRs):
> 
>     * how eazy would it to include the days of previous months in those spaces where they belong (see Google calendar)
>       image
> 
>     * how eazy would it to link to days if they are clicked on
> 
>     * how eazy would it to change the current button navigation to scroll navigation
> 
> 
> Do you agree?

@CommanderStorm Yes, I totally agree. 
- I think the inclusion of the days of the previous and next month is a bit difficult but feasible. 
- The handling of a click on a day should be performed by a pop-up view of the day (see Google calendar)
- I think it will be easy to add scroll navigation but an animation should also be added. And should we then get rid of the next and previous months buttons?
- Should I open issues for these topics when this PR is merged?

CommanderStorm:
Closing in favor of #1558 (I cant add commits here)

Issue #1536, opened 2022-11-16
[BUG] Transportation json shows up in search results


ctsk:
**Describe the bug**
After looking at the transportation card, a json blob appears in the search results.

**To Reproduce**
1. Click on transportation card
2. Exit transportation card
3. Open drawer, go to search
4. See some json in the recent searches, Clicking on it doesn't do anything

**Expected behavior**
That json blob shouldn't show in the search results.

**Screenshots**
![image](https://user-images.githubusercontent.com/9384305/202152231-1bc0ef96-4824-4f7c-a94d-1bd8b14b935b.png)

**Smartphone (please complete the following information):**
* Phone: pixel
* OS version: 12
* Language: english

PR #1541, opened 2022-12-01
Fixed the recent search bug

Marius1501:
## Issue

- Resolves: #1536 

Now the undesirable station-search is filtered from the recent searches.


app/src/main/java/de/tum/in/tumcampusapp/component/ui/search/SearchViewModel.kt
```diff
@@ -175,7 +175,7 @@ class SearchViewModel @Inject constructor(
         val recentSearchesDao: RecentsDao = TcaDb.getInstance(context).recentsDao()
         val recentSearches: List<Recent> = recentSearchesDao.allRecentSearches.reversed() ?: emptyList()
         state.value = state.value.copy(
-            recentSearches = recentSearches
+            recentSearches = recentSearches.filter { x -> x.type != 1 } // filter for undesirable stations
```
tobiasjungmann, line 178:
```suggestion
            recentSearches = recentSearches.filter { x -> x.type != STATIONS } // filter for undesirable stations
```
Please change this to the constant to make the code easier to understand. This is exactly the type you want to use. (And import `import de.tum.in.tumcampusapp.component.other.general.RecentsDao.STATIONS`)

tobiasjungmann (changes requested):

tobiasjungmann:
Otherwise an elegant Fix for these unwanted results. The code works for me

tobiasjungmann:
Please import the constant with the import from the review, otherwise the build will fail. I wasn't able to add this in the suggestion directly.

Issue #1542, opened 2022-12-01
[BUG] Back button is missing in the person-details view


Marius1501:
**Describe the bug**
There is no back-button in the person-details view

**To Reproduce**
1. Go to the drawer and open search
2. Search for any person
3. Click on the person
4. See that there is no back-button

**Expected behavior**
There should be a working back-button

**Screenshots**
![grafik](https://user-images.githubusercontent.com/109673982/205035134-ca78e56e-64e2-464c-9fca-2d3cacbb3096.png)

**Smartphone:**
* Phone: Pixel 6
* OS version: 33

PR #1543, opened 2022-12-01
Fixed the person-details toolbar

Marius1501:
## Issue

- Resolves: #1542 

Adds toolbar and back button to the person-details view


tobiasjungmann (approved):
Thank you for fixing making the app design more much more consistent! Code looks good and works.

PR #1544, opened 2022-12-01
Fixed the mensa-notification toolbar

Marius1501:
## Issue

This fixes the following issue(s):
- Resolves:: #1542 

## Screenshot
grafik

## Why this is useful for all students
It solves the same problem as #1543 but for the cafeteria-notification view

PR #1545, opened 2022-12-01
Fixed the feedback toolbar

Marius1501:
## Issue

- Resolves: #1542 

## Screenshot
grafik

It solves the same problem as #1543  but for the feedback view


tobiasjungmann (commented):
Ah, there are more similar PRs :D Please merge this with https://github.com/TUM-Dev/Campus-Android/pull/1544 for acombined PR and rebase it onto master, then we can directly merge it.

PR #1546, opened 2022-12-01
Fixed the information toolbar

Marius1501:
## Issue

- Resolves: #1542 

## Screenshot
grafik

It solves the same problem as #1543 but for the information view


tobiasjungmann (approved):
Thanks again for adding the missing back button! If you find more occurrences, you can add them in the same PR, they are close enough thematically :)

Issue #1547, opened 2022-12-09
[BUG] Icon changes when refreshing the edit grade view


Marius1501:
**Describe the bug**
When refreshing the edit grades site the toolbar changes and the icons from the grades list appear.

**To Reproduce**
1. Go to edit grades site
2. Refresh the site by swiping down
3. See that the toolbar changes

**Expected behavior**
The toolbar should stay the same after refreshing.

**Smartphone:**
* Phone: Pixel 6
* OS version: 33
* Language: English

PR #1549, opened 2022-12-12
Removed Faulty "invalidateOptionsMenu()"

tobiasjungmann:
## Issue
https://github.com/TUM-Dev/Campus-Android/issues/1547
- Icon is now not set to the edit icon after the grades have been fetched from the server
- Removed a dangling `invalidateOptionsMenu()` call which had no use.
- Additionally cleared some uses of with(binding) in the grades fragment.

## Screenshot
Will be added shortly


CommanderStorm (approved):
Thanks for this work, no complaints 🥳

Issue #1540, opened 2022-11-29
[Feature] Show current lecture in progress


VictorHartman:
The next lecture block is useful to know where my next class is. However when I am late, even just one minute, and I look on my phone this info is no longer available.

The solution is to show the current class in progress. This way I will know where I am supposed to be when I am delayed.

xarantolus:
I'm interested in implementing this change.
I experimented by displaying all lectures that have not yet ended (which will then show the currently in progress lecture), but I'm not sure if that's the best solution (because then at the end of a lecture, it will still show the current lecture).
Maybe it makes sense to display the current lecture if the current time is before or up to 15/30 minutes after the start time?

CommanderStorm:
We could also display multiple lectures and have 15 m overlap where two lectures take place (assuming overlap).
This does also solve the problem that sometimes lectures overlap (one online+recorded, one not).

VictorHartman:
What I had in mind was to display two cards. One for currently in session and one for coming up.

xarantolus:
I have [tried out](https://github.com/TUM-Dev/Campus-Android/compare/master...xarantolus:TUM-Campus-Android:feature/lecture-card-time) splitting the lectures into in progress/upcoming (or by next lectures that start within 15 minutes of each other as fallback) and then displaying multiple in the card without expanding.
I also tried more complex criteria, showing multiple overlapping lectures etc, but I feel like often there's a case where one could argue that something is behind the "expand" button when it should not be.
I feel like it would make sense to just display up to 4 lectures of the current day (the same day as the first lecture) that haven't ended. That way, the card gives you a quick overview of the rest of the day, including currently ongoing and overlapping lectures. The expand button would then only show lecture that are not today.
So the question is whether we want more complex logic that tries to minimize the number of lectures shown, or we just take a bit more screen space and show e.g. up to 4 of the next items (if they are today, rest behind expand button).

CommanderStorm:
Sounds reasonable to me

PR #1550, opened 2022-12-20
Next lecture card: show currently ongoing lectures

xarantolus:
This PR changes the "Next lecture" card of the overview screen to show up to 4 lectures of the current day (the day of the next calendar event) that have not yet ended instead of just one.

## Issue

- Resolves: #1540

## Screenshot
| Normal | Expanded |
| ---- | ---- |
| normal | expanded |

This allows you to get a quick overview of the rest of the day, while still showing currently ongoing lectures (and thus still showing you where to go if you are late) as well as overlapping lectures

----

This PR is currently a draft to discuss the following:
* In settings this is called "Nächste Vorlesung", but now it shows multiple. Should I rename that as well (including the class names)?
* Should we change the time formatting code to work for past lectures, e.g. in case of the screenshot it could make sense to display 08:30-10:00 instead of just the start time?
    * I think it would also make sense to display e.g. "Morgen 11:00" instead of just "Morgen" in the expanded item of the screenshot

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarDao.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/tumui/calendar/CalendarDao.kt
@@ -31,7 +31,7 @@ interface CalendarDao {
 
     @get:Query("SELECT * FROM calendar " +
             "WHERE status!='CANCEL' " +
-            "AND dtstart > datetime('now', 'localtime') " +
+            "AND dtend > datetime('now', 'localtime') " +
             "GROUP BY title, dtstart, dtend " +
             "ORDER BY dtstart LIMIT 4")
     val nextUniqueCalendarItems: List<CalendarItem>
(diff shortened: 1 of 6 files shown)


VictorHartman:
This looks great! I like how you can expand the card to view the lectures of the next day. I have some questions/comments:

- Does the expansion of the card only hide the lectures of the next day? So, for example, if I have a lot of lectures on one day will the card be really long? Not necessarily an issue, but you could think about limiting the amount of lectures that are shown in short card view.
- Is the 'current lecture' a placeholder? It should show the name of the lecture, right? The end time and location should be included as well. Suggestions to indicate the lecture is in progress: (1) just add 'in progress' before the name. May not look that great. (2) give the time a different colour. 
- May also be an idea to show time remaining for current lecture.

xarantolus:
> Does the expansion of the card only hide the lectures of the next day?

The card gets up to 4 lectures, and these four are then split into "day of first lecture" and everything else (which is hidden behind the expander). So I think the length shouldn't be a problem.

> Is the 'current lecture' a placeholder? 

It's the title of the event in my TUMonline calendar, so it's "real" data that just doesn't have any info attached to it :)

> give the time a different colour

Sounds really nice, I'll try that!

> May also be an idea to show time remaining for current lecture.

Yes, I definitely want to improve the time displayed there

kordianbruck (approved):

kordianbruck:
Thanks :beers:

PR #1551, opened 2022-12-21
fix mensa widget by adding FLAG_IMMUTABLE to Intent

jfoerste:
## Issue

The mensa widget was blank since it's Intent was forgotten in #1429.

## Solution
Fixed by adding `FLAG_IMMUTABLE` to Intent.


tobiasjungmann (approved):
Great, this fixes it for Android 12

PR #1556, opened 2023-02-25
Make search result type buttons nicer

jfoerste:
## Issue
- The buttons which filter search results by type flicker upon making a selection
- The animation list animation looks janky when changing a lot of items

## Solution
- Replace custom made buttons with material design chips
- Stop replacing button views when only the selection state changes
- Remove default ItemAnimator from the main RecyclerView

There are some slight visual changes, the Chips could be assimilated further to the custom button design if necessary.

### Before
https://user-images.githubusercontent.com/40441040/221376433-e7874f20-afdc-4767-ae5c-f5aab4b0a8d5.mp4
### After
https://user-images.githubusercontent.com/40441040/221376431-969a68bf-8136-4628-ae87-6d1496ac0953.mp4


app/src/main/java/de/tum/in/tumcampusapp/component/ui/search/adapter/ResultTypesAdapter.kt
```diff
@@ -42,49 +39,65 @@ class ResultTypesAdapter(
...
         holder.bind(resultType)
     }
+
+    override fun onBindViewHolder(holder: ResultTypeViewHolder, position: Int, payloads: MutableList<Any>) {
+        if (payloads.isEmpty()) {
+            super.onBindViewHolder(holder, position, payloads)
+        } else {
+            holder.setIsSelected(payloads[0] as Boolean)
+        }
+    }
```
CommanderStorm, line 88: I dont get this part: Could you explain this (What is this doing?)

app/src/main/java/de/tum/in/tumcampusapp/component/ui/search/adapter/ResultTypesAdapter.kt
```diff
@@ -42,49 +39,65 @@ class ResultTypesAdapter(
...
+            chip.text = chip.context.resources.getText(textId)
+            setIsSelected(typeData.selected)
+
+            if (typeData.type == SearchResultType.ALL) {
+                itemView.margin(left = 16F)
             } else {
-                typeTextView.textColorResource = R.color.text_primary
-                layout.setBackgroundResource(R.drawable.search_result_type_background)
+                itemView.margin(left = 8F)
             }
```
CommanderStorm, line 57: I dont fully know about this part (this is fine too)

what do you think about this alternative?
````kotlin
val leftMargin = when (typeData.type == SearchResultType.ALL) {
                true -> 16F
                false -> 8F
            };
            itemView.margin(left = leftMargin)

CommanderStorm (commented):
Gennerally, I really like these changes. This could be merged imo

There is one associated issue, which this PR touches on, but does not fix (this can be its own PR though):
Recording 2023-03-15 at 17 07 47

What do you think about this flickering?

app/src/main/java/de/tum/in/tumcampusapp/component/ui/search/adapter/ResultTypesAdapter.kt
jfoerste: Because I have overridden `getChangePayload()`, whenever new data is submitted and the only difference between old and new buttons of the same name is the "selected"-state, the new state is added as a payload. In these cases it is not necessary to replace the existing view using `super.onBindViewHolder()` as the existing one can be updated using `setIsSelected()` instead.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/search/adapter/ResultTypesAdapter.kt
jfoerste: I have added an even prettier variant using the xml-attribute `clipToPadding`, removing this logic entirely.

app/src/main/java/de/tum/in/tumcampusapp/component/ui/search/adapter/ResultTypesAdapter.kt
CommanderStorm: I think adding a comment explaining this circumstance could be worthwhile

app/src/main/java/de/tum/in/tumcampusapp/component/ui/search/adapter/ResultTypesAdapter.kt
CommanderStorm: Looks way better. Agree

app/src/main/java/de/tum/in/tumcampusapp/component/ui/search/adapter/ResultTypesAdapter.kt
jfoerste: I've added two comments which should explain the idea.

CommanderStorm (approved):

PR #1563, opened 2023-05-23
Add dependabot to check for dependency updates

ropapermaker:
## Issue

- General security issues.
- Add Dependabot to check for security and version updates for dependencies..

Dependabot automatically creates pull requests for new package versions and security updates which can then be manually merged. The `.github/dependabot.yml` file can be edited to allow updates only on specific packages or ignore other ones.

For more info: [configuring-dependabot-version-updates
](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates)

I tested the dependabot funcionality on my fork and below in the screenshots are how it would look like if it were to be activated.
### Dependabot Security Alerts and version updates can be added from the settings tab:

<hr>

### Security alerts:

<hr>

A more secure app

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,16 @@
+# Basic set up for gradle and bundler
+
+version: 2
+updates:
+
+  # Maintain dependencies for Bundler
+  - package-ecosystem: "bundler"
+    directory: "/"
+    schedule:
+      interval: "daily"
+
+  # Maintain dependencies for gradle
+  - package-ecosystem: "gradle"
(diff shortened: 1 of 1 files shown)


CommanderStorm:
Good idea, but I think we should choose Renovate, as we have made better experiences with it in other Repos (namely https://github.com/TUM-Dev/navigatum and https://github.com/joschahenningsen/TUM-Live)

ropapermaker:
That also seems like a good idea. Thank you for the feedback!

PR #1564, opened 2023-05-24
Fix/remove review exception cast

CommanderStorm:
Successor to #1532 as I can't update said branch

## Issue

- Resolves: #1531

## Why this is ~useful for all students~ an improvement

Could be inconsequential, could be that the bug is also encountered "in the wild". At the least it helps developers :)


CommanderStorm:
@joschahenningsen could you approve this, so we can merge this?

joschahenningsen (approved):

Issue #1589, opened 2023-06-22
[BUG] App crashes when opening item from NavigationDrawer after opening anything from About


kunzef:
**Describe the bug**
The app crashes when trying to select an activity from the NavigationDrawer after opening any of the Activites listed under _About_. It only seems to be happening when I have selected English as language. I have attached the stacktrace below.

~~~
FATAL EXCEPTION: main
Process: de.tum.in.tumcampus, PID: 8692
java.lang.IllegalArgumentException: Invalid menu item Opening Hours provided 
    at de.tum.in.tumcampusapp.component.other.generic.drawer.DrawerMenuHelper.findNavItem(DrawerMenuHelper.kt:106) 
    at de.tum.in.tumcampusapp.component.other.generic.activity.BaseNavigationActivity$initDrawer$1$1.invoke(BaseNavigationActivity.kt:93) 
    at de.tum.in.tumcampusapp.component.other.generic.activity.BaseNavigationActivity$initDrawer$1$1.invoke(BaseNavigationActivity.kt:92) 
    at de.tum.in.tumcampusapp.utils.ExtensionsKt$closeDrawers$1.onDrawerClosed(Extensions.kt:103) 
    at androidx.drawerlayout.widget.DrawerLayout.dispatchOnDrawerClosed(DrawerLayout.java:891) 
    at androidx.drawerlayout.widget.DrawerLayout.updateDrawerState(DrawerLayout.java:861) 
    at androidx.drawerlayout.widget.DrawerLayout$ViewDragCallback.onViewDragStateChanged(DrawerLayout.java:2249) 
    at androidx.customview.widget.ViewDragHelper.setDragState(ViewDragHelper.java:920) 
    at androidx.customview.widget.ViewDragHelper$2.run(ViewDragHelper.java:345) 
    at android.os.Handler.handleCallback(Handler.java:942) at android.os.Handler.dispatchMessage(Handler.java:99) 
    at android.os.Looper.loopOnce(Looper.java:201) at android.os.Looper.loop(Looper.java:288) 
    at android.app.ActivityThread.main(ActivityThread.java:7884) at java.lang.reflect.Method.invoke(Native Method) 
    at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548) 
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:936)
~~~

**To Reproduce**
1. Open the App
2. Set to english
[...]

Issue #1314, opened 2019-06-15
Streamline new navigation logic


thellmund:
Can we do this better than linearely searching through all the items?
Maybe do a Map or something?

_Originally posted by @pfent in https://github.com/TUM-Dev/Campus-Android/pull/1313_

CommanderStorm:
Is prformance not good enough? On my devie search works fine as it is...
Shure enough fancy, but neccesary?

Issue #1593, opened 2023-07-06
[BUG] App fails to set Locale correctly


kunzef:
**Describe the bug**
The application fails to set its own Local correctly. This becomes visible when the System Locale and the Application Locale differ. This can lead to the UI being in German while the App language is set to English. This bug most likely caused #1589, #1585 and maybe #1517.  

**To Reproduce**
1. Set System Locale/language to english and Application Locale/language to german or vice versa
2. Open any of the Activities under the About section of the NavigationDrawer.
3. Press the back button
4. Open any of the Fragments under except the Activites under the About section
5. All Fragments except the Activites under the About section are now in the System language while the NavigationDrawer and the Activies under the About are still in the Application language.

**Expected behavior**
Everyhting should be in the Application language.

**Smartphone (please complete the following information):**
* Phone: Pixel 6
* OS version: 13
* Language: System German, Application English (same the other way around)

**Suspected cause**
The Application Locale is currently set in the _BaseActivity_ under _initLanguage_. This works fine as long as only Fragments are openend and no Activity. The problem with the approach of setting the Locale as it is currently done is that is is only set for one Context. I suspect that when calling an Activity the Context changes thus reseting the Locale. This could in theory be counterd by setting the Locale again for the new Context. But in practice this was not succesfull. I believe the reason for this is that the Fragments are partly created by using injection and the injection is controlled by a Factory which Context seems to be generated outside of the normal ApplicationContext thus using the System Locale. For me it was not possible to access this Context and set the Locale manually. I have described what I believe to be the best and most straightforward fix below.

**Fix for the issue**
In SDK 33 Android introduced an _LocaleManager_ or the option to set _android:localeConfig_ see [here](https://developer.android.com/training/basics/supporting-devices/languages#app-languages). I believe this could fix the problem as the Locale is set Application wide and not only for one context. Unfortunalty the Application is currently using SDK 32 as its build SDK as well as SDK 26 as min SDK. While I do think that the min SDK of 26 should not be a problem we would definitely have to bump the build SDK.

Issue #1585, opened 2023-06-11
[BUG] Mixed languages


Garfield96:
**Describe the bug**
While the app language is set to English, some pages show the title in German (e.g. "Lecture Information" or "About the app") (see screenshot). Other elements on these pages use the correct translation, e.g. the "Lecture Appointments" button.

**To Reproduce**
1. Set language to English.
2. Go to "Lecture Information" or "About the app"
3. See incorrect title

**Expected behavior**
All elements should use the same language (except for elements fetched from external sources, like TUMonline).

**Screenshots**
<img src="https://github.com/TUM-Dev/Campus-Android/assets/15248220/c66da90f-da41-4177-8339-631f7add7484" width="300">

**Smartphone :**
* Phone: Redmi Note 8 Pro
* OS version: 11 RP1A.200720.011
* Language: German

**Additional context**
Could be related to switching the language (maybe some caching issue).

SecretIdetity:
possibly also the title in the about section

![Screenshot_2023-06-12-14-03-52-248_de.tum.in.tumcampus~2.jpg](https://github.com/TUM-Dev/Campus-Android/assets/66640362/732b1ae7-a7cc-440d-a2e1-6eb79a0c2ae6)

PR #1594, opened 2023-07-06
App crash navigation drawer fix

### LINKED ISSUE #1517: [BUG] App crashes when using search function
**Describe the bug**
When clicking on the search button in the navigation drawer (second from top), the application closes every time with standard error message.

**To Reproduce**
1. Go to the navigation drawer on the left side
2. Click on the search button (in German labeled "Suchen")
3. See error

**Expected behavior**
I can search for persons, rooms, or whatever

**Smartphone (please complete the following information):**
* Phone: Samsung Galaxy A20e
* OS version: Android 11
* Language: German

kunzef:
## Issue

- Resolves: #1314 by replacing the old find method with a map
- Resolves: #1589
- Maybe resolves #1517 as the reason the app crashes when opening the search-function could be the same as in #1589

## Description of the fix
The error source of #1589 is the difference between the System locale and the Application locale. Somehow the Campus-App fails to correctly set its own Application locale. This leads to the crash as the NavigatioDrawer causing it, tries to find the clicked NavItem with a localized String resource. For example see this part of code from the NavigationDrawer `menuItem.title == activity.getString(HOME.titleRes)` here _menuItem.title_ is based in the Application locale where under certain cirqumstances _activity.getString()_ uses the System locale. It uses the System locale after any of the _NavItem.ActivityDestination_ has been opened. 

While it does not solve the underlying problem it still prevents the App from crashing and thus is at least a bit.

## Further todos
The main problem causing this other problems and #1585 as well as maybe others is not solved with this. In regards to this I have openend the issue #1593 See there for further details.


tobiasjungmann (approved):
The code looks good to me and I was not able to crash the app. The map init calls for loops at the first glance but every solution without the copy past I could come up with would not be that intuitive :D It is a nice solution for a lot of issues!

Issue #1582, opened 2023-06-01
MainScreen / CardView customization not working [BUG] 


kunzef:
**Describe the bug**
Multiple bugs with the MainScreen / CardView customization 
1. On first start MensaCard not displayed but enabled in Settings. Appears after opening Settings once
2. Removing a Card by swiping does not work for certain instances of NewsCard and the MensaCard. The Settings of the removed Cards are not updated.
3. `Restore cards` does not work at all. After exiting the CardView and reentering it  the `Restore cards`  button is not visible.

4. There are more bugs than the ones listed above including variations of the above ones as different Cards are behaving differently. I will subsequently add them to this issue. There are also more different Cards and thus more bugs when the user is logged in. Because I had to delete the App data every time as `Restore cards` does not work I have not yet included these.

**To Reproduce**
Bug 1          |  Bug 2 | Bug 3
:-------------------------:|:-------------------------:|:-------------------------:
https://github.com/TUM-Dev/Campus-Android/assets/45203772/93336b27-f562-4036-9843-2d35f948657f | https://github.com/TUM-Dev/Campus-Android/assets/45203772/fbdd8c81-c161-47ec-8655-8e24c99a4269 | https://github.com/TUM-Dev/Campus-Android/assets/45203772/6f21e5d5-c935-4b18-a5c4-e75da99e5992

**Smartphone (please complete the following information):**
* Phone: Google Pixel 6
* OS version: 13
* Language: Tried both English and German

**Speculated reasons**
- diveregence between CardView and Settings may be the result of different locations of the SharedPreferences

PR #1581, opened 2023-05-28
Replace anko

kunzef:
## Issue

This fixes the following issue(s):
- Resolves: #1573
- Improves how Asynchronous work is handled 

## Why this is useful for all students
Most likely users won't notice. But Anko has been archived since 2020 and thus not received improvements. Also the way how asynchronous work is done is not up to date / deprecated classes are used. This PR fixes this in all cases where previously Anko was used.

## How is Anko replaced
- for services through the corresponding native implementation
- for SharedPreferences through the androidx PreferenceManager
- asynchronous handling:
    - for impersistent work by binding to lifecycle using coroutines where possible
    - for persistent work by using WorkManager
- all other cases through native implementations see commit messages for every replacement

PR #1595, opened 2023-07-06
Fixes the CardView customization mentioned in #1582

kunzef:
## Issue

- Resolves: #1582 

## What it does
This PR rewrites the way Cards handle its discard and alyways hide functionality. It moves all information stored for the card in regards to the discard functionality into one SharedPreference per Card. This shared preference can then be completly wiped when restoring the card. The alyways hide functionality is set in the default SharedPreferences by a key defined in the new CardType enum. This key is stored in the String resource file making it easy to use across the Application and preventing accidental differnces in these keys. The usage of String resource also allows for seamless integration into the Settings.xml ensuring synchronity between the state of the Card. Also settings option for every Card were added because previously there was no way to restore these once alyawys hide was pressed. The CardTypes enum and the combination with the string resource also makes the addition of new Cards less error prone as it is more centralized. The PR also fixes a bug when refreshing the CardView leading to an error when using doAsync from Anko and a crash when using coroutines as mentioned [here](https://github.com/TUM-Dev/Campus-Android/issues/1581#issuecomment-1571176965).

This PR fixes the issues mentioned in #1582. Thus enhancing the user experience by restoring the excepted functionality.

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/ui/overview/MainFragment.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/ui/overview/MainFragment.kt
@@ -126,7 +126,6 @@ class MainFragment : BaseFragment<Unit>(
         swipeRefreshLayout.isRefreshing = true
         onRefresh()
         downloadNewsAlert()
-        println("Refresh finished")
     }
 
     private fun onNewCardsAvailable(cards: List<Card>) {
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/ui/overview/MainFragment.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/ui/overview/MainFragment.kt
@@ -126,6 +126,7 @@ class MainFragment : BaseFragment<Unit>(
         swipeRefreshLayout.isRefreshing = true
         onRefresh()
         downloadNewsAlert()
+        println("Refresh finished")
     }
 
     private fun onNewCardsAvailable(cards: List<Card>) {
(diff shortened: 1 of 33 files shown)


app/src/main/java/de/tum/in/tumcampusapp/component/ui/overview/MainFragment.kt
```diff
@@ -126,6 +126,7 @@ class MainFragment : BaseFragment<Unit>(
         swipeRefreshLayout.isRefreshing = true
         onRefresh()
         downloadNewsAlert()
+        println("Refresh finished")
```
tobiasjungmann, line 129: Ideally use logging or remove it to make it more coherent with the rest

tobiasjungmann (changes requested):
Please try to not format the whole file in the future - either use a different PR for reformatting / adapt the formatting rules in your IDE to the previously used version. That also happens to me every time :D
Some of your commit messages can be more intuitive ;)
Add some sort of comment (e.g. "Value is only written in the editor, changes are not yet copied back in the shared preferences.") to line 118 in `card.kt` to make obvious that the changes are not yet applied. I cannot annotate the function with a comment myself since there are no changes. Afterwards it should be ready to merge!

kunzef:
> Please try to not format the whole file in the future - either use a different PR for reformatting / adapt the formatting rules in your IDE to the previously used version. 

Are you refering to the first `Fixed formatting` commit? I think my error there was to accept the option _Remove custom line breaks_.
Otherwise I adopte the suggested changes.
Thanks for the feedback.

kunzef:
I was able to fix both issues. However, the fix for the visibility of the news card does not fix the root cause described here #1597. But it will work perfectly for the startpage without further issues.

tobiasjungmann (approved):
Amazing, thank you for looking into it! I could not spot any bug anymore.

PR #1599, opened 2023-07-22
Obsolescence Notification in Favour of Flutter

COM8:
Added a PSA about the planed replacement and the current state of development to the README.md.

This is partially in preparation for the OSLab next semester since a lot of students want to contribute to TC Apps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,14 @@
 </a>
 
 # TUM Campus App for Android
-### An excellent app for students by students
+### 🚨 Public Service Announcement 🚨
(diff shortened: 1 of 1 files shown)


kordianbruck (approved):

CommanderStorm:
@philippzagar what do you think: Should we add a similar message (different rationale, similar PSA) to https://github.com/TUM-Dev/Campus-iOS  ?

COM8:
I will also add something like this to the Windows TCA.

philippzagar:
Thanks for the ping, will add that to the iOS TCA Readme soon!

PR #1607, opened 2023-09-02
Added Chinese language support

YangDai2003:
## Issue

- Resolves: none

As a student at the Technical University of Munich and a Campus user, I would like to bring up a key aspect that could significantly increase the impact and impact of this project - adding Chinese language support. Here are some compelling reasons why I think supporting Chinese is critical:

1. **Diversified user group:** Among the student population of Technical University of Munich, Chinese students account for a considerable proportion. By providing support for Chinese, these Chinese students can use this app more conveniently and get the services they want, especially for those students who are new to Germany.

2. **Enhanced Accessibility:** Language can be a significant barrier in app usage, providing Chinese option will break this barrier.

3. **Competitive Advantage:** As the global education technology landscape continues to evolve, providing multilingual support is becoming a competitive advantage. Considering the user groups of different languages can better reflect the inclusiveness of the Technical University of Munich.

If you need help with the translation process or any related task, please know that I am eager to contribute to this work in any way possible. Together, we can make the Campus App an even more valuable resource for the student community.

### DIFF
--- a/app/src/main/java/de/tum/in/tumcampusapp/component/other/generic/drawer/DrawerMenuHelper.kt
+++ b/app/src/main/java/de/tum/in/tumcampusapp/component/other/generic/drawer/DrawerMenuHelper.kt
@@ -31,8 +31,8 @@ import de.tum.`in`.tumcampusapp.utils.plusAssign
 import java.util.Locale
 
 class DrawerMenuHelper(
-    private val activity: AppCompatActivity,
-    private val navigationView: NavigationView
+        private val activity: AppCompatActivity,
+        private val navigationView: NavigationView
 ) {
     private val germanContext: Context
         get() {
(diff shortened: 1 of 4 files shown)


CommanderStorm:
In general, I don't think this would be a bad thing (more languages).
The problem is that we are [currently migrating apps](https://github.com/TUM-Dev/Campus-Android?tab=readme-ov-file#-public-service-announcement-) to a platform-agnostic approach.
I am not quite convinced that adding more platform-specific features is the way to go.
If you want to contribute translatability to the future campus apps, this is the issue: https://github.com/TUM-Dev/Campus-Flutter/issues/81 ^^

Furthermore, such a change would need to come with a more scalable language subsystem as none of the current team members speak Chinese and blocking PRs based on translations is not a great workflow.
I think if a new (non-ubuquous, like Chinese) language is added, [weblate](https://docs.weblate.org/en/latest/formats.html) would be nessesary.
Currently, this is not a concern as everybody can speak english and german if they are studying here.

YangDai2003:
> In general, I don't think this would be a bad thing (more languages). The problem is that we are [currently migrating apps](https://github.com/TUM-Dev/Campus-Android?tab=readme-ov-file#-public-service-announcement-) to a platform-agnostic approach. I am not quite convinced that adding more platform-specific features is the way to go. If you want to contribute translatability to the future campus apps, this is the issue: [TUM-Dev/Campus-Flutter#81](https://github.com/TUM-Dev/Campus-Flutter/issues/81) ^^
> 
> Furthermore, such a change would need to come with a more scalable language subsystem as none of the current team members speak Chinese and blocking PRs based on translations is not a great workflow. I think if a new (non-ubuquous, like Chinese) language is added, [weblate](https://docs.weblate.org/en/latest/formats.html) would be nessesary. Currently, this is not a concern as everybody can speak english and german if they are studying here.

Well, you have a point, and unfortunately I wasn't able to help. 
Making multi-platform applications can greatly reduce workload and maintenance costs, and flutter is indeed a good choice, although I think Compose Multiplatform is even better👍. 
Anyway, good luck with your new project and look forward to the day when the new app is officially released🎉.

CommanderStorm:
I am going to close this issue, as currently I don't see us adding this feature for the above-mentioned reasons.
Sorry for the effort you spend on this.

